PR #44, opened 2022-02-01
Treat 0 as truthy in req() unless it's an action button

### LINKED ISSUE #43: Think harder about truthiness and action buttons
Currently `req(0)` throws a `SilentException` (because `if 0:` is falsy in python), which means we don't really need to special case `input_action_button()` like we do in R since `isTruthy(0)==TRUE`

cpsievert:
Closes #43


cclauss:
I am not a fan of treating 0 as truthy because it surprises the reader and thus would not be considered pythonic.

What about using `if arg or arg == 0:ˋ instead?

PR #65, opened 2022-02-05
Implement launching browser

jcheng5:
This works, but I can't get it to reuse a browser window (my [comment here](https://github.com/rstudio/prism/pull/59#issuecomment-1030254115) was incorrect--I don't know why I thought that worked... unless maybe I was on Linux when I tried it last?).

I really don't like the way this works now, if you hit Save on a .py file 5 times, now you've got 5 new browser tabs. I think we need to put a little more smarts into the browser. Maybe shiny.js can detect this kind of termination somehow, and after disconnect, repeatedly try to establish a websocket connection again (with exponential backoff, maybe). If a connection succeeds, then `window.location.reload()`.

### DIFF
--- a/shiny/app.py
+++ b/shiny/app.py
@@ -1,7 +1,9 @@
 __all__ = ("App",)
 
 import contextlib
+import os
 from typing import Any, List, Optional, Union, Dict, Callable, cast
+import webbrowser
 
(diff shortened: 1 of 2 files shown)


jcheng5:
Things have changed a lot since Feb 4. Shiny now does autoreload properly on uvicorn reload, so the only browser we need to launch is the initial one.

PR #127, opened 2022-04-19
Add `include_javascript()`, `include_css()`, and `include_html()`

cpsievert:
Currently requires https://github.com/rstudio/py-htmltools/pull/32 (since, if we want to return a `Tag`, we need to know the final `src`/`href`).

TODO: 
- Add a `include_markdown()` that utilizes `markdown()`?
- Probably split up `include_html()` into `include_html()` and `include_html_frame()`?
- Finish `include_html()`/`include_html_frame()` examples

### DIFF
--- a/shiny/examples/include_css/app.py
+++ b/shiny/examples/include_css/app.py
@@ -1,12 +1,12 @@
-import os
+from pathlib import Path
 
 from shiny import *
 
-css_file = os.path.join(os.path.dirname(__file__), "css/styles.css")
+css_file = Path(__file__).parent / "css" / "styles.css"
 
 app_ui = ui.page_fluid(
     "Almost before we knew it, we had left the ground!!!",
-    ui.include_css(css_file, method="link_files"),
+    ui.include_css(css_file),
     ui.div(
         # Style individual elements with an attribute dictionary.
         {"style": "font-weight: bold"},
--- a/shiny/examples/include_javascript/app.py
+++ b/shiny/examples/include_javascript/app.py
@@ -1,8 +1,8 @@
-import os
+from pathlib import Path
 
 from shiny import *
 
-js_file = os.path.join(os.path.dirname(__file__), "js/app.js")
+js_file = Path(__file__).parent / "js" / "app.js"
 
 app_ui = ui.page_fluid(
     "If you see this page before 'OK'-ing the alert box, something went wrong",
--- a/shiny/ui/_include_helpers.py
+++ b/shiny/ui/_include_helpers.py
@@ -82,12 +82,13 @@ def include_js(
     ~ui.tags.script
     ~include_css
     """
+    file_path = check_path(path)
 
     if method == "inline":
-        return tags.script(read_utf8(path), **kwargs)
+        return tags.script(read_utf8(file_path), **kwargs)
 
     include_files = method == "link_files"
-    path_dest, hash = maybe_copy_files(path, include_files)
+    path_dest, hash = maybe_copy_files(file_path, include_files)
 
     dep, src = create_include_dependency("include-js-" + hash, path_dest, include_files)
 
@@ -159,11 +160,12 @@ def include_css(
     ~include_js
     """
 
+    file_path = check_path(path)
     if method == "inline":
-        return tags.style(read_utf8(path), type="text/css")
+        return tags.style(read_utf8(file_path), type="text/css")
 
     include_files = method == "link_files"
-    path_dest, hash = maybe_copy_files(path, include_files)
+    path_dest, hash = maybe_copy_files(file_path, include_files)
 
     dep, src = create_include_dependency(
         "include-css-" + hash, path_dest, include_files
@@ -175,6 +177,17 @@ def include_css(
 # ---------------------------------------------------------------------------
 # Include helpers
 # ---------------------------------------------------------------------------
+
+
+def check_path(path: Path | str) -> Path:
+    path = Path(path)
+    if not path.exists():
+        err = f"""
+        {path.absolute()} does not exist.
+        Files are typically placed in the app directory and refered to with 'Path(__file__) / {path.name}'
+        """
+        raise RuntimeError(err)
+    return path
 
 
 def create_include_dependency(
@@ -232,8 +245,8 @@ def get_hash(path: Path | str, include_files: bool) -> str:
 
 
 def get_file_key(path: Path | str) -> str:
-    path_str = str(path) if isinstance(path, Path) else path
-    return path_str + "-" + str(os.path.getmtime(path_str))
+    path = Path(path)
+    return str(path) + "-" + str(path.stat().st_mtime)
 
 
 def hash_deterministic(s: str) -> str:


Polkas:
linked to https://github.com/rstudio/py-htmltools/issues/42
Please remember there is no `htmltools::singleton` now in the python version. In my opinion we need such especially when connected with this PR functions will be introduced.

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,300 @@
...
+
+from .._docstring import add_example
+
+# TODO: it's bummer that, when method="link_files" and path is in the same directory
+# as the app, the app's source will be included. Should we just not copy .py/.r files?
+
+
+@add_example()
+def include_javascript(
+    path: str, *, method: Literal["link", "link_files", "inline"] = "link"
```
wch, line 21: I think we'll want to support attributes to pass along to the `<script>` tag. At the very least some users (like me) will want to use `type="module"`.

shiny/examples/include_css/css/styles.css
```diff
@@ -0,0 +1,10 @@
+@font-face {
+  font-family: 'Square Peg';
+  src: url('SquarePeg-Regular.ttf');
```
wch, line 3: I think that this example shouldn't include the font, because it lives in `/shiny/examples`. That means that the font will be distributed with the wheel file (and make it 120KB bigger), but almost no one will ever make use of it.

We could put an example with a font in `/examples`, because that dir isn't distributed with the built package.

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,300 @@
...
+from htmltools._util import hash_deterministic
+
+from .._docstring import add_example
+
+# TODO: it's bummer that, when method="link_files" and path is in the same directory
+# as the app, the app's source will be included. Should we just not copy .py/.r files?
+
+
+@add_example()
+def include_javascript(
```
wch, line 20: I think `include_js` is a better name -- it's shorter, still clear, and has a similar form as `include_css`.

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,300 @@
...
+    may result in a Flash of Unstyled Content (FOUC). To instead place the CSS in the
+    :func:`~ui.tags.head` of the document, you can wrap it in ``head_content``:
+
+    .. code-block:: python
+
+    from htmltools import head_content from shiny import ui
+
+    ui.fluidPage(
+        head_content(ui.include_css("custom.css"))
+    )
```
wch, line 133: It would be very helpful if there was also an example that showed how to inline css (with something like `head_content(ui.tags.style(" .... "))`, because many users will see `include_css` in the API docs and will be looking for that.

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,300 @@
...
+    document, you can wrap it in ``head_content`` (in this case, just make sure you're
+    aware that the DOM probably won't be ready when the script is executed).
+
+    .. code-block:: python
+
+    from htmltools import head_content from shiny import ui
+
+    ui.fluidPage(
+        head_content(ui.include_javascript("custom.js"))
+    )
```
wch, line 66: It would be very helpful if there was also an example that showed how to inline JS (with something like `head_content(ui.tags.script(" .... "))`.

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,300 @@
...
+    return dep, src
+
+
+def maybe_copy_files(path: str, include_files: bool) -> Tuple[str, str]:
+    hash = get_hash(path, include_files)
+
+    # To avoid unnecessary work when the same file is included multiple times,
+    # use a directory scoped by a hash of the file.
+    tmpdir = os.path.join(tempfile.gettempdir(), "shiny_include_files", hash)
+    path_dest = os.path.join(tmpdir, os.path.basename(path))
```
wch, line 261: I find it a bit concerning that we copy files to a temp dir and there's no provision for removing them after app stops or when Python exits... but then again, I see a lot of junk in my temp dir that has been there for a while, so maybe this isn't so bad.

wch (changes requested):

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,300 @@
...
+    dep, src = create_include_dependency(
+        "include-css-" + hash, path_dest, include_files
+    )
+
+    return tags.link(dep, href=src, rel="stylesheet")
+
+
+# TODO: maybe support remote URLs?
+@add_example()
+def include_html(
```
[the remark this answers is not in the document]
GShotwell: I agree let's remove it. I also think that include_html should be packaged with include_md and maybe include_qmd.

shiny/examples/include_html/app.py
```diff
@@ -0,0 +1,10 @@
+from shiny import *
+
+app_ui = ui.page_fluid()
```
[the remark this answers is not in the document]
GShotwell: Resolved because we're removing include_html

shiny/ui/_include_helpers.py
GShotwell: Also these files are probably going to be pretty small relative to what goes into temp

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,247 @@
...
+    )
+
+    See Also
+    --------
+    ~ui.tags.script
+    ~include_css
+    """
+
+    if method == "inline":
+        return tags.script(read_utf8(path), type="text/javascript", **kwargs)
```
wch, line 87: The `type="text/javascript"` here will conflict with `type="module"`, if the user provides the latter argument.

We can just remove the `type="text/javascript"` here. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type):
> Authors are encouraged to omit the attribute if the script refers to JavaScript code rather than specify a MIME type.

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,247 @@
...
+            should be located under ``/app/js/``).
+          * ``"inline"``: Inline the JS file contents within a :func:`~ui.tags.script`
+            tag.
+    *kwargs
+        Attributes which are passed on to `~ui.tags.script`
+
+
+    Returns
+    -------
+    A :func:`~ui.tags.script` tag.
```
wch, line 58: The `Returns` section needs somewhat different formatting. For example, see:
https://github.com/rstudio/py-shiny/blob/4ea14db2c6744a6a38f79a591d31595e073546f1/shiny/reactive/_reactives.py#L145-L148

```suggestion
    Returns
    -------
    :
        A :func:`~ui.tags.script` tag.
```

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,247 @@
...
+    This places a :func:`~ui.tags.script` tag in the :func:`~ui.tags.body` of the
+    document. If instead, you want to place the tag in the :func:`~ui.tags.head` of the
+    document, you can wrap it in ``head_content`` (in this case, just make sure you're
+    aware that the DOM probably won't be ready when the script is executed).
+
+    .. code-block:: python
+
+
+
+    ui.fluidPage(
```
wch, line 71:
```suggestion
    .. code-block:: python

    ui.fluidPage(
```

shiny/examples/include_css/css/styles.css
wch: This was marked as resolved, but I think the line of code still needs to be removed -- otherwise, won't the CSS still try to load the font?

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,247 @@
...
+from .._docstring import add_example
+from .._typing_extensions import Literal
+
+# TODO: it's bummer that, when method="link_files" and path is in the same directory
+# as the app, the app's source will be included. Should we just not copy .py/.r files?
+
+
+@add_example()
+def include_js(
+    path: str,
```
wch, line 23: Sorry I just realized this now, but it would be more convenient for users if this were `Path | str` (and we had `from pathlib import path` above).

shiny/ui/_include_helpers.py
wch: Also, I think the actual code needs to be indented. (At least, all the examples I see out there have indented code.)

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,247 @@
...
+            then ``path`` should be somewhere like ``/app/css/custom.css`` (and all the
+            other relevant accompanying 'safe' files should be located under
+            ``/app/css/``).
+          * ``"inline"``: Inline the CSS file contents within a :func:`~ui.tags.style`
+            tag.
+
+    Returns
+    -------
+    If ``method="inline"``, returns a :func:`~ui.tags.style` tag; otherwise, returns a
+    :func:`~ui.tags.link` tag.
```
wch, line 129: Same formatting issue as above, with `:`.

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,247 @@
...
+
+    ui.fluidPage(
+        head_content(ui.include_css("custom.css")),
+
+        # You can also inline css by passing a dictionary with a `style` element.
+        ui.div(
+            {"style": "font-weight: bold;"},
+            ui.p("Some text!"),
+        )
+    )
```
wch, line 150: Indent

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,247 @@
...
+    path
+        A path to a JS file.
+    method
+        One of the following:
+          * ``"link"``: Link to the JS file via a :func:`~ui.tags.script` tag. This
+            method is generally preferrable to ``"inline"`` since it allows the browser
+            to cache the file.
+          * ``"link_files"``: Same as ``"link"``, but also allow for the CSS file to
+            request other files within ``path``'s immediate parent directory (e.g.,
+            ``fetch()`` the contents of another file). Note that this isn't the default
```
wch, line 42:
```suggestion
          * ``"link_files"``: Same as ``"link"``, but also allow for the JS file to
            request other files within ``path``'s immediate parent directory (e.g.,
            ``import()` another file, if it is loaded with `type="module"`). Note that this isn't the default
```

shiny/ui/_include_helpers.py
```diff
@@ -0,0 +1,247 @@
...
+            behavior because you should **be careful not to include files in the same
+            directory as ``path`` that contain sensitive information**. A good general
+            rule of thumb to follow is to have ``path`` be located in a subdirectory of
+            the app directory. For example, if the app's source is located at
+            ``/app/app.py``, then ``path`` should be somewhere like
+            ``/app/js/custom.js`` (and all the other relevant accompanying 'safe' files
+            should be located under ``/app/js/``).
+          * ``"inline"``: Inline the JS file contents within a :func:`~ui.tags.script`
+            tag.
+    *kwargs
```
wch, line 52:
```suggestion
    **kwargs
```

wch (changes requested):

shiny/examples/include_css/css/styles.css
GShotwell: Whoops, you're correct, sorry about missing that.

shiny/ui/_include_helpers.py
```diff
@@ -227,8 +231,9 @@ def get_hash(path: str, include_files: bool) -> str:
     return hash_deterministic(key)
 
 
-def get_file_key(path: str) -> str:
-    return path + "-" + str(os.path.getmtime(path))
+def get_file_key(path: Path | str) -> str:
+    path_str = str(path) if isinstance(path, Path) else path
+    return path_str + "-" + str(os.path.getmtime(path_str))
```
wch, line 236: If `path` is a string, `str(path)` will return the value unchanged, so you could simplify for clarity:

```suggestion
    path = str(path)
    return path + "-" + str(os.path.getmtime(path))
```

Although I personally prefer working with `Path` objects, like this:

```suggestion
    path = Path(path)
    return str(path) + "-" + str(path.stat().st_mtime)
```

shiny/examples/include_css/app.py
```diff
@@ -0,0 +1,17 @@
+import os
+
+from shiny import *
+
+css_file = os.path.join(os.path.dirname(__file__), "css/styles.css")
```
wch, line 5:
```suggestion
from pathlib import Path

css_file = Path(__file__).parent / "css" / "styles.css"
```

(Note that you should edit this locally because isort will reorganize the imports on save.)

shiny/examples/include_javascript/app.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+from shiny import *
+
+js_file = os.path.join(os.path.dirname(__file__), "js/app.js")
```
wch, line 5:
```py
from pathlib import Path
js_file = Path(__file__).parent / "js" / "app.js"
```

wch (commented):
I had some more thoughts about path handling and the user experience. What happens if someone writes this reasonable-looking code?

```py
app_ui = ui.page_fluid(
    ui.include_css("css/my_styles.css"),
    ...
)
```

It's a relative path, but we can make any assumptions about which directory is the current working directory. This could result in a not-very-clear error that will leave the user scratching their head.

So we may want to test that the path is an absolute path at the top of `include_js` and `include_css`, with something roughly like:

```py
path = Path(path)
if not path.is_absolute():
    raise RuntimeError('include_js() requires an absolute path. Please use
  include_js(Path(__file__)).parent / "myfile.css"') 
```

(But with a better error message)

The other comments I have about `Path` are related to this. When the PR was originally created, we mostly used `os.path`, but since then we generally switched to `pathlib.Path`.

GShotwell:
@wch It seems like `Path("css/my_css.css").absolute()` will always return the absolute path if it can find the file. Do we really need that check?

wch:
The issue is that if you start the app from a different directory, then the path will be relative to that directory. For example:

```
[winston@MBP16:~]$ pwd
/Users/winston

[winston@MBP16:~]$ cat temp/app.py
from pathlib import Path

print(Path("css/my_css.css").absolute())

...

[winston@MBP16:~]$ shiny run temp/app.py
/Users/winston/css/my_css.css
```

We would want it the printed path to be `/Users/winston/temp/css/my_css.css`, not `/Users/winston/css/my_css.css`.

As far as I know, the only reasonable way to do that is to have the user provide `__file__` somwhere.

GShotwell:
Oh I see, so we really want to be checking if the file exists? I'm not sure we want to enforce the `__file__` pattern because they might be using css which is hosted outside of the app folder right? Like if you had a single css file which was used by multiple apps. This isn't a great pattern but I don't think it should provoke an error. 

What about this:

```
def check_path(path: Path) -> Path:
    path = Path(path)
    if not path.exists():
        err = f'{path.name} does not exist. Typically you should refer to this file with 'Path(__file__) / {path.name}'")'
        raise RuntimeError(err)
    return path
    ```

wch:
> Oh I see, so we really want to be checking if the file exists?

Yes, that makes sense.

> I'm not sure we want to enforce the `__file__` pattern because they might be using css which is hosted outside of the app folder right? Like if you had a single css file which was used by multiple apps. This isn't a great pattern but I don't think it should provoke an error.

Good point. I think we should use the `__file__` pattern in our examples, though.

> 
> What about this:
> 
> ```
> def check_path(path: Path) -> Path:
>     path = Path(path)
>     if not path.exists():
>         err = f'{path.name} does not exist. Typically you should refer to this file with 'Path(__file__) / {path.name}'")'
>         raise RuntimeError(err)
>     return path
>     ```
> ```

That sounds good to me.

wch (approved):

PR #285, opened 2022-07-30
Upgrade GitHub Actions

cclauss:
https://github.com/actions/checkout/releases
https://github.com/actions/setup-python/releases
https://github.com/pypa/gh-action-pypi-publish/releases

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -18,10 +18,10 @@ jobs:
       fail-fast: false
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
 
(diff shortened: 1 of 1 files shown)


wch:
I made some changes to the workflow on `main`, because these repositories are now public. Can you rebase onto `main` and push?

.github/workflows/pytest.yaml
```diff
@@ -98,7 +98,7 @@ jobs:
 
       # test deploy ----
       - name: "Test Deploy to PyPI"
-        uses: pypa/gh-action-pypi-publish@release/v1
+        uses: pypa/gh-action-pypi-publish@v1
```
wch, line 101: Are you sure this is right? I see a branch named `release/v1`, but not one named `v1`. (And same for the other instance below.)

cclauss:
```suggestion
        uses: pypa/gh-action-pypi-publish@release/v1
```

.github/workflows/pytest.yaml
```diff
@@ -107,7 +107,7 @@ jobs:
 
       ## prod deploy ----
       - name: "Deploy to PyPI"
-        uses: pypa/gh-action-pypi-publish@master
+        uses: pypa/gh-action-pypi-publish@v1
```
cclauss, line 110:
```suggestion
        uses: pypa/gh-action-pypi-publish@release/v1
```

wch:
Thanks!

PR #292, opened 2022-08-01
Avoid automatically namespaced input read in render.plot

### LINKED ISSUE #290: @render.plot doesn't yield a result inside a @module.server
For example this should yield a plot, but doesn't (currently):

```py
from shiny import ui, App, render, module
import numpy as np


@module.ui
def my_ui():
    return ui.TagList(
        ui.input_slider(
            "n", "input_slider()", min=10, max=100, value=50, step=5, animate=True
        ),
        ui.output_plot("p"),
    )

app_ui = ui.page_fluid(my_ui("ns"))

@module.server
def my_server(input, output, session):
    @output
    @render.plot
    def p():
        np.random.seed(19680801)
        x_rand = 100 + 15 * np.random.randn(437)
        fig, ax = plt.subplots()
        ax.hist(x_rand, int(input.n()), density=True)
        return fig


    my_server("ns")

app = App(app_ui, server)
```

cpsievert:
Closes #290

### DIFF
--- a/shiny/render/_render.py
+++ b/shiny/render/_render.py
@@ -209,7 +209,6 @@ class RenderPlot(RenderFunction[object, Union[ImgData, None]]):
         # static type checker that it's synchronous. wrap_async() is smart -- if is
         # passed an async function, it will not change it.
         self._fn: RenderPlotFuncAsync = _utils.wrap_async(fn)
-        print(self.__dict__)
 
     def __call__(self) -> Union[ImgData, None]:
         return _utils.run_coro_sync(self._run())
--- a/shiny/render/_render.py
+++ b/shiny/render/_render.py
@@ -53,6 +53,7 @@ if TYPE_CHECKING:
     from ..session import Session
     from ..session._utils import RenderedDeps
 
+from .._namespaces import ResolvedId
 from .. import _utils
 from ..types import ImgData
 from ._try_render_plot import (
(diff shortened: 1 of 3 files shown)


shiny/render/_render.py
```diff
@@ -208,20 +208,22 @@ def __init__(
         # static type checker that it's synchronous. wrap_async() is smart -- if is
         # passed an async function, it will not change it.
         self._fn: RenderPlotFuncAsync = _utils.wrap_async(fn)
+        print(self.__dict__)
```
wch, line 211: Is this left over from debugging?

shiny/render/_render.py
```diff
@@ -208,20 +208,22 @@ def __init__(
...
 
     def __call__(self) -> Union[ImgData, None]:
         return _utils.run_coro_sync(self._run())
 
     async def _run(self) -> Union[ImgData, None]:
+
         # Reactively read some information about the plot.
         pixelratio: float = typing.cast(
-            float, self._session.input[".clientdata_pixelratio"]()
+            float, self._session.input._map[".clientdata_pixelratio"]()
```
wch, line 220: I think it's probably a bit dangerous to access `_map` directly from outside the object, because the `Inputs` object sometimes does special handling before accessing `_map`. For example, here it populates an item if it doesn't already exist, so that it doesn't throw an error:

https://github.com/rstudio/py-shiny/blob/8ae23bfc3f2dbd05f2124dac43df3fe0dd3eefd0/shiny/session/_session.py#L863-L871

Maybe we should have a method like this?

```py
def _get(self, use_ns=True): ...
```

jcheng5: Or else, `session.parent` and/or `session.root`, this is how I fixed a similar issue in shinywidgets https://github.com/rstudio/py-shinywidgets/pull/50/files#diff-7f04fb8c4f02a66b160fbbf98fa8df1a0aea7e744a9613b9abf6d90a49674fb4R66-R67

shiny/render/_render.py
cpsievert: Oops, yep

jcheng5 (approved):

jcheng5:
Looks good now. Now that we're public, do we need to start maintaining a CHANGELOG?

PR #294, opened 2022-08-04
Remove dependence on Starlette's lifespan mechanism

### LINKED ISSUE #279: reactivity failing when mounting app as route on starlette_app
Attached is a small reproducible example. The issue I am facing is a shiny app that is defined, runs as expected as standalone. 
But when I mount the app as a route in a starlette app, the render function does not appear to trigger.

[mount_issue.zip](https://github.com/rstudio/py-shiny/files/9207825/mount_issue.zip)

with source attached to reproduce:

```
unzip mount_issue.zip
virtualenv .venv
source .venv/bin/activate
pip install -r requirements.txt
# works as expected
uvicorn issue:app_shiny
# no plot appears
uvicorn issue:star_app
```

jcheng5:
Fixes #279.

Reactive flush is what causes outputs and errors to get queued up
for sessions; a second step (unforunately also called "flush") is
necessary for those outputs and errors to be sent to clients. This
was achieved by subscribing to reactive._core.on_flushed with a
handler that triggers session._flush.

Shiny's reactive environment is global to a Python process. It is
possible for a single process to run multiple Shiny apps over time.
So the subscription to reactive._core.on_flushed was happening
during app startup and unsubscribed during app shutdown, as reported
via the Starlette lifespan feature.

However, if a Shiny app is mounted to a Starlette route, then its
lifespan never comes into play. This resulted in issue "#279:
reactivity failing when mounting app as route on starlette_app".

This commit fixes the problem by no longer using lifespan, but
instead, having each session subscribe/unsubscribe to on_flushed.

### DIFF
--- a/shiny/session/_session.py
+++ b/shiny/session/_session.py
@@ -249,7 +249,7 @@ class Session(object, metaclass=SessionMeta):
                     {
                         "config": {
                             "workerId": "",
-                            "sessionId": str(self.id),
+                            "sessionId": self.id,
                             "user": None,
                         }
                     }
--- a/shiny/session/_session.py
+++ b/shiny/session/_session.py
@@ -1,5 +1,6 @@
 __all__ = ("Session", "Inputs", "Outputs")
 
+import contextlib
 import dataclasses
 import enum
 import functools
@@ -50,7 +51,7 @@ from .._namespaces import Id, ResolvedId, Root
 from ..http_staticfiles import FileResponse
 from ..input_handler import input_handlers
 from ..reactive import Effect, Effect_, Value, flush, isolate
-from ..reactive._core import lock
+from ..reactive._core import lock, on_flushed
 from ..render import RenderFunction
 from ..types import SafeException, SilentCancelOutputException, SilentException
 from ._utils import RenderedDeps, read_thunk_opt, session_context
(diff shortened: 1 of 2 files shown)


jcheng5:
The change to _session.py is much less scary looking if you hide whitespace changes.

shiny/session/_session.py
```diff
@@ -242,72 +243,86 @@ def verify_state(expected_state: ConnectionState) -> None:
...
-            await self._send_message(
-                {"config": {"workerId": "", "sessionId": str(self.id), "user": None}}
-            )
+        with contextlib.ExitStack() as stack:
+            try:
+                await self._send_message(
+                    {
+                        "config": {
+                            "workerId": "",
+                            "sessionId": str(self.id),
```
wch, line 252: I know this is just due to indentation changes, but as long as you're in here, the `str()` is no longer needed.

```suggestion
                            "sessionId": self.id,
```

shiny/session/_session.py
```diff
@@ -242,72 +243,86 @@ def verify_state(expected_state: ConnectionState) -> None:
...
-                        verify_state(ConnectionState.Start)
+                        if message_obj["method"] == "init":
+                            verify_state(ConnectionState.Start)
 
-                        conn_state = ConnectionState.Running
-                        message_obj = typing.cast(ClientMessageInit, message_obj)
-                        self._manage_inputs(message_obj["data"])
+                            # When a reactive flush occurs, flush the session's outputs,
+                            # errors, etc. to the client.
+                            unreg = on_flushed(self._flush)
```
wch, line 282: Can you add a comment making it explicit that this is `reactive._core.on_flushed`? I know the comment mentions "reactive flush", but I still was confused for a while and thought this was the `self.on_flushed` method.

wch:
This looks good to me. I like that it simplifies the code.

One question: with this change, how would shinylive run an app? Currently it uses `app._lifespan` to start and stop an app. See here for starting, and a few lines below for stopping:
https://github.com/rstudio/py-shinylive/blob/194640d9a1122c26e99dcad1fad88998c0249b4c/src/hooks/usePyodide.tsx#L393-L395

PR #299, opened 2022-08-05
Allow app UI to be provided as a function

### LINKED ISSUE #269: Is it possible to create a non-cached UI?
In R-shiny, wrapping the UI in a `function(req)` results in a UI that gets generated on runtime of every session. This means that we can have some dynamic content in the UI.

I couldn't find anything in the documentation nor with experimentation that shows this feature in py-shiny.

jcheng5:
Fixes #269

### DIFF
--- a/shiny/_app.py
+++ b/shiny/_app.py
@@ -1,5 +1,4 @@
 import copy
-from inspect import iscoroutinefunction
 import os
 import secrets
 from typing import Any, List, Union, Dict, Callable, cast, Optional
@@ -19,6 +18,7 @@ from ._autoreload import autoreload_url, InjectAutoreloadMiddleware
 from ._connection import Connection, StarletteConnection
 from ._shinyenv import is_pyodide
 from ._error import ErrorMiddleware
+from ._utils import is_async_callable
 from .html_dependencies import require_deps, jquery_deps, shiny_deps
 from .http_staticfiles import StaticFiles
 from .session import Inputs, Outputs, Session, session_context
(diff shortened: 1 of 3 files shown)


shiny/_app.py
```diff
@@ -138,6 +140,15 @@ def _server(inputs: Inputs, outputs: Outputs, session: Session):
 
         self.starlette_app = starlette_app
 
+        if isfunction(ui):
```
wch, line 143: This won't catch non-function callable objects. I think you can use `callable()` instead: https://docs.python.org/3/library/functions.html#callable

(And same below on line 280)

wch (approved):
Other than the one comment, looks good to me!

shiny/_app.py
```diff
@@ -138,6 +140,17 @@ def _server(inputs: Inputs, outputs: Outputs, session: Session):
 
         self.starlette_app = starlette_app
 
+        if is_uifunc(ui):
+            if iscoroutinefunction(ui):
```
wch, line 144: I think you'd want to use `_utils.is_async_callable()` here instead of `iscoroutinefunction()`.

But then again, perhaps we should allow async callables? One thing we could do is pass both sync and async functions to `_utils.wrap_async()`. That function always returns an async function, so when we call it later on, we would do `await self.ui(request)`.

That said, I don't foresee a whole lot of use cases for async UI functions, so only allowing sync functions would probably be fine in practice.

jcheng5: I'm thinking we'll do async on sort of an as-needed basis, and like you I didn't foresee a lot of use cases for it here.

PR #304, opened 2022-08-09
Change order of types

wch:
With pyright 1.1.266, we started seeing new errors like this:

```
  /home/runner/work/py-shiny/py-shiny/shiny/reactive/_reactives.py:411:20 - error: Expression of type "CalcAsync_[Awaitable[T@Calc] | T@Calc]" cannot be assigned to return type "Calc_[T@Calc]"
    TypeVar "T@Calc_" is invariant
      Type "Awaitable[T@Calc] | T@Calc" cannot be assigned to type "T@Calc" (reportGeneralTypeIssues)
```

This is due to a bug (I think) in pyright: https://github.com/microsoft/pyright/issues/3813

This PR changes the order of types in the `Union` to avoid the false positive.

### DIFF
--- a/shiny/reactive/_reactives.py
+++ b/shiny/reactive/_reactives.py
@@ -406,7 +406,7 @@ def Calc(
     ~shiny.event
     """
 
-    def create_calc(fn: Union[CalcFunctionAsync[T], CalcFunction[T]]) -> Calc_[T]:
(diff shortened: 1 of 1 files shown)


jcheng5:
How are you able to figure these out? I think this is the second or third of these you’ve sleuthed out.

wch:
What usually happens is that I see new errors in our CI checks, then I try to come up with a minimal reproducible example to get an understanding of whether it's our code or pyright that has a problem.

Issue #275, opened 2022-07-28
Support a few common actions on inputs: reset, hide, show, disable, enable


daattali:
This is for much later on, when py-shiny is out of alpha. I think it would be useful for a sizeable proportion of users if a few of the most common {shinyjs} functions existed in py-shiny natively. The ones in the title are in my opinion the most popular. They often feel like very basic shiny actions that I wish were included in base shiny.

jcheng5:
It's a good time to revisit this for Shiny for R as well, probably. @cpsievert @gregswinehart

daattali:
At whatever point in the future when shiny is ready to start planning these features, I'd like to pitch in a few ideas if you could reach out

sdhutchins:
I'd add...the python community will probably hop on a lot of this fairly quickly. I've wanted to see this for a while, and I have a number of ideas on how to incorporate base python even better.

PR #306, opened 2022-08-09
Add function to run javascript from server

andyquinterom:
This is a simple funciton that allows the user to run javascript from the server. I took my inspiration from the shinyjs package on the R side. This is a feature I pretty much require in every app I build so I thought it would be cool to introduce it.

I placed it as part of the UI module since I belive it makes sense. The javascript runs on the client. However if you disagree there's no problem.

### DIFF
--- a/shiny/session/_session.py
+++ b/shiny/session/_session.py
@@ -581,6 +581,9 @@ class Session(object, metaclass=SessionMeta):
         msg: Dict[str, object] = {"progress": {"type": type, "message": message}}
         self._send_message_sync(msg)
 
+    def _send_run_js(self, code: str) -> None:
+        self._send_message_sync({"javascript": code})
+
     @add_example()
     async def send_custom_message(self, type: str, message: Dict[str, object]) -> None:
         """
--- a/shiny/ui/__init__.py
+++ b/shiny/ui/__init__.py
@@ -49,6 +49,7 @@ from ._input_update import (
     update_navs,
 )
 from ._insert import insert_ui, remove_ui
+from ._run_js import run_js
 from ._markdown import markdown
 from ._modal import modal_button, modal, modal_show, modal_remove
 from ._navs import (
@@ -154,6 +155,7 @@ __all__ = (
     "update_navs",
     "insert_ui",
     "remove_ui",
+    "run_js",
     "markdown",
     "modal_button",
     "modal",
--- a/shiny/ui/_run_js.py
+++ b/shiny/ui/_run_js.py
@@ -1,0 +1,43 @@
+__all__ = "run_js"
+
+import sys
+from typing import Optional
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
+
+from .._docstring import add_example
+from ..session import Session, require_active_session
+
+
+@add_example()
+def run_js(
+    code: str, immediate: bool = False, session: Optional[Session] = None
+) -> None:
+    """
+    Run JavaScript from server
+
+    Parameters
+    ----------
+    code
+        The arbitrary code snippet you want the session client to run.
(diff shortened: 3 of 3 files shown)


Polkas:
linked to https://github.com/rstudio/py-shiny/issues/275

cpsievert:
If the intention here is to just execute some JavaScript, that's already possible by providing a `ui.tags.script()` to `ui.insert_ui()`.

andyquinterom:
Could there be significant differences in the way browsers interpret appending script tags rather than directly running some js? I've had issues in the past dealing with insertUI and script tags on different browsers. Could be worth looking into.

cpsievert:
> Could there be significant differences in the way browsers interpret appending script tags rather than directly running some js?

It could be, but I've never encountered this. If you have any hard evidence, please let us know!

> I've had issues in the past dealing with insertUI and script tags

The most recent version of R Shiny actually includes a fix for this exact use case (I'm actually not sure if, in the past, the script would actually execute at all) https://github.com/rstudio/shiny/blame/main/NEWS.md#L17

andyquinterom:
> > Could there be significant differences in the way browsers interpret appending script tags rather than directly running some js?
> 
> It could be, but I've never encountered this. If you have any hard evidence, please let us know!
> 
> > I've had issues in the past dealing with insertUI and script tags
> 
> The most recent version of R Shiny actually includes a fix for this exact use case (I'm actually not sure if, in the past, the script would actually execute at all) https://github.com/rstudio/shiny/blame/main/NEWS.md#L17

This is pretty much it. I do not believe it executed in the past. Since this seems to be it I will leave it as it is. Perhaps developing a separate set or wrappers for such an insertion could be useful to users as mentioned in #275

jcheng5:
@cpsievert IIRC us explicitly supporting `<script>` tags in `insert_ui` makes it difficult for us to ever move on from jQuery if we wanted to, right? (It's already all but impossible on the R side for backwards compatibility reasons, but it's not out of the question for Python)

andyquinterom:
@jcheng5 @cpsievert should we open an issue on this specific topic or continue in this PR?

jcheng5:
Let's continue discussing here, thanks...

andyquinterom:
What components have jquery as a dependencie?

Selectize, datatables, slider, datepicker.

Any others?

cpsievert:
> IIRC us explicitly supporting <script> tags in insert_ui makes it difficult for us to ever move on from jQuery if we wanted to, right?

That might be the case, but I don't know for sure. I could be wrong, but vaguely inserting UI the "vanilla JS" way will execute top-level `<script>` tags, but I don't think it's straightforward to execute child `<script>` tags, which is unfortunate for `input_slider()`, `panel_absolute()`, etc

wch:
In plain JS (not jQuery), if you insert a `<script>` tag by setting a DOM element's `innerHTML`, it will not execute the script. But if you _create_ the tag with `document.createElement('script')`, then it will execute the script.
See https://stackoverflow.com/a/13392818/412655

If we were to drop jQuery and people run JS code by using `insert_ui` with HTML containing `<script>` tags, then on the JS side, we would probably have to do something like: parse the HTML, extract the `<script>` tag, and then insert it with `document.createElement('script')`.

wch:
> What components have jquery as a dependencie?
> 
> Selectize, datatables, slider, datepicker.

One of the challenges for dropping jQuery is that there are a number of third-party Shiny extensions that expect jQuery to be present. (We are currently using the exact same client-side shiny.js code in Python and in R.)

Updating our own code to drop jQuery is doable, but doing it in a way that doesn't break these extensions will be difficult.

aghasemi:
Hi. Does the `tags` method also work for running some JS code _"and getting the result/return value"_?

PR #310, opened 2022-08-10
Delegate shinylive deployment to code in shinylive

wch:
This PR makes it so that when `shiny static` is called, it runs code provided by the shinylive bundle. This makes it so that shinylive can make changes to how it does deployment, without having to synchronize it with the shiny package.

Along with https://github.com/rstudio/py-shinylive/pull/14, it also makes the following possible:

*****

Multiple applications can be deployed in one command, with:

```py
shiny static app1 site --subdir a1
shiny static app2 site --subdir a2
shiny static app3 site --subdir a3
```

This will create the following directory structure:

```
site
├── a1
│   ├── app.json
│   ├── edit
│   │   └── index.html
│   └── index.html
├── a2
│   ├── app.json
│   ├── edit
│   │   └── index.html
│   └── index.html
├── a3
│   ├── app.json
│   ├── edit
│   │   └── index.html
│   └── index.html
├── serviceworker.js
└── shinylive     # Other JS stuff in here, including pyodide
```

All of the deployed apps share the same shinylive resources.

Note that if you want one to be deployed to the top level, it can be done by just not using `--subdir`:

```py
shiny static app1 site
shiny static app2 site --subdir a2
```

```
site
├── app.json   # Top level application, from source dir app1/
├── edit
│   └── index.html
├── index.html
├── a2
│   ├── app.json
│   ├── edit
│   │   └── index.html
│   └── index.html
├── serviceworker.js
└── shinylive
```

*****

`shiny static` now by default only includes packages needed to run the deployed application(s). Previously, a deployment would be about 200MB because it included all of the packages from Pyodide, as well as a few other shiny-related packages. With the change in rstudio/py-shinylive#14, a minimal shinylive app deployment is now about 27MB.

If deployed this way, then when a user goes to the `/edit/` page and makes changes or runs code in the terminal, they will not be able to load any extra packages that normally are part of the distribution (they will be able to load packages from PyPI, but as always, these packages must not contained any compiled code).

In order to include the full Shinylive distribution, the option `--full-shinylive` can be used.

### DIFF
--- a/shiny/_static.py
+++ b/shiny/_static.py
@@ -1,12 +1,10 @@
-import base64
-import json
+import importlib.util
 import os
 import re
 import shutil
 import sys
 from pathlib import Path
-from typing import Callable, List, Optional, Union
-
+from typing import List, Optional, Union
 
 if sys.version_info >= (3, 8):
     from typing import Literal, TypedDict
@@ -14,7 +12,7 @@ else:
     from typing_extensions import Literal, TypedDict
 
 _SHINYLIVE_DOWNLOAD_URL = "https://pyshiny.netlify.app/shinylive"
-_SHINYLIVE_DEFAULT_VERSION = "0.0.1dev"
+_SHINYLIVE_DEFAULT_VERSION = "0.0.2dev"
 
 # This is the same as the FileContentJson type in TypeScript.
 class FileContentJson(TypedDict):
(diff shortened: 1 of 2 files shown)


shiny/_static.py
```diff
@@ -24,208 +22,38 @@ class FileContentJson(TypedDict):
 
 
 def deploy_static(
-    appdir: Union[str, Path],
+    appdirs: Tuple[Union[str, Path]],
```
jcheng5, line 25: Is this type annotation right? Doesn’t this restrict you to a tuple
of length 1?

wch: Good catch, thanks.

PR #314, opened 2022-08-14
Plot using pyplot global figure

### LINKED ISSUE #312: Support `matplotlib.pyplot` interface (as opposed to matplotlib's object-oriented interface)
Hello py-shiny team,

Awesome work porting `shiny` to `python`. I would like to know if it is possible to directly call `@render.plot` on a plot object directly. For instance  for https://github.com/rstudio/py-shiny/issues/211, instead of 

```
  @output()
    @render.plot(alt="A histogram")
    def plot():
        np.random.seed(19680801)
        x = 100 + 15 * np.random.randn(437)
        fig, ax = plt.subplots()
        plt.tight_layout()
        ax.hist(x, input.n(), density=True)
        return fig
```

I would like to be able to directly do the following 

```
  @output()
    @render.plot(alt="A histogram")
    def plot():
        np.random.seed(19680801)
        x = 100 + 15 * np.random.randn(437)
        return plt.hist(x, input.n(), density=True)

```

Currently, this will fail with

>@render.plot doesn't know to render objects of type '<class 'tuple'>'. Consider either requesting support for this type of plot object, and/or explictly saving the object to a (png) file and using @render.image

Thank you and sorry if this has already been asked elsewhere. I could not readily find it.

NelsonGon

jcheng5:
Fixes #312.

Turns out it is pretty easy to get at the pyplot global figure. However, this is a bad idea if you are using async, or multiple threads (we don't currently allow this but we might later on).

### TODO

- [x] Throw error if global pyplot figure is used and user function is async

### QA notes

I added an example at `examples/global_pyplot/app.py`, will write tests against it when we have that infrastructure ready.

### DIFF
--- a/examples/global_pyplot/app.py
+++ b/examples/global_pyplot/app.py
@@ -1,0 +1,18 @@
+import matplotlib.pyplot as plt
+from shiny import App, Inputs, Outputs, Session, render, ui
+
+app_ui = ui.page_fluid(
+    ui.input_checkbox("render", "Render", value=True),
+    ui.output_plot("mpl"),
+)
+
+
+def server(input: Inputs, output: Outputs, session: Session):
+    @output
+    @render.plot
+    def mpl():
+        if input.render():
+            plt.hist([1, 1, 2, 3, 5])
+
+
+app = App(app_ui, server)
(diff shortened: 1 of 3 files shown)


jcheng5:
Hmmm, I wonder why pyright wasn't failing for me locally.

wch (approved):

wch:
I just clicked the "Approve" button but then realized that there are some type-checking issues. I'll investigate.

wch:
It looks like the reason that pyright behaves differently from pylance is because pylance ships with some type stubs and pyright does not.

https://github.com/microsoft/pylance-release/issues/659

The pylance type stubs are available in this repository:
https://github.com/microsoft/python-type-stubs

PR #320, opened 2022-08-24
End-to-end testing infrastructure

### LINKED ISSUE #324: End-to-end testing infrastructure
Use Playwright to run end-to-end tests of Shiny apps, akin to how we use shinytest for Shiny for R.

* Tests should be runnable both locally and over GHA
* Multiple browsers supported
* Separate invocation of unit tests vs. e2e tests

jcheng5:
Fixes #324

This PR adds `make e2e`, which runs a series of automated end-to-end tests using the [Playwright for Python](https://playwright.dev/python/docs/intro) framework.

The tests exist under `./e2e/*/test_*.py` files, which test against either e2e-test-specific `app.py` files in those same directories, or `./examples/*/app.py` example apps.

## TODO

- [ ] ~Consider renaming `e2e` to `tests-e2e`~ I'm not going to do this, doesn't feel worth the time


e2e/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+The Python files directly in this subdirectory are for Pytest fixtures and helper code
+to make test writing easier. (Eventually this logic may move to the `shiny` package
+itself or its own dedicated package, so that Shiny app authors can set up their own e2e
+tests against their apps.)
+
+The actual tests are in subdirectories. Each subdirectory contains one or more Pytest
+files (`test_*.py`) containing [Playwright](https://playwright.dev/python/) assertions,
+and optionally, a single app (`app.py`) that the assertions test against. (The app is
+optional, because the tests may also be for apps in the `examples/` directory.)
```
cpsievert, line 13:
```suggestion
optional, because the tests may also be for apps in the `examples/` or `shiny/examples/` directory.)
```

e2e/README.md
```diff
@@ -0,0 +1,81 @@
...
+tox e2e/async
+
+# Run just the tests in e2e/async/, in headed mode
+tox -- --headed e2e/async
+```
+
+## Shiny app fixtures
+
+Playwright for Python launches and controls (headless or headful) browsers, but does not
+know how to start/stop Shiny app processes. Insteaed, we have our own [Pytest
```
cpsievert, line 35:
```suggestion
know how to start/stop Shiny app processes. Instead, we have our own [Pytest
```

e2e/README.md
```diff
@@ -0,0 +1,81 @@
...
+fixtures](https://docs.pytest.org/en/latest/explanation/fixtures.html) for
+Shiny app process management.
+
+### Testing a local `app.py`
+
+The `local_app: ShinyAppProc` fixture launches the `app.py` in the current directory
+as the calling `test_*.py` file.
+
+```python
+# pyright: reportUnknownMemberType=false
```
cpsievert, line 45: Since this won't be needed in a future version

```suggestion
```

e2e/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+import re
+from playwright.sync_api import Page, expect
+from conftest import ShinyAppProc
```
cpsievert, line 49: Would it be more clear where this import is coming from if it isn't was `from .conftest import ShinyAppProc`?

cpsievert, line 49: I'm not sure what the full implication of a relative import might be here, but to me it makes it more clear where the conftest module lives? 

```suggestion
from .conftest import ShinyAppProc
```

e2e/README.md
```diff
@@ -0,0 +1,81 @@
...
+### Testing an app from `../examples`
+
+It's also possible to test apps that live in the `../examples` directory: at the test
+module level, create a fixture with the help of `conftest.create_example_fixture`, and
+use it from test funcs.
+
+```python
+# See https://github.com/microsoft/playwright-python/issues/1532
+# pyright: reportUnknownMemberType=false
+
```
cpsievert, line 67:
```suggestion
```

e2e/conftest.py
```diff
@@ -0,0 +1,208 @@
...
+def random_port():
+    while True:
+        port = random.randint(1024, 49151)
+        with socket.socket() as s:
+            try:
+                s.bind(("127.0.0.1", port))
+                return port
+            except Exception:
+                # Let's just assume that port was in use; try again
+                continue
```
cpsievert, line 37: Could merge #329 first and get rid of this?

e2e/conftest.py
```diff
@@ -0,0 +1,208 @@
...
+            while not self._io.closed:
+                try:
+                    line = self._io.readline()
+                except ValueError:
+                    # This is raised when the stream is closed
+                    break
+                if line is None:
+                    break
+                if line != "":
+                    # logging.warning(f"> {line}")
```
cpsievert, line 69: Not sure why this is here?

```suggestion
```

e2e/README.md
jcheng5: `conftest` wouldn't be in the same directory as a test, but in the parent directory. So it'd be `..conftest`. But I'm not sure that would work because the e2e stuff isn't in a package.

jcheng5: Yeah, just tried it, it indeed does not work.

e2e/conftest.py
jcheng5: Dang it, I _knew_ I had written this somewhere already! 😆

e2e/README.md
jcheng5: Let me implement this real quick...

e2e/conftest.py
```diff
@@ -0,0 +1,208 @@
...
+        ):
+            return
+        else:
+            raise RuntimeError("Shiny app exited without ever becoming ready")
+
+
+def run_shiny_app(
+    app_file: Union[str, PurePath],
+    *,
+    port: Optional[int] = None,
```
cpsievert, line 152: For consistency with #329...would also require changing line 159 below

```suggestion
    port: int = 0,
```

e2e/conftest.py
```diff
@@ -0,0 +1,208 @@
...
+        [sys.executable, "-m", "shiny", "run", "--port", str(port), str(app_file)],
+        bufsize=bufsize,
+        executable=sys.executable,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        cwd=cwd,
+        encoding="utf-8",
+    )
+
+    # TODO: If wait_for_start, ensure the app is listening before returning
```
cpsievert, line 171: IIUC this is already happening (below on 177)?

```suggestion
```

e2e/controls.py
```diff
@@ -0,0 +1,116 @@
...
+    @property
+    def expect_start(self):
+        return expect(self.loc.locator("input:first-of-type"))
+
+    @property
+    def expect_end(self):
+        return expect(self.loc.locator("input:last-of-type"))
+
+
+class SliderInput:
```
cpsievert, line 59: How come there is no `expect` property here? Is it because the markup is different  depending on whether `value` is of length 1 or 2?

e2e/controls.py
```diff
@@ -0,0 +1,116 @@
...
+    def __init__(self, page: Page, inputId: str):
+        self.loc = page.locator(f"#{inputId}.js-range-slider").locator("xpath=..")
+
+    def move_slider(self, fraction: float) -> None:
```
cpsievert, line 63: Perhaps this could be done in a less brittle way by calling `setValue` on the input binding (and/or working with ion range slider API more directly)?

e2e/controls.py
jcheng5: I was just creating those on-demand. Especially because I just didn't think about it very much as I was creating controls.py. Barret is going to do a more thoughtful buildout of controls.py.

e2e/controls.py
jcheng5: I specifically wanted our end-to-end tests to work "like a user". It might also be useful to have a way to do `setValue`, but the purpose for that would be to test that slider's `setValue` works, whereas `move_slider` tests that slider's UI works.

e2e/conftest.py
jcheng5: Done!

cpsievert (approved):

e2e/conftest.py
jcheng5: I'll unify them afterwards just so we can get this in now.

PR #321, opened 2022-08-25
Improve use of pyright

jcheng5:
* In a few places, we were using pyright comments that affected the
  entire file instead of just single lines as was intended
* Use pandas and matplotlib type stubs instead of ignoring all those issues

## Testing notes

No need for testing besides CI, this just affects type checking


wch:
Closed in favor of #613.

PR #338, opened 2022-09-28
NavSet's `tagify()` method now returns tagified contents

### LINKED ISSUE #337: `ui.nav()` doesn't handle `Tagifiable` input correctly
Discovered via https://github.com/rstudio/py-shiny/issues/332. Here is a more minimal demonstration of  the problem:

```py
from shiny import ui

class Foo:
    def tagify(self):
        return ui.div()

res = ui.navset_tab(ui.nav("a", Foo()))

print(ui.TagList(res).render())
```

```py
  File "<stdin>", line 1, in <module>
  File "/Users/cpsievert/github/py-htmltools/htmltools/_core.py", line 226, in render
    return {"dependencies": deps, "html": cp.get_html_string()}
  File "/Users/cpsievert/github/py-htmltools/htmltools/_core.py", line 250, in get_html_string
    html_ += line_prefix + child.get_html_string(indent, eol)
  File "/Users/cpsievert/github/py-htmltools/htmltools/_core.py", line 576, in get_html_string
    html_ += self.children.get_html_string(
  File "/Users/cpsievert/github/py-htmltools/htmltools/_core.py", line 250, in get_html_string
    html_ += line_prefix + child.get_html_string(indent, eol)
  File "/Users/cpsievert/github/py-htmltools/htmltools/_core.py", line 576, in get_html_string
    html_ += self.children.get_html_string(
  File "/Users/cpsievert/github/py-htmltools/htmltools/_core.py", line 254, in get_html_string
    raise RuntimeError(
RuntimeError: Encountered a non-tagified object. x.tagify() must be called before x.render()
```

cpsievert:
Closes #337

Note that this is needed since `TagList`'s `tagify()` method doesn't recursively call `tagify()` (maybe it should, @wch?)

https://github.com/rstudio/py-htmltools/blob/ae298449b10de5a193031d7c72d7216f7611cf5a/htmltools/_core.py#L172-L194

### DIFF
--- a/shiny/ui/_navs.py
+++ b/shiny/ui/_navs.py
@@ -369,7 +369,7 @@ class NavSet:
         nav, content = render_navset(
             *self.args, ul_class=ul_class, id=id, selected=self.selected, context={}
         )
-        return self.layout(nav, content)
+        return self.layout(nav.tagify(), content.tagify())
 
     def layout(self, nav: TagChildArg, content: TagChildArg) -> Union[TagList, Tag]:
         return TagList(nav, self.header, content, self.footer)
--- a/shiny/ui/_navs.py
+++ b/shiny/ui/_navs.py
@@ -369,7 +369,7 @@ class NavSet:
         nav, content = render_navset(
             *self.args, ul_class=ul_class, id=id, selected=self.selected, context={}
         )
-        return self.layout(nav.tagify(), content.tagify())
+        return self.layout(nav, content).tagify()
 
     def layout(self, nav: TagChildArg, content: TagChildArg) -> Union[TagList, Tag]:
         return TagList(nav, self.header, content, self.footer)
--- a/tests/test_navs.py
+++ b/tests/test_navs.py
@@ -173,3 +173,27 @@ def test_nav_markup():
           <div class="row">Page footer</div>
         </div>"""
     )
+
+
+def test_nav_tagify():
+    class Foo:
+        def tagify(self):
+            return ui.span()
+
+    x = with_private_seed(ui.navset_tab, ui.nav("a", ui.div(Foo())))
+
+    assert TagList(x).render()["html"] == textwrap.dedent(
+        """\
+        <ul class="nav nav-tabs" data-tabsetid="7311">
+          <li class="nav-item">
+            <a data-bs-toggle="tab" data-toggle="tab" data-value="a" role="tab" class="nav-link active" href="#tab-7311-0">a</a>
+          </li>
+        </ul>
+        <div class="tab-content" data-tabsetid="7311">
+          <div class="tab-pane active" role="tabpanel" data-value="a" id="tab-7311-0">
+            <div>
+              <span></span>
+            </div>
+          </div>
+        </div>"""
+    )


wch:
I think that the current behavior for `TagList`'s `tagify()` is correct -- I think it makes sense that if an object's `tagify()` method is called, it is responsible for return a valid tag tree, and this may involve calling `tagify()` on its children. But I'm happy to discuss if something about that doesn't seem right.

cpsievert:
I can get on board with leaving it as is for the sake of keeping our code simple, but on the other hand, the issue was pretty difficult to debug/identify, and it seems like it could be a very easy/common mistake to make. Also, FWIW, I'm pretty sure `htmltools:::tagify()` doesn't suffer from this same problem (it recurses into the results of `as.tags()`)

wch:
After thinking about it some more, I'm feel more strongly that the current behavior is correct. If `x` is a `Tag` object, you'd expect to be able to call `x.tagify()` and let that method handle the recursion, instead of having to recurse manually. The same should be true if `x` is a `TagList` -- it would be odd to call `y.tagify()` on each child node and then after that go into each child node and call `z.tagify()` on each of _its_ children.

On a related note, it might make sense for a `TagList` to not have a method named `.tagify()`, but instead `.tagify_children()` (this is related to a discussion with @jcheng5 on Slack).

jcheng5:
The reason for not wanting `TagList` to have `tagify`, is because when it _does_ have `tagify`, it means that `TagList` is an instance of `TagChild` (because the `Tagifiable` protocol is one of the union elements of `TagChild`). And as Winston explained it to me, we explicitly do not want `TagList` to be an instance of `TagChild`, as `TagChild` is for representing tag hierarchies post-flattening (not 100% sure about this last part).

jcheng5:
If we want to be as strict as @wch is suggesting, I think it's important that `get_html_string` strictly check that it only encounters the exact kinds of objects it expects (`Tag`, `MetadataNode`, and `str`) and throw on anything else.

But also, @wch, it feels like the responsibility to implement recursive `tagify` perfectly is pretty annoying to live up to, no? The type checker doesn't help you because 1) the return type of `tagify` allows `Tagifiable` objects, and 2) you have to remember to make a copy of any `MetadataNode` descendants (if [this comment](https://github.com/rstudio/py-htmltools/blob/ae298449b10de5a193031d7c72d7216f7611cf5a/htmltools/_core.py#L77-L79) is accurate).

wch:
> But also, @wch, it feels like the responsibility to implement recursive `tagify` perfectly is pretty annoying to live up to, no? The type checker doesn't help you because 1) the return type of `tagify` allows `Tagifiable` objects, and 2) you have to remember to make a copy of any `MetadataNode` descendants (if [this comment](https://github.com/rstudio/py-htmltools/blob/ae298449b10de5a193031d7c72d7216f7611cf5a/htmltools/_core.py#L77-L79) is accurate).

For the sake of argument, let's say that `TagList` has a `tagify_children()` method (instead of `tagify()`). Also keep in mind that each `Tag` object contains a `TagList`, and each `TagList` can contain any number of `Tag` or `Tagifiable` objects. (Set aside the `MetadataNode` and `str` for now.)

So the code looks roughly like this:

```py
class Tag:
    def tagify(self) -> Tag:
        cp = copy(self)
        cp.children = cp.children.tagify()  # Note that cp.children is a TagList
        return cp
       
class TagList(List[Union[Tag, Tagifiable]]):    
    def tagify_children(self) -> TagList:
        cp = copy(self)
        for i in range(len(cp)):
                cp[i] = cp[i].tagify()
        return cp
```

Notice that in `TagList.tagify_children`, it calls `tagify()` on each child, and that's all it needs to do.

If we were to do the thing where we call `tagify()` on a child, and then go into _its_ children and call `tagify()` on them, then that would be kind of weird. In most cases -- when the result of calling `tagify()` on a child is a `Tag`, then we'd end up calling `tagify()` on the grandchildren twice. Basically, if we say that an object's `tagify()` method is sometimes responsible for recursing into its children (as is the case with `Tag`), and sometimes not (as might be the case with `NavSet`), then we're going to end up with a confusing mess.

I wonder if it's possible to get the type system to model the two kinds of trees we're interested in, without going into really awkward contortions:
* One containing `Tag` and `Tagifiable` objects.
* One containing just `Tag` objects. This is what the other kind of tree gets transformed into after `tagify()` is called on it.

We should at the very least have a run time check like the one @jcheng5 suggested for `get_html_string`.

jcheng5:
@wch I think you may be thinking of `tagify` differently than @cpsievert and I? Like for you it’s basically a pre-rendering step, whereas to us it’s more like “type conversion to Tag/TagList”.

wch:
I think we're in agreement that `tagify()` is for type conversion to `Tag`/`TagList`. In practice we use it during pre-rendering, but I wouldn't say that's the only place where it's valid to call it.

schloerke (changes requested):

schloerke:
I don't want to close this PR due to the discussion above. But the PR as be merged in to #580 and will be merged to `main`

cpsievert:
Seems like we can close #337 then @schloerke?

schloerke:
> I think we're in agreement that `tagify()` is for type conversion to `Tag`/`TagList`. In practice we use it during pre-rendering, but I wouldn't say that's the only place where it's valid to call it. 

Yes. Sorry I missed this last night.

PR #342, opened 2022-10-04
Improve CI: Add isort, move flake8 to pre-commit, redefined `make check` command.

fpgmaas:
This MR aims to improve developer experience by not having to run multiple commands locally to verify code formatting. Right now, `make check`, `make lint` and optionally `pre-commit run -a` need to be run separately. This MR proposes to have `make check` verify all aspects of code quality. Next to that, it proposes to introduce `isort` to improve code quality.

- Add [isort](https://github.com/PyCQA/isort) in `pre-commit-config.yaml` to sort imports.
- Fix `black` URL in `pre-commit-config.yaml`.
- Move `flake8` to `pre-commit-config.yaml`. 
- Removed `black` from `requirements-dev.txt`, since it is not needed; it's already installed in the `pre-commit` environment. Removed `flake8` as well, since that is moved to `pre-commit-config.yaml`.
- Repurposed the `make check` command to not only check type hints, but actually perform all code quality checks.
- Adapted the GitHub workflow accordingly.
- Removed the `check-old` command, since it does not seem to be used anymore.


wch:
Thanks for the PR. Some of these things seem like definite improvements, but others seem like personal workflow preferences.

Some comments:
* I'd prefer not to move the formatting and linting completely into `pre-commit`. Our VS Code config file uses `black` for formatting on save, and `flake8` for linting in the editor, so it would be best to keep them in the `requirements-dev.txt`.
* It would be nice to have VS Code also run `isort` on save, so it might be good to add it to `requirements-dev.txt` also.
* I think it still makes sense to have separate make targets for pyright and flake8.

Can you make a separate PR that just has the config file changes (and not the .py file changes)? I saw at least one change to a .py file that was potentially dangerous.

Makefile
```diff
@@ -51,14 +51,12 @@ clean-test: ## remove test and coverage artifacts
 typings/uvicorn/__init__.pyi:
 	pyright --createstub uvicorn
 
-check: typings/uvicorn/__init__.pyi ## type check with pyright
+check:
+	echo "Checking types with pyright"
+	$(MAKE) typings/uvicorn/__init__.pyi
```
wch, line 56: Why not keep this file as a dependency for the make target?

setup.cfg
```diff
@@ -68,11 +68,13 @@ console_scripts =
...
 # E302: Expected 2 blank lines
 # E501: Line too long
+# F401: imported but unused. pyright already checks this.
 # F403: 'from module import *' used; unable to detect undefined names
 # F405: Name may be undefined, or defined from star imports
 # W503: Line break occurred before a binary operator
-ignore = E302, E501, F403, F405, W503
+ignore = E302, E501, F401, F403, F405, W503
 exclude = docs, .venv
+max-line-length = 127
```
wch, line 77: Can we get rid of this? It seems like it's redundant with `E501`.

fpgmaas: Will remove it. I moved it from the `Makefile` to the configuration file, originally it's from [here](https://github.com/rstudio/py-shiny/blob/main/Makefile#L61).

Makefile
fpgmaas: Good point, will fix this!

fpgmaas:
@wch  Thanks for your feedback! I will work on an updated MR soon. I will (initially) create the MR without the `.py` files modified. However, to eventually merge the MR, I think they should be modified as well, to pass the CI/CD. Which potentially breaking change did you see?

I personally think `pre-commit` would be a better solution than relying on IDE configuration, but I agree that that's partially personal preference. To lower the bar for contributors, I do suggest creating a `make check` command that runs `black`, `isort`, `flake8` and `pyright`., similar to what I did here. The reason for that is that otherwise people who do not have the possibility to run these on file saves will need to run four manual commands before committing. We can of course also keep separate make commands for them, and just add a fifth that calls the other four. 

Curious for your thoughts! Let me know and I'll work on an updated MR tomorrow.

fpgmaas:
Closed in favor of https://github.com/rstudio/py-shiny/pull/347

shiny/_fileupload.py
```diff
@@ -52,7 +52,8 @@ def file_begin(self) -> None:
         file_info["datapath"] = os.path.join(
             self._dir, str(self._n_uploaded) + file_ext
         )
-        self._current_file_obj = open(file_info["datapath"], "ab")
+        with open(file_info["datapath"], "ab") as f:
+            self._current_file_obj = f
```
wch, line 56: @fpgmaas This is the change that looked dangerous. I think it'll close the file handle after it exits the `with`, so `self._current_file_obj` will not be usable.

fpgmaas: @wch My bad; this is not a change `isort` made, but a change resulting by me adding a `flake8` extension and not properly solving the issue.. I have opened a MR that only shows the difference in non-`.py` files. When I have received feedback there and I have processed the feedback, I will add a single commit with only the output of `isort .` applied.

PR #345, opened 2022-10-04
Change the use of coverage to pytest-cov

fpgmaas:
From the [pytest-cov](https://pypi.org/project/pytest-cov/) website:

---

This plugin produces coverage reports. Compared to just using coverage run this plugin does some extras:

- Subprocess support: you can fork or run stuff in a subprocess and will get covered without any fuss.
- Xdist support: you can use all of pytest-xdist’s features and still get coverage.
- Consistent pytest behavior. If you run coverage run -m pytest you will have slightly different sys.path (CWD will be in it, unlike when running pytest).

---

Functionality of `make coverage` is still exactly the same. It seems `coverage` was missing from the development requirements, but I added `pytest-cov` now. I also modified the ` make test` command to output an `xml` file, in preparation for eventually adding coverage checks to the CI/CD pipeline.


wch:
Sorry for the very long delay. If you can sign the CLA, then I'll merge. (Even if you signed the CLA in the past, you may have to sign it again because we updated it to say Posit instead of RStudio.)

PR #346, opened 2022-10-04
Fixed `_installed_shinylive_versions` and added unit tests

fpgmaas:
Contents of this MR:

- Fixed `_installed_shinylive_versions` which seemed to be broken, since `shinylive_dir.iterdir()` will return the directories including the root, e.g. `shinylive/shinylive-1` and thus the regex `re.sub("^shinylive-", "", str(s))` did not work. 
- Added unit tests to increase code coverage for this file from `32%` to `82%`.
- Fixed typing; the function `remove_shinylive_local` is only ever called with a `Path` as argument, so changed `Union[str, Path, None]` to `Optional[Path]`.

### DIFF
--- a/shiny/_static.py
+++ b/shiny/_static.py
@@ -2,11 +2,11 @@ import os
 import re
 import shutil
 from pathlib import Path
-from typing import List, Optional, Union
+from typing import List, Optional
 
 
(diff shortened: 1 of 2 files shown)


wch:
Thanks!

Issue #335, opened 2022-09-27
Feature Request:  Expose uvicorn "--workers" as a top-level parameter to shiny::run()


bschulth:
- Consider exposing uvicorn `--workers`
   -  https://www.uvicorn.org/settings/#production

- as a top-level parameter to `run()`
   - https://github.com/rstudio/py-shiny/blob/main/shiny/_main.py#L106

- to more easily set the level of concurrency. By default, I think uvicorn starts single-threaded with a single worker.

jcheng5:
Unfortunately, setting `--workers` to >1 can't work with Shiny; Uvicorn's approach to multi-worker does not allow for sticky load balancing, which Shiny requires. It will _appear_ to work but certain features that require XHR/HTTP calls back to the server will randomly fail (for example, `input_file` or `session.download`).

bschulth:
Ah, bummer, I was hoping Python/uvicorn workers was going to overcome the R-single-threaded-behavior and give massive concurrency out of the box, but I get the sticky-session problem.

Perhaps consider checking if the environment variable 'WEB_CONCURRENCY' is set and either fail miserably or issue a stern warning?

I was setting this environment variable to enable this feature for testing concurrent web request s...which for my simple tests appeared to work (as you indicated it might)

Thanks again, great work!

jcheng5:
Thank you for the suggestion, I didn’t realize that `WEB_CONCURRENCY` was a thing!

PR #350, opened 2022-10-05
Add --reload-dir command line option

jcheng5:
This is helpful when working on a package that extends Shiny; you can have the Shiny app reload whenever the package changes.

Also warns if `WEB_CONCURRENCY` is set to a value other than "1" (see #335).


nealrichardson:
Does #353 supercede this?

wch:
Closing in favor of #353.

PR #355, opened 2022-10-20
Playwright e2e tests

MadhulikaTanuboddi:
This PR is an attempt to adding playwright tests for some of our example apps located at either https://github.com/rstudio/py-shiny/tree/main/examples or https://github.com/rstudio/py-shiny/tree/main/shiny/examples

TODO:
- [ ] Work on ToDo s from the tests/controls.py
- [x] Remove non-test based apps 
- [ ] Any Clean up
- [ ] Implement PR feedback

Appreciate any feedback meantime.


e2e/controls.py
```diff
@@ -87,6 +87,58 @@ def move_slider(self, fraction: float) -> None:
         )
         mouse.up()
 
+    def move_slider_range(self, fractionFrom: float, fractionTo: float) -> None:
```
jcheng5, line 90: There's lots of duplicated logic between `move_slider` and `move_slider_range`. It would be great to reduce this duplication by introducing reusable functions. Break it down into, "what sub-problems are being solved inside these functions?" and introduce individual functions for each of those sub-problems.

I have some ideas but I would like to see what you come up with first, then we can discuss!

e2e/controls.py
```diff
@@ -114,3 +166,128 @@ def locate_by_label(self, label: str) -> Locator:
...
+
+        self.loc.locator(f".selectize-dropdown-content .option[data-value='{value}']").click()
+
+        self.loc.locator(f".selectize-input .item[data-value='{value}']")
+
+        # click outside the dropdown box to close the dropdown menu
+        self.page.keyboard.press("Escape")
+
+    #TODO
+    #def remove_option(self):
```
jcheng5, line 210: I think so far we've mostly stuck to simulating things the user can do--removing an option is not one of these, right? (Unless you mean, deselecting one of the options that is currently selected?)

e2e/controls.py
```diff
@@ -114,3 +166,128 @@ def locate_by_label(self, label: str) -> Locator:
...
+
+
+class FileInput():
+    def __init__(self, page: Page, inputId: str):
+        self.loc = page.locator(f"//*[@id='{inputId}']//parent::div[contains(@class, 'shiny-input-container')]")
+
+    def upload_file(self, fileName: str) -> None:
+        browse = self.loc.locator("//span[contains(@class, 'btn-file')]")
+        #TODO: Enhancement: Check PurePath option to upload files
+        browse.set_input_files(files=[f"e2e/data-files/{fileName}"])
```
jcheng5, line 220: This feels too specific--I think `upload_file` should take `filenames: list[str]` as a parameter and pass it straight through to `set_input_files`, and make it the responsibility of the caller to point to e2e/data-files/ if that is what they want to do.

e2e/controls.py
```diff
@@ -114,3 +166,128 @@ def locate_by_label(self, label: str) -> Locator:
...
+
+        # click outside the dropdown box to close the dropdown menu
+        self.page.keyboard.press("Escape")
+
+    #TODO
+    #def remove_option(self):
+
+
+class FileInput():
+    def __init__(self, page: Page, inputId: str):
```
jcheng5, line 214: We should change all camelCase param names to snake_case--in this case it should be `id` to match `ui.input_file`'s `id` parameter (I know I started the `inputId` thing, sorry)

e2e/controls.py
```diff
@@ -114,3 +166,128 @@ def locate_by_label(self, label: str) -> Locator:
...
+
+# Shiny Outputs
+class TextOutput():
+    def __init__(self, page: Page, inputId: str):
+        self.loc = page.locator(f"#{inputId}.shiny-text-output.shiny-bound-output")
+
+    def get_text(self) -> str:
+        return self.loc.inner_text()
+
+    def contains_digit(self, text: str) -> bool:
```
jcheng5, line 230: This doesn't belong on the TextOutput class, since it doesn't relate to the TextOutput instance. Also, this searches to see if _any_ character in the text is a digit, is this really what you want? (versus testing to see if the entire string is some kind of number?)

e2e/events/test_events_app.py
```diff
@@ -0,0 +1,45 @@
...
+
+def test_event(page: Page, event_app: ShinyAppProc):
+    page.goto(event_app.url)
+
+    # sync
+    click_me_sync = ActionButton(page, "btn")
+    click_me_sync.loc.click()
+
+    output1 = UIOutput(page, "btn_value")
+    assert output1.get_text() == "1"
```
jcheng5, line 16: @schloerke and I realized this week that `assert output1.get_text() == "1"` is not the same as `expect(output1.loc).to_have_text("1")`. In the latter case, Playwright will keep retrying the assertion for up to 5 seconds (by default) until it passes; in the former case, it will only try one. See https://playwright.dev/python/docs/test-assertions

I think this greatly affects the way we need to write these controls, any approach where we retrieve a value out of the control and do assertions on it will likely be flawed.

e2e/controls.py
MadhulikaTanuboddi: Correct, I meant deselecting selected options (sorry for the confusion)

e2e/controls.py
MadhulikaTanuboddi: no problem. I will fix these.

e2e/controls.py
MadhulikaTanuboddi: Thanks for bringing this up. Yes, my thought was to check that the textoutput shows the "Latitude" and "Longitude" of the ipyleaflet map. Test app using this function is https://github.com/rstudio/py-shiny/pull/355/files#diff-592926b6d35d2782ae85d50863febd3c90255d3da656ea16db327ecc8b9481e1R38.

So, I am checking if that string contains atleast one digit. we are good to believe that map coordinates are shown in the output.as 

However, I know this is not a good approach and has a TODO to fix. I will think about what you suggested and implement this in a way that we dont have to use `assert` also as you mentioned in the later comments.

e2e/events/test_events_app.py
MadhulikaTanuboddi: Definitely. I learned more about this and agree `expect` is the more stable way to do these checks. I will update any tests that use `assert`. Thanks again for all the feedback.

wch:
Closing because e2e tests were implemented elsewhere.

PR #362, opened 2022-10-25
Moved requirements-dev.txt to setup.cfg

fpgmaas:
This simplifies the repository, and makes it adhere to the standards from [setuptools](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html) for configuration using `setup.cfg`

- Removes the need for a separate `requirements-dev.txt` file
- Allows the user to install the environment for development with one instead of two commands.


jcheng5:
Not a reason that I’m aware of. Thanks!

wch (approved):

PR #366, opened 2022-10-26
Stop using chunked encoding

jcheng5:
When using --reload, the HTTP response for the homepage was using Transfer-Encoding: chunked, which is more complicated for proxies than it needs to be. This was happening because we had a script autoinjection middleware that was modifying the body and then just stripping the Content-Length header instead of making it the correct value.

### DIFF
--- a/shiny/_autoreload.py
+++ b/shiny/_autoreload.py
@@ -113,7 +113,7 @@ class InjectAutoreloadMiddleware:
         if scope["type"] != "http" or scope["path"] != "/" or len(self.script) == 0:
             return await self.app(scope, receive, send)
 
-        def mangle_callback(body: bytes):
+        def mangle_callback(body: bytes) -> Tuple[bytes, bool]:
             if b"</head>" in body:
                 return (body.replace(b"</head>", self.script, 1), True)
             else:
@@ -220,7 +220,7 @@ class ResponseMangler:
 
     def __init__(
         self, send: ASGISendCallable, mangler: Callable[[bytes], Tuple[bytes, bool]]
-    ):
+    ) -> None:
         # The underlying ASGI send function
         self._send = send
         # The caller-provided logic for rewriting the body. Takes a single `bytes`
@@ -236,7 +236,7 @@ class ResponseMangler:
         # rewritten before we send it
         self._response_start: Optional[HTTPResponseStartEvent] = None
         # All the response body bytes we have seen so far
-        self._body = b""
+        self._body: bytes = b""
 
     async def send(self, event: ASGISendEvent) -> None:
         if self._done:
--- a/shiny/_autoreload.py
+++ b/shiny/_autoreload.py
@@ -6,13 +6,14 @@ import os
 import secrets
 import threading
 import webbrowser
-from typing import Optional, Tuple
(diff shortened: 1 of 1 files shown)


shiny/_autoreload.py
```diff
@@ -111,40 +112,15 @@ async def __call__(
...
-                        # The entire response was seen, and we never encountered
-                        # any </head>. Just send everything we have
-                        event["body"] = body
-                        body = b""  # Allow gc
-
-            return await send(event)
-
-        await self.app(scope, receive, rewrite_send)
+
+        def mangle_callback(body: bytes):
```
cpsievert, line 116: Should this have type annotations on the return value?

shiny/_autoreload.py
```diff
@@ -231,3 +207,97 @@ async def process_request(
...
+    over ASGI. This would be easy if not for 1) response bodies are potentially sent in
+    chunks, over multiple events; 2) the first response event we receive is the one that
+    contains the Content-Length, which can be affected when we do rewriting later on.
+    The ResponseMangler handles the buffering and content-length rewriting, leaving the
+    caller to only have to worry about the actual body-modifying logic.
+    """
+
+    def __init__(
+        self, send: ASGISendCallable, mangler: Callable[[bytes], Tuple[bytes, bool]]
+    ):
```
cpsievert, line 223:
```suggestion
    ) -> None:
```

shiny/_autoreload.py
```diff
@@ -231,3 +207,97 @@ async def process_request(
...
+        self._mangler = mangler
+
+        # If True, the mangler is done and any further data can simply be passed along
+        self._done: bool = False
+
+        # Holds the http.response.start event, which may need its Content-Length header
+        # rewritten before we send it
+        self._response_start: Optional[HTTPResponseStartEvent] = None
+        # All the response body bytes we have seen so far
+        self._body = b""
```
cpsievert, line 239:
```suggestion
        self._body: bytes = b""
```

shiny/_autoreload.py
```diff
@@ -231,3 +207,97 @@ async def process_request(
...
+
+            new_len = len(self._body)
+            if new_len != old_len:
+                # The mangling check changed the length of the body. Add the difference
+                # to the content-length header (if content-length is even present)
+                _add_to_content_length(self._response_start, new_len - old_len)
+
+            more_body = "more_body" in event and event["more_body"]
+
+            if done or not more_body:
```
cpsievert, line 271:
```suggestion
            more_body = event.get("more_body", None)

            if done or more_body is None:
```

jcheng5: It's possible for `more_body` to be present but `False`--I'll take the rest of your suggestion though, thanks!

cpsievert:
Just curious, how did this come up (feel free to Slack me the answer)

shiny/_autoreload.py
jcheng5, line 116:
```suggestion
        def mangle_callback(body: bytes) -> Tuple[bytes, bool]:
```

cpsievert (approved):
LGTM besides the minor nitpicks

PR #374, opened 2022-11-16
Add e2e testing methods

schloerke:
Add end-to-end testing control classes. These classes are essentially wrappers around `Locator` methods from `playwright`. 

Control classes have support for UI methods:

```
* input_action_button
* input_action_link
* input_checkbox
* input_checkbox_group
* input_date
* input_date_range
* input_file
* input_numeric
* input_password
* input_radio_buttons
* input_select
* input_selectize
* input_slider
* input_switch
* input_text
* input_text_area
* output_image
* output_plot
* output_table
* output_text
* output_text_verbatim
* output_ui
```

The control class names have been CamelCased for conformity.

-----------------

Things to remember:

* `.value()` **should not be trusted** as it will return the _current_ value. It will not wait for currently-being-processed-by-shiny value to be retrieved. Because of this, these methods are dangerous. Only use `Locator` objects.


e2e/playground.py
```diff
@@ -0,0 +1,413 @@
...
+            page,
+            id=id,
+            loc=f"input#{id}[type=number].shiny-bound-input",
+        )
+
+        # Must be last
+        if verify:
+            self.verify()
+
+    def verify(self):
```
jcheng5, line 198: Remove verify() methods

e2e/playground.py
```diff
@@ -0,0 +1,2076 @@
...
+
+
+def maybe_cast_attr(
+    fn: typing.Callable[[typing.Any], R],
+    loc: Locator,
+    attr_name: str,
+    timeout: Timeout = None,
+) -> typing.Union[R, None]:
+    ret = loc.get_attribute(attr_name, timeout=timeout)
+    if not isinstance(ret, type(None)):
```
jcheng5, line 92:
```suggestion
    if ret is not None:
```

e2e/playground.py
```diff
@@ -0,0 +1,2076 @@
...
+
+
+def expect_attr(
+    loc: Locator,
+    name: str,
+    value: typing.Union[AttrValue, None],
+    timeout: Timeout = None,
+):
+    """Expect an attribute to have a value. If `value` is `None`, then the attribute should not exist."""
+    if isinstance(value, type(None)):
```
jcheng5, line 122:
```suggestion
    if value is None:
```

e2e/playground.py
```diff
@@ -0,0 +1,2076 @@
...
+    playwright_expect(loc).to_have_attribute(name=name, value=value, timeout=timeout)
+
+
+def get_el_style(
+    loc: Locator,
+    css_key: str,
+    timeout: Timeout = None,
+):
+    ret = loc.get_attribute("style", timeout=timeout) or ""
+    m = re.search(css_key + r":\s*([^\s;]+)", ret)
```
jcheng5, line 139: Do more precise match on `css_key` (`\b`)

jcheng5: Also, regex-escape `css_key`

e2e/playground.py
```diff
@@ -0,0 +1,2076 @@
...
+        return
+
+    playwright_expect(loc).to_have_attribute(name=name, value=value, timeout=timeout)
+
+
+def get_el_style(
+    loc: Locator,
+    css_key: str,
+    timeout: Timeout = None,
+):
```
jcheng5, line 137: Return type annotations (throughout file)

e2e/playground.py
```diff
@@ -0,0 +1,2076 @@
...
+
+# from abc import ABC
+
+
+# class InputA(ABC):
+#     @abstractmethod
+#     def my_abstract_method(self, arg1):
+#         pass
+
+
```
jcheng5, line 211:
```suggestion
```

e2e/playground.py
```diff
@@ -0,0 +1,2076 @@
...
+    @property
+    def expect(self):
+        return playwright_expect(self.loc)
+
+    # # Requires a PR to playwright to call `obj.__expect__()` method; Desired API
+    # def __expect__(self) -> LocatorAssertions:
+    #     return playwright_expect(self.loc)
+    def foo(self, x: int, y: int, z: int):
+        return x + y + z
```
jcheng5, line 240:
```suggestion
```

e2e/playground.py
```diff
@@ -0,0 +1,2076 @@
...
+        )
+
+        if isinstance(loc_label, str):
+            loc_label = self.loc_container.locator(loc_label)
+        self.loc_label = loc_label
+
+    def value_label(self, *, timeout: Timeout = None) -> typing.Union[str, None]:
+        return self.loc_label.text_content(timeout=timeout)
+
+    def expect_label_to_have_text(self, value: str, *, timeout: Timeout = None):
```
jcheng5, line 304: Allow value to also be a pattern

e2e/playground.py
```diff
@@ -0,0 +1,2076 @@
...
+#     * Locators will stay as properties
+#   * Don't sub-class. For now, use `_` separatation and use `loc` or `value` as a prefix
+# * Approach
+#   * Use locators / playwright_expect as much as possible
+#     * It should not be necessary to use `assert` directly.
+#     * MUST wait for `Locator`s to do their job
+#   * Provide `value` methods as a convenience
+
+
+class _WidthLoc:
```
jcheng5, line 321: Consider adding a suffix to Mixins indicating they are mixins; and put a comment in here somewhere that says what mixins are/what their rules are

jcheng5, line 321: Should `_WidthLoc` inherit from `_InputBaseP`?

e2e/playground.py
```diff
@@ -0,0 +1,2076 @@
...
+        return float_attr(self.loc, "max", timeout=timeout)
+
+    def value_step(self, *, timeout: Timeout = None) -> typing.Union[float, None]:
+        return float_attr(self.loc, "step", timeout=timeout)
+
+    def expect_value(self, value: str, *, timeout: Timeout = None):
+        self.expect.to_have_value(value, timeout=timeout)
+
+    def expect_min_to_have_value(self, value: AttrValue, *, timeout: Timeout = None):
+        assert type(value) is not None
```
jcheng5, line 403: Can be None. (Same for next method)

e2e/playground.py
```diff
@@ -0,0 +1,2076 @@
...
+    # step: Optional[float] = None,
+    # width: Optional[str] = None,
+    def __init__(self, page: Page, id: str):
+        super().__init__(
+            page,
+            id=id,
+            loc=f"input#{id}[type=number].shiny-bound-input",
+        )
+
+    def set(self, value: float, *, timeout: Timeout = None):
```
jcheng5, line 372: Allow value to be None (the `input_numeric` function should do that too)

e2e/playground.py
```diff
@@ -0,0 +1,2076 @@
...
+
+    def value(self, *, timeout: Timeout = None) -> str:
+        return self.loc.input_value(timeout=timeout)
+
+    def expect_value(self, value: str, *, timeout: Timeout = None):
+        self.expect.to_have_value(value, timeout=timeout)
+
+
+class InputPassword(
+    _Placeholder,
```
jcheng5, line 503: Add _WidthLoc

e2e/playground.py
```diff
@@ -0,0 +1,2076 @@
...
+
+        self.loc_label = self.loc_container.locator("label")
+
+    def set(self, value: str, *, timeout: Timeout = None):
+        self.loc.fill(str(value), timeout=timeout)
+
+    def value(self, *, timeout: Timeout = None) -> str:
+        return self.loc.input_value(timeout=timeout)
+
+    def expect_value(self, value: str, *, timeout: Timeout = None):
```
jcheng5, line 498: Value could also be pattern or None

e2e/playground.py
```diff
@@ -0,0 +1,2052 @@
...
+    def expect_width_to_have_value(
+        self: _InputBaseP,
+        value: typing.Union[AttrValue, None],
+        *,
+        timeout: Timeout = None,
+    ):
+        expect_attr(self.loc, "width", value=value, timeout=timeout)
+
+
+class _WidthContainer:
```
schloerke, line 314: Can the mixin classes inherit from abstract meta classes to prevent init and property values?

e2e/playground.py
schloerke: 71e58e7

e2e/playground.py
schloerke: When defined as `class _WidthLoc(_InputBaseP):`, we get the error:

```
Class derives from one or more protocol classes but does not implement all required members
  Member "id" is declared in protocol class "_InputBaseP"
  Member "loc" is declared in protocol class "_InputBaseP"
  Member "page" is declared in protocol class "_InputBaseP"
```

We do not want to add these properties and would go against our goal of inheriting from an ABC.

Keeping the code as is.

e2e/playground.py
schloerke: 20deb85

e2e/playground.py
schloerke: 828ae7e

e2e/playground.py
schloerke: ee4ae48 and 331266d

e2e/playground.py
schloerke: ee4ae48

e2e/playground.py
schloerke: a48f920

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+            page,
+            id=id,
+            loc=f"select#{id}.shiny-bound-input{select_class}",
+            loc_label=f"label[for='{id}']",
+        )
+        # `loc_container` does not need to contain selected items
+        # But we should have `self.loc` be `self.loc_selected`
+        self.loc = self.loc_container.locator("option:checked")
+        # Same value
+        self.loc_selected = self.loc
```
jcheng5, line 693: `self.loc` should be the `<select>` itself, not the selected choices (which will be `self.loc_selected`).

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+            id=id,
+            select_class="",
+        )
+
+    # selectize: bool = False,
+    def expect_selectize(self, selectize: bool, *, timeout: Timeout = None) -> None:
+        # class_=None if selectize else "form-select",
+        ex_selectize = playwright_expect(self.loc_select)
+        if selectize:
+            ex_selectize.not_to_have_class(re.compile("form-select"), timeout=timeout)
```
jcheng5, line 844: These should be strings, not regexes

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+            id=id,
+            loc=f"a#{id}.action-button.shiny-bound-input",
+        )
+
+
+# * click:
+#     * input_checkbox_group
+#     * input_radio_buttons
+
+
```
jcheng5, line 920: Left off here

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+class InputSwitch(InputCheckboxBase):
+    def __init__(
+        self,
+        page: Page,
+        id: str,
+    ) -> None:
+        super().__init__(
+            page,
+            id=id,
+            loc=f"input#{id}[type=checkbox].shiny-bound-input",
```
jcheng5, line 975: Add switch-specific CSS class here?

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+
+            has_locator = self.page.locator(f"{el_type}{is_checked_str}")
+            # Go up one element as CSS does not have "self" selector
+            # We can leverage this as the HTML structure is very nested.
+            # This approach will break if there are matching siblings
+            has_locator = has_locator.nth(i).locator("..")
+            # Find direct child that matches selector
+            has_locator = has_locator.locator(
+                f"> {el_type}[{key}='{item}']{is_checked_str}"
+            )
```
jcheng5, line 1043:
```suggestion
            has_locator = has_locator.locator(
                f"[{key}='{item}']{is_checked_str}"
            )
```

jcheng5, line 1043:
```suggestion
            has_locator = self.page.locator(f"{el_type}{is_checked_str}")
            has_locator = has_locator.nth(i)
            has_locator = has_locator.locator(
                "xpath=.[@{key}='{item}']"
            )
```

jcheng5: Also need to xpath-escape `{item}`

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        if inline:
+            playwright_expect(self.loc_container).to_have_class(
+                "shiny-input-container-inline", timeout=timeout
+            )
+        else:
+            playwright_expect(self.loc_container).not_to_have_class(
+                "shiny-input-container-inline", timeout=timeout
+            )
+
+    def _expect_locator_values_in_list(
```
jcheng5, line 1003: This could be a standalone utility function

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+            page,
+            id=id,
+            loc="label input[type=checkbox]",
+            loc_container=f"div#{id}.shiny-input-checkboxgroup.shiny-bound-input",
+            loc_label=f"label#{id}-label",
+        )
+
+        # `loc_container` does not need to contain checked items
+        # But we should have `self.loc` be `self.loc_selected`
+        self.loc = self.loc_container.locator("label input[type=checkbox]:checked")
```
jcheng5, line 1080: Would like `self.loc` to be the `.shiny-bound-input` element.

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        self,
+        selected: typing.Union[str, typing.List[str]],
+        *,
+        timeout: Timeout = None,
+        **kwargs: typing.Any,
+    ) -> None:
+
+        if isinstance(selected, str):
+            selected = [selected]
+        # We are wanting to delay retriving the value of the checkbox as long as possible
```
jcheng5, line 1100:
```suggestion
        # We are wanting to delay retrieving the value of the checkbox as long as possible
```

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        checkbox_loc = self.loc_choices
+        checkbox_loc.nth(0).wait_for(state="attached", timeout=timeout)
```
jcheng5, line 1102: Wait for container instead of `nth(0)` choice

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+
+        # TODO-barret; Could do with multiple locator calls, but unchecking the values without the known values would be difficult.
```
jcheng5, line 1104: It would be great to have a variant of `expect_choices()` to be called here, that ensures that all the choices you want to set/unset are actually present.

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+
+            loc_container = loc_container.locator(
+                # Return self
+                "xpath=.",
+                has=has_locator,
+            )
+        # Make sure other items are not in set
+        loc_inputs = loc_container.locator(f"{el_type}{is_checked_str}")
+        # TODO-barret; Look into adding a try-catch around this and then performing the locator check using multiple expectations to get better error messages
+        playwright_expect(loc_inputs).to_have_count(len(arr), timeout=timeout)
```
jcheng5, line 1053: This will fail if `arr` is empty. Need some earlier logic to prevent us from getting here.

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+    def set(
+        self,
+        selected: str,
+        *,
+        timeout: Timeout = None,
+        **kwargs: typing.Any,
+    ) -> None:
+
+        self.loc_container.locator(
+            f"label input[type=radio][value='{selected}']"
```
jcheng5, line 1189: `{selected}` needs escaping

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        timeout: Timeout = None,
+    ) -> None:
+        # Playwright doesn't like lists of size 0. Instead, use `None`
+        if selected is None:
+            playwright_expect(self.loc_selected).to_have_count(0, timeout=timeout)
+            return
+
+        playwright_expect(self.loc_selected).to_have_value(selected, timeout=timeout)
+
+
```
jcheng5, line 1219: Left off here

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+    def expect_files(
+        self,
+        files: typing.Union[str, typing.List[str], None],
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        # TODO-barret; Find method to get files, or remove method
+        # TODO-barret; Test value being sent to shiny?
+        NotImplementedError("`expect_files()` is not implemented")
```
jcheng5, line 1274: Change to `expect_complete` or whatever. See https://github.com/rstudio/shiny/blob/e7d62f55ca93225e6837bfca00c5bdc11aa74f99/srcts/src/file/fileProcessor.ts#L97

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+            str,
+            pathlib.Path,
+            FilePayload,
+            typing.List[typing.Union[str, pathlib.Path]],
+            typing.List[FilePayload],
+        ],
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        self.loc.set_input_files(file_path, timeout=timeout)
```
jcheng5, line 1264: By default, call `self.expect_complete`. This will probably want 1) a different timeout than `timeout`, and 2) to have a way to opt out.

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        super().__init__(
+            page,
+            id=id,
+            loc=f"input#{id}",
+            loc_label=f"label#{id}-label",
+        )
+
+    def expect_value(self, value: AttrValue, *, timeout: Timeout = None) -> None:
+        # TODO-barret; implement
+        NotImplementedError("Need to get the value somehow")
```
jcheng5, line 1347: This could just be the handle text to find the value; if we consider this to be just "what does the user see with their eyeballs". Also note that range sliders have two handles.

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        if grid_bb is None:
+            raise RuntimeError("Couldn't find bounding box for .irs-grid")
+
+        mouse = self.loc_container.page.mouse
+        mouse.move(handle_center[0], handle_center[1])
+        mouse.down()
+        mouse.move(
+            grid_bb.get("x") + (fraction * grid_bb.get("width")), handle_center[1]
+        )
+        mouse.up()
```
jcheng5, line 1377: To implement "set value", simulate what the user would do: move the mouse and look at the handle value to see if it matches the desired value. (Ideally we'd do this with a single mouse.down()/mouse.up(), just lots of mouse.move())

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        interval_str = str(interval) if interval else None
+        expect_attr(
+            self.loc_container.locator(".slider-animate-container a"),
+            "data-interval",
+            interval_str,
+            timeout=timeout,
+        )
```
jcheng5, line 1413: Add method to click the play/pause button

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+    ) -> None:
+        expect_attr(self.loc, "capture", capture, timeout=timeout)
+
+    def expect_placeholder_to_have_value(
+        self, value: AttrValue, *, timeout: Timeout = None
+    ) -> None:
+        expect_attr(self.loc_file_display, "placeholder", value=value, timeout=timeout)
+
+
+class InputSlider(_WidthLoc, _InputWithLabel):
```
jcheng5, line 1316: Need range slider functionality (maybe as a separate class?)

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+    # min: Optional[Union[date, str]] = None,
+    # max: Optional[Union[date, str]] = None,
+    # format: str = "yyyy-mm-dd",
+    # startview: str = "month",
+    # weekstart: int = 0,
+    # language: str = "en",
+    # width: Optional[str] = None,
+    # autoclose: bool = True,
+    # datesdisabled: Optional[List[str]] = None,
+    # daysofweekdisabled: Optional[List[int]] = None,
```
jcheng5, line 1507: Add init method?

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        value: typing.Union[datetime.date, str, None],
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        value_str = _date_str(value)
+        if value_str is None:
+            self.loc.fill("", timeout=timeout)
+            return
+
+        self.loc.fill(value_str, timeout=timeout)
```
jcheng5, line 1519: Consider using https://playwright.dev/docs/api/class-keyboard

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        # TODO-barret; How to trigger the update without opening the date picker?
+        self.loc.evaluate('(el) => $(el).bsDatepicker("update");')
+
+    def expect_value(
+        self,
+        value: typing.Union[datetime.date, AttrValue],
```
jcheng5, line 1525: Due to `language` parameter of `dateInput`, don't allow `datetime.date`, instead have the caller provide the exact string

jcheng5: Same for the rest of the class

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+    ) -> None:
+        # Not using `_date_str()` as we want ability to supply non ISO formatted dates
+        if isinstance(value, datetime.date):
+            value = str(value)
+
+        if value is None:
+            self.expect.to_be_empty(timeout=timeout)
+        else:
+            self.expect.to_have_value(value, timeout=timeout)
+
```
jcheng5, line 1537: Left off here

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+    def expect_autoclose(
+        self,
+        # TODO-barret; None value supported?
+        value: typing.Union[AttrValue, bool],
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        if isinstance(value, bool):
+            value = str(value).lower()
+        expect_attr(self.loc, "data-date-autoclose", value=str(value), timeout=timeout)
```
jcheng5, line 1598: Using string value instead

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        return None
+    elif isinstance(date, typing.Pattern):
+        return date.pattern
+    elif isinstance(date, datetime.date):
+        return str(date)
+    else:
+        return str(datetime.date.fromisoformat(date))
+
+
+class _DateBase(_WidthContainer, _InputWithLabel):
```
jcheng5, line 1494: `_DateBase` should be a mixin that doesn't have `_WidthContainer` and `_InputWithLabel`, since its use in `DateRangeInput` doesn't have either of those. Introduce a second subclass for use within `DateRangeInput`.

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+            loc_container=self.loc_container,
+        )
+
+    def set(
+        self,
+        # TODO-barret; Should this be a list or a tuple?
+        value: typing.Union[
+            typing.Tuple[
+                typing.Union[datetime.date, str, None],
+                typing.Union[datetime.date, str, None],
```
jcheng5, line 1698: 1. Don't support `datetime.date` (because of the language parameter).
2. Make `None` a no-op.

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+            ],
+            None,
```
jcheng5, line 1700: Don't allow `None`.

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        self.date_start.set(value=value[0], timeout=timeout)
+        self.date_end.set(value=value[1], timeout=timeout)
+
+    def expect_value(
+        self,
+        # TODO-barret; Should this be a list or a tuple?
+        value: typing.Union[
+            typing.Tuple[
+                typing.Union[datetime.date, PatternOrStr, None],
+                typing.Union[datetime.date, PatternOrStr, None],
```
jcheng5, line 1716: Remove `datetime.date`

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+            ],
+            None,
```
jcheng5, line 1718: Remove this too

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        if start_val is None:
+            start_val = ""
+        if end_val is None:
+            end_val = ""
+
+        # We can not use `[value={value}]` within Locators.
+        # The physical `value` attribute is never set, so we can not select on it.
+        # We must as the start and end values individually, rather than at the same time like the checkboxgroup input.
+        self.date_start.expect_value(start_val, timeout=timeout)
+        self.date_end.expect_value(end_val, timeout=timeout)
```
jcheng5, line 1744: Add note that these two lines should be atomic (when we do the project to have composable assertions)

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        #     has=self.page.locator(f"input:nth-child(3)[type=text][value={end_val}]"),
+        # )
+        # playwright_expect(loc_dates).to_have_count(
+        #     int(not (value is None)),
+        #     timeout=timeout,
+        # )
+        # if False:
+        #     # TODO-future?; This should be in a try catch around the item above for better diagnostics
+        #     self.date_start.expect_value(start_val, timeout=timeout)
+        #     self.date_end.expect_value(end_val, timeout=timeout)
```
jcheng5, line 1762: Delete

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+    # min: Optional[Union[date, str]] = None,
+    def expect_min_date(
+        self,
+        value: typing.Union[datetime.date, AttrValue],
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        # These values should be the same, so checking both independently seems fair
+        self.date_start.expect_min_date(value, timeout=timeout)
+        self.date_end.expect_min_date(value, timeout=timeout)
```
jcheng5, line 1773: Also composable

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        if value is None:
+            value = ""
+        playwright_expect(self.loc_separator).to_have_text(value, timeout=timeout)
+
+    # width: Optional[str] = None,
+
+    # autoclose: bool = True,
+    def expect_autoclose(
+        self,
+        value: bool,
```
jcheng5, line 1846: Make this match the single-date input version

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+
+
+class _OutputBaseP(Protocol):
+    id: str
+    loc: Locator
+
+
+class _OutputBase:
+    id: str
+    loc: Locator
```
jcheng5, line 1867: Add `page: Page`

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+
+
+######################################################
+# # Outputs
+######################################################
+
+
+class _OutputBaseP(Protocol):
+    id: str
+    loc: Locator
```
jcheng5, line 1862: Add `page: Page`?

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+            loc = page.locator(loc)
+        self.loc = loc
+
+    @property
+    # TODO; Return type
+    def expect(self):
+        return playwright_expect(self.loc)
+
+
+class _OutputTextValue(_OutputBase):
```
jcheng5, line 1889: We should be able to run assertions for:

0. It has not received its initial value
1. It's showing a valid output result
    1. Optionally, pattern or string match on the result
2. It's showing an error
    1. Optionally, pattern or string match on the result
3. It's in progress

0, 2, and 3 should be added to `OutputBase` because they don't vary.

(This is another candidate for composing assertions)

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+class _OutputContainerP(_OutputBaseP, Protocol):
+    expect_container_tag: typing.Callable[
+        # [
+        #     _OutputBaseP,
+        #     typing.Union[Literal["span", "div"], str],
+        #     Timeout,
+        # ],
+        ...,  # TODO-barret; Can't get this to work
+        None,
+    ]
```
jcheng5, line 1916:
```suggestion
    def expect_container_tag(
        self: _OutputBaseP,
        tag_name: typing.Union[Literal["span", "div"], str],
        *,
        timeout: Timeout = None,
    ) -> None:
        ...
```

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        tag_name: typing.Union[Literal["span", "div"], str],
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        # Could not find an expect method to find the tag name
+        # So trying to perform the expectation manually by waiting for attached state,
+        # then asserting
+
+        # Make sure the tag exists
+        self.loc.wait_for(state="attached", timeout=timeout)
```
jcheng5, line 1931: Maybe this???
```suggestion
        self.loc.location(f"xpath=self::{tag_name}").expect_count...
```

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+
+    loc_img: Locator
+
+    def expect_height_to_have_value(
+        self,
+        value: AttrValue,
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        self.expect_inline(inline=value is None, timeout=timeout)
```
jcheng5, line 1996:
```suggestion
```

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        expect_el_style(self.loc, "height", value, timeout=timeout)
+
+    def expect_width_to_have_value(
+        self,
+        value: AttrValue,
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        self.expect_inline(inline=value is None, timeout=timeout)
```
jcheng5, line 2005:
```suggestion
```

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+
+    def expect_img_style(
+        self,
+        value: AttrValue,
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        expect_attr(self.loc_img, "style", value, timeout=timeout)
+
+
```
jcheng5, line 2048: Left off here

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+class OutputUi(_OutputContainer, _OutputBase):
+    # id: str,
+    # inline: bool = False,
+    # container: Optional[TagFunction] = None,
+    def __init__(self, page: Page, id: str) -> None:
+        super().__init__(page, id=id, loc=f"#{id}")
+
+    # TODO-barret; Should we do `expect_html_to_have_value()`?
+    # Thinking they can call `expect(self.loc).to_have_html(value)` directly as Shiny does not own that value, the user does.
```
jcheng5, line 2073: How about a method that asserts that it's not empty? (and succeeded)

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+        self,
+        labels: typing.Union[typing.List[PatternOrStr], None],
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        if isinstance(labels, list) and len(labels) == 0:
+            labels = None
+
+        if labels is None:
+            playwright_expect(self.loc.locator("thead th")).to_have_count(
```
jcheng5, line 2092:
```suggestion
            playwright_expect(self.loc.locator("> thead > th")).to_have_count(
```

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+                0, timeout=timeout
+            )
+        else:
+            playwright_expect(self.loc.locator("thead th")).to_have_text(
```
jcheng5, line 2096:
```suggestion
            playwright_expect(self.loc.locator("> thead > th")).to_have_text(
```

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+    def expect_column_text(
+        self,
+        column: int,
+        # Can't use `None` as we don't know how many rows exist
+        text: typing.List[PatternOrStr],
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        playwright_expect(
+            self.loc.locator(f"tbody tr td:nth-child({column})")
```
jcheng5, line 2109: Add `>` as appropriate

e2e/playground.py
```diff
@@ -0,0 +1,2135 @@
...
+    def expect_n_row(
+        self,
+        n: int,
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        playwright_expect(self.loc.locator("tbody tr")).to_have_count(
+            n,
+            timeout=timeout,
+        )
```
jcheng5, line 2135: `expect_cell_value` (takes row, col number)

e2e/playground.py
schloerke: 87dd73d

e2e/playground.py
schloerke: 64d5694

e2e/playground.py
schloerke: 64d5694

e2e/playground.py
schloerke: 64d5694

e2e/playground.py
schloerke: 64d5694

e2e/playground.py
schloerke: 551a7c4

e2e/playground.py
schloerke: 4316e7f

e2e/playground.py
schloerke: 4316e7f

e2e/playground.py
schloerke: 806a064

e2e/playground.py
schloerke: 806a064

e2e/playground.py
schloerke: 806a064

e2e/playground.py
schloerke: 806a064

e2e/playground.py
schloerke: f03fa74

e2e/playground.py
schloerke: a6a800b

e2e/playground.py
schloerke: They must be regexes, `to_have_class()` compares against the whole `class` value. Sad it does not check for whole class values

schloerke: 4820eb9

e2e/playground.py
schloerke: Even though I requested this, I'm keeping it here as it only pertains to this base class

e2e/playground.py
schloerke: 430d6e4

`.[@foo='bar']` is invalid. Using `self::*[@foo='bar']`. (This took more time, but I've learned a lot more about XPATH.)

e2e/playground.py
schloerke: e712415

e2e/playground.py
schloerke: ec2afc9

e2e/playground.py
schloerke: a739fe3

schloerke: This is a satisfying solution. Much more intuitive

e2e/playground.py
schloerke: a739fe3

schloerke: Resolving this comment as we'll need to make a `InputSliderRange` class which should have two values. 
Implemented: `InputSlider.expect_value(value: PatternOrStr)` using the label above the slider handle

e2e/playground.py
```diff
@@ -782,9 +683,9 @@ def expect_choice_groups(
 
         _RadioButtonCheckboxGroupBase._expect_locator_values_in_list(
             self,
-            "optgroup",
```
jcheng5: Make `_expect_locator_values_in_list` either a static or a function

e2e/playground.py
```diff
@@ -132,6 +168,20 @@ def expect_attr(
...
 
 
+def expect_class_value(
+    loc: Locator,
+    cls: str,
+    has_class: bool,
+    timeout: Timeout = None,
+) -> None:
+    """Expect a locator to have (or not to have) a class value"""
+    cls_regex = re.compile(rf"\b{cls}\b")
```
jcheng5, line 178:
```suggestion
    cls_regex = re.compile(rf"\b{re.escape(cls)}\b")
```

e2e/playground.py
```diff
@@ -991,70 +884,111 @@ def expect_choice_labels(
...
+        is_checked: typing.Union[bool, None],
+    ) -> str:
+        if is_checked is None:
+            return ""
+        elif is_checked:
+            return ":checked"
         else:
-            playwright_expect(self.loc_container).not_to_have_class(
-                "shiny-input-container-inline", timeout=timeout
+            NotImplementedError("`is_checked = FALSE` is not verified yet")
```
jcheng5, line 908:
```suggestion
            raise NotImplementedError("`is_checked = FALSE` is not verified yet")
```

e2e/playground.py
```diff
@@ -991,70 +884,111 @@ def expect_choice_labels(
...
-            # This approach will break if there are matching siblings
-            has_locator = has_locator.nth(i).locator("..")
-            # Find direct child that matches selector
-            has_locator = has_locator.locator(
-                f"> {el_type}[{key}='{item}']{is_checked_str}"
-            )
+            # Get the `n`th matching element
+            has_locator = has_locator.nth(i)
+            # Make sure that element has the correct attribute value
+            has_locator = has_locator.locator(f'xpath=self::*[@{key}="{item}"]')
```
jcheng5, line 974: `key` and `item` need some kind of escaping

e2e/playground.py
```diff
@@ -991,70 +884,111 @@ def expect_choice_labels(
...
+        arr: typing.List[str],
+        *,
+        is_checked: typing.Optional[bool] = None,
+        timeout: Timeout = None,
+        key: str = "value",
+    ) -> None:
+        # Make sure the locator contains all of `arr`
+
+        # Make sure the locator has len(uniq_arr) input elements
+        _RadioButtonCheckboxGroupBase._assert_is_unique(
```
jcheng5, line 925: Same thing as before--make a regular method call or make static/function

e2e/playground.py
```diff
@@ -1161,20 +1102,18 @@ def __init__(
...
-        # `loc_container` does not need to contain checked items
-        # But we should have `self.loc` be `self.loc_selected`
-        self.loc = self.loc_container.locator("label input[type=radio]:checked")
-        # Same value
-        self.loc_selected = self.loc
-
-        self.loc_choices = self.loc_container.locator("label input[type=radio]")
-        self.loc_choice_labels = self.loc_container.locator(
+        self.loc_selected = self.loc.locator("label > input[type=radio]:checked")
+        self.loc_choices = self.loc.locator("label > input[type=radio]")
```
jcheng5, line 1113:
```suggestion
        self.loc_selected = self.loc.locator("> label > input[type=radio]:checked")
        self.loc_choices = self.loc.locator("> label > input[type=radio]")
```

e2e/playground.py
```diff
@@ -1260,19 +1205,20 @@ def set(
...
+    def expect_complete(
         self,
-        files: typing.Union[str, typing.List[str], None],
         *,
         timeout: Timeout = None,
     ) -> None:
-        # TODO-barret; Find method to get files, or remove method
-        # TODO-barret; Test value being sent to shiny?
-        NotImplementedError("`expect_files()` is not implemented")
+        expect_el_style(self.loc_progress, "width", "100%", timeout=timeout)
```
jcheng5, line 1219: Let's make sure that if the upload errors out, `expect_complete()` fails.

e2e/playground.py
```diff
@@ -1341,12 +1291,74 @@ def __init__(
...
+                found = True
+                break
+
+            # Not found; move handle to the right
+            mouse.move(grid_bb["x"] + pxls, handle_center[1])
+            pxls += 1
+
+        mouse.up()
+        if not found:
+            values_found_txt = ", ".join([f"'{key}'" for key in values_found.keys()])
```
jcheng5, line 1355: Limit if `values_found` is a very large list

e2e/playground.py
```diff
@@ -1691,95 +1678,65 @@ def __init__(self, page: Page, id: str) -> None:
...
-        # )
-        # if False:
-        #     # TODO-future?; This should be in a try catch around the item above for better diagnostics
-        #     self.date_start.expect_value(start_val, timeout=timeout)
-        #     self.date_end.expect_value(end_val, timeout=timeout)
+        # TODO-future; Composable expectations
+        if not (start_val is None):
+            self.date_start.expect_value(start_val, timeout=timeout)
+        if not (end_val is None):
+            self.date_end.expect_value(end_val, timeout=timeout)
```
jcheng5, line 1719: Let's revisit the inconsistency between what None means here versus in set_text.

e2e/playground.py
```diff
@@ -2062,23 +2001,42 @@ def __init__(self, page: Page, id: str) -> None:
...
+        text: TextValue,
+        row: int,
+        col: int,
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        if text is None:
+            text = ""
+        playwright_expect(
+            self.loc.locator(f"xpath=./table/tbody/tr[{row}]/td[{col}]")
```
jcheng5, line 2037: Can we match against `th` too, not just `td`?

e2e/playground.py
```diff
@@ -0,0 +1,2093 @@
+"""Barret Facade classes for working with Shiny inputs/outputs in Playwright"""
```
jcheng5, line 1: Rename playground.py to something else

e2e/playground.py
schloerke: Made issue: https://github.com/rstudio/py-shiny/issues/409

Resolving

e2e/kitchen_input_radio_checkbox/test_input_radio_checkboxgroup.py
```diff
@@ -77,3 +77,78 @@ def assert_selected(
...
+        not_exist = InputRadioButtons(page, "does-not-exist")
+        not_exist.expect_choices(["a", "b", "c"], timeout=timeout)
+    except AssertionError as e:
+        assert "expected to have count '1'" in str(e)
+        assert "Actual value: 0" in str(e)
+
+    check1 = InputCheckboxGroup(page, "check1")
+
+    # Make sure it works
+    check1.expect_choices(["red", "green", "blue"])
```
wch, line 98: Make sure it can accept tuples in addition to lists.

e2e/playground.py
```diff
@@ -76,15 +79,43 @@
...
+R = typing.TypeVar("R")
+M1 = typing.TypeVar("M1")
+M2 = typing.TypeVar("M2")
+
+
+def is_missing(x: typing.Any) -> typing.TypeGuard[MISSING_TYPE]:
+    return isinstance(x, MISSING_TYPE)
+
+
+def not_is_missing(x: typing.Union[R, MISSING_TYPE]) -> typing.TypeGuard[R]:
```
wch, line 101: Can you make a comment about the typeguard/narrowing thing that makes this necessary?

e2e/playground.py
```diff
@@ -76,15 +79,43 @@
...
 
 Timeout = typing.Union[float, None]
 InitLocator = typing.Union[Locator, str]
 
+R = typing.TypeVar("R")
+M1 = typing.TypeVar("M1")
+M2 = typing.TypeVar("M2")
+
+
+def is_missing(x: typing.Any) -> typing.TypeGuard[MISSING_TYPE]:
```
wch, line 97: Probably better to use `typing.object` instead of `Any`.

wch: And do the same for the rest of the file.

e2e/playground.py
```diff
@@ -176,24 +154,40 @@ def expect_class_value(
     timeout: Timeout = None,
 ) -> None:
     """Expect a locator to have (or not to have) a class value"""
-    cls_regex = re.compile(rf"\b{cls}\b")
+    cls_regex = re.compile(rf"\b{re.escape(cls)}\b")
```
wch, line 157: If `cls` is `"foo"`, then this regex will match `"foo-bar"`, which is not what you want.

e2e/playground.py
```diff
@@ -1053,15 +1069,13 @@ def __init__(
 
     def set(
         self,
-        selected: typing.Union[str, typing.List[str]],
+        selected: typing.List[PatternOrStr],
```
wch, line 1072: Don't support `Pattern`s in `set` methods.

e2e/kitchen_input_radio_checkbox/test_input_radio_checkboxgroup.py
schloerke: `typing.Tuple` needs a known size. While `typing.Sequence` works for tuples, lists, and range, Playwright only accepts `typing.List[]` objects. So we would be converting them to Lists in all of these places. 

Not pursuing. Resolving comment.

shiny/ui/_input_select.py
```diff
@@ -205,7 +205,7 @@ def _normalize_choices(x: SelectChoicesArg) -> _SelectChoices:
 
 
 def _render_choices(
-    x: _SelectChoices, selected: Optional[str | list[str]] = None
+    x: _SelectChoices | None, selected: Optional[str | list[str]] = None
```
wch, line 208: Revert, and add `# pyright: ignore[whatever]` for line 212

wch: Do same for rest of file.

shiny/ui/_input_text.py
```diff
@@ -19,7 +19,7 @@
 def input_text(
     id: str,
     label: TagChildArg,
-    value: str = "",
+    value: Union[str, None] = "",
```
wch, line 22: Revert

shiny/ui/_input_text.py
```diff
@@ -87,14 +87,14 @@ def input_text(
 def input_text_area(
     id: str,
     label: TagChildArg,
-    value: str = "",
+    value: Union[str, None] = "",
```
wch, line 90: Revert

shiny/examples/input_selectize/app.py
```diff
@@ -5,11 +5,10 @@
...
         "Choose a state:",
         {
-            "East Coast": {"NY": "NY", "NJ": "NJ", "CT": "CT"},
-            "West Coast": {"WA": "WA", "OR": "OR", "CA": "CA"},
-            "Midwest": {"MN": "MN", "WI": "WI", "IA": "IA"},
+            "East Coast": {"NY": "New York", "NJ": "New Jersey", "CT": "Connecticut"},
+            "West Coast": {"WA": "Washington", "OR": "Oregon", "CA": "California"},
+            "Midwest": {"MN": "Minnesota", "WI": "Wisconsin", "IA": "Iowa"},
         },
-        multiple=True,
```
wch, line 12: Revert

e2e/kitchen_input_switch/app.py
```diff
@@ -0,0 +1,15 @@
+from shiny import *
```
wch, line 1: Please add this example to `shiny/examples/`

e2e/input_numeric/test_input-numeric.py
```diff
@@ -9,10 +9,13 @@
 
 
 def test_input_numeric(page: Page, app: ShinyAppProc) -> None:
+    # with page and app:
```
wch, line 12: Rename file to `test_input_numeric.py`, and do same for other files.

wch: And get rid of `kitchen`

e2e/playground.py
```diff
@@ -0,0 +1,2363 @@
...
+    from typing_extensions import Protocol
+
+import typing
+
+from playwright.sync_api import FilePayload, FloatRect, Locator, Page, Position
+from playwright.sync_api import expect as playwright_expect
+
+from shiny.types import MISSING, MISSING_TYPE
+
+"""
```
wch, line 21: Clean up these comments.

e2e/notes.txt
```diff
@@ -0,0 +1,165 @@
+"""Barret Facade classes for working with Shiny inputs/outputs in Playwright"""
```
wch, line 1: Move relevant info to README and delete this file.

Also, add note to README explaining why some input tests have their own directory.

schloerke:
- [ ] Fix failing tests
- [ ] Rename `playground.py` to `conftest.py`. Remove Joe's old file first.
- [ ] Run e2e testing nightly with dev versions
- [ ] Run all shiny/examples apps and make sure no Python or JS errors occur
- [ ] Rename test folders to `inputs`, `output`s. Do not keep `kitchen` name

PR #394, opened 2023-02-08
Use correct argument to input_selectize

jcheng5:
Without this fix, the example throws an error instead of starting.


cpsievert (approved):
Thanks!

PR #395, opened 2023-02-09
Prefer the use of page_fixed over page_fluid

jcheng5:
Most apps look better in page_fixed


jcheng5:
I was wrong about this, the Bootstrap notion of fixed page makes very little difference

Issue #417, opened 2023-03-08
e2e tests don't use currently-installed versions of packages


wch:
In our GHA workflow, we are manually installing htmltools from github, but the e2e tests aren't recognizing that installed copy of htmltools.

Error here:
https://github.com/rstudio/py-shiny/actions/runs/4359881025/jobs/7622174378#step:13:23

Installation of correct version of htmltools here:
https://github.com/rstudio/py-shiny/actions/runs/4359881025/jobs/7622174378#step:5:28

schloerke:
Related #414

schloerke:
@jcheng5 May we remove `tox` as a testing wrapper? If so, we can just use the installed packages with pytest.

I do remember you saying that `tox` can test in multiple platforms, but we have this with GHA testing. So I do not believe the duplication of e2e dependencies and headache of trying to get GitHub dependencies installed within tox.

I've added a `make test-e2e` command in https://github.com/rstudio/py-shiny/pull/418/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52

PR #418, opened 2023-03-08
Consolidate typing extensions imports; Install shinyverse for e2e

### LINKED ISSUE #415: Consolidate all `typing_extensions` imports

schloerke:
Fixes #415 
Related #417


shiny/_typing_extensions.py
```diff
@@ -0,0 +1,41 @@
...
+else:
+    from typing_extensions import TypeGuard
+
+# Even though TypedDict is available in Python 3.8, because it's used with NotRequired,
+# they should both come from the same typing module.
+# https://peps.python.org/pep-0655/#usage-in-python-3-11
+if sys.version_info >= (3, 11):
+    from typing import NotRequired, TypedDict
+else:
+    from typing_extensions import NotRequired, TypedDict
```
schloerke, line 19: All of the pyright problems stem from `NotRequired` not being recognized with `TypedDict`. If I perform this import statement within the destination file (like the original code), it works as expected.

@wch Any idea why this is failing? It'd be nice to not have the version checking throughout the repo and consolidate it all into a single file.

wch: I think the problem may actually be totally different -- the main branch of htmltools has updated types, which would cause failures in CI here. PR #416 in this repo makes use of the new types.

What's surprising to me is that in GHA, it's simply skipping the "Type check with pyright" step on Python 3.10 and 3.11. Now that I look at it, I think it's because in the workflow file, `"3.10"` and `"3.11"` are strings and the numeric comparison won't work for strings.

shiny/_typing_extensions.py
```diff
@@ -0,0 +1,41 @@
...
+
+if sys.version_info >= (3, 11):
+    from typing import assert_type
+else:
+    from typing_extensions import assert_type
+
+if sys.version_info < (3, 10):
```
wch, line 26: Can you reverse the logic here to keep it the same as the others, as in `>= (3, 10)`.

wch: Also, can you consolidate these with the other 3.10 types above. And similarly, consolidate the 3.11 types.

shiny/_typing_extensions.py
schloerke: Pyright just fixed this in https://github.com/microsoft/pyright/commit/0a206e884124e20dd31efd1d2a48573ff9d8f6e0

schloerke: Pylance will hopefully resolve this soon

.github/requirements_gha.txt
```diff
@@ -0,0 +1,3 @@
+https://github.com/rstudio/py-htmltools/tarball/main
+https://github.com/rstudio/py-shinywidgets/tarball/main
+https://github.com/rstudio/py-faicons/tarball/main
```
schloerke, line 3: @wch Should we test against more than just `{htmltools}` github repo?

schloerke:
Test failure is a random e2e test failure. This is unrelated to the PR.

shiny/ui/_modal.py
```diff
@@ -7,19 +7,14 @@
...
-from shiny.types import MISSING, MISSING_TYPE
-
-if sys.version_info >= (3, 8):
-    from typing import Literal
-else:
-    from typing_extensions import Literal
-
 from htmltools import HTML, Tag, TagAttrs, TagAttrValue, TagChild, div, tags
 
+from shiny.types import MISSING, MISSING_TYPE
```
wch, line 14: `from ..types`

.github/workflows/pytest.yaml
```diff
@@ -28,17 +28,9 @@ jobs:
...
-          pip install wheel https://github.com/rstudio/py-htmltools/tarball/main
-
-      # - name: Check out and install fontawesome
-      #   run: |
-      #     cd ..
-      #     git clone https://github.com/rstudio/py-fontawesome.git
-      #     cd py-fontawesome
-      #     pip install -r requirements-dev.txt
-      #     make install
+          pip install wheel -r .github/requirements_gha.txt
```
wch, line 33: I think that using hyphen in the name `requirements-gha.txt` is more in line with naming conventions (like `requirements-dev.txt`).

README.md
```diff
@@ -28,6 +28,12 @@ If you want to do development on Shiny for Python:
 pip install -e ".[dev,docs,test]"
 ```
 
+To install all Shiny related packages from GitHub:
```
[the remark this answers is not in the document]
schloerke: Should I just remove the section? Seems like it'll be more harm than good.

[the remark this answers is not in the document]
wch: And come to think of it, the `.github/requirements_gha.txt` path and filename strongly suggests that the purpose of that requirements file is something else, which would be confusing. I think it's probably best to just give instructions to install htmltools (and maybe faicons) from github, without using the `requirements_gha.txt` file here.

.github/requirements_gha.txt
```diff
@@ -0,0 +1,3 @@
+https://github.com/rstudio/py-htmltools/tarball/main
+https://github.com/rstudio/py-shinywidgets/tarball/main
```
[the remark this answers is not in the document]
schloerke: I could remove `{shinywidgets}`. I don't think it's used in any of the tests. So the bleeding edge of `{shinywidgets}` isn't necessary

.github/workflows/pytest.yaml
```diff
@@ -54,7 +46,7 @@ jobs:
           make test
 
       - name: Type check with pyright
-        if: matrix.python-version >= 3.8 && steps.install.outcome == 'success' && (success() || failure())
+        if: matrix.python-version != '3.7' && steps.install.outcome == 'success' && (success() || failure())
         run: |
           make pyright
```
wch, line 52: Skip e2e tests on release, because we want the release process to be reliable.

README.md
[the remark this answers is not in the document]
schloerke: Removed the section and added the deps into `dev` section.

.github/workflows/pytest.yaml
schloerke: Renamed to `requirements-tox.txt`

.github/requirements_gha.txt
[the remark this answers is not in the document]
schloerke: Moved into standard dev install script: `pip install -e ".[dev]"

.github/requirements_gha.txt
schloerke: Removed shinywidgets

PR #420, opened 2023-03-08
Update shinylive editor CSS class

### LINKED ISSUE #419: Pressing <- or -> keys in live examples in API reference changes the docs page.
To reproduce, go to a page like https://shiny.rstudio.com/py/api/reference/shiny.ui.input_numeric.html

Now put the cursor in the code editor and press `<-` or `->`. Another page will get loaded (instead of the cursor moving)

cpsievert:
Closes #419 (I haven't actually confirmed this fixes the issue, but it should)


wch:
I'm not so sure this will work, since there is a div with class `shinylive-container` on the page.

wch:
I think the actual problem is that we no longer have jQuery, but it's trying to use jQuery to set up the handler.

image

wch:
I made a different fix in 0efd5d2.

PR #422, opened 2023-03-14
feat: Pass options to input_selectize

### LINKED ISSUE #158: `input_selectize()`/`update_selectize()` should allow for custom selectize.js options
And some options need to be literal JS code (in R folks currently do this via `I()`) https://shiny.rstudio.com/articles/selectize.html

GShotwell:
Closes #158

I use the `HTML` class to identify which options need to be rendered directly rather than being passed on to selectize. This is similar to how we use `I()` on the R side but I don't love this interface. It seems better to have two arguments `selectize_options` and `rendered_options` which both take dictionaries.


cpsievert:
I think it'd make sense to introduce a new a new JS class/function (and use that over `HTML`). It's unfortunate, that in R, to pass literal JS to selectize options, you need to know to use `I()`, but if you want to do the same in htmlwidgets, you're supposed to use `htmlwidgets::JS()`

shiny/ui/_input_select.py
```diff
@@ -162,11 +170,17 @@ def input_select(
...
+    if options is not None and selectize is False:
+        raise Exception("Options can only be set when selectize is `True`.")
 
     choices_ = _normalize_choices(choices)
     if selected is None and not multiple:
         selected = _find_first_option(choices_)
 
+    if options is None:
+        options = {}
+
```
nealrichardson, line 182: In R, there's more that happens in the code that handles selectize options: https://github.com/rstudio/shiny/blob/main/R/input-select.R#L199-L231

Should we support that here?

nealrichardson (commented):
Should there also be a test added? In R there is this test, which confirms that you can add pluigins to selectize: https://github.com/rstudio/shiny/blob/main/tests/testthat/test-input-select.R#L37-L45

nealrichardson:
> I think it'd make sense to introduce a new JS class/function (and use that over `HTML`). It's unfortunate, that in R, to pass literal JS to selectize options, you need to know to use `I()`, but if you want to do the same in htmlwidgets, you're supposed to use `htmlwidgets::JS()`

The point of `HTML()` is just to say "don't HTML-escape this, it's fine as is", right? `JS()` would just do the same?

Maybe you want a more generic `safe()` or `no_escape()` function that is essentially an alias for `HTML()`. FWIW Hugo calls this function [safeHTML](https://gohugo.io/functions/safehtml/); others often have an "escape" or "[allow unsafe](https://docs.streamlit.io/library/api-reference/write-magic/st.write)" argument to input functions that defaults to escaped/safe but that can be turned off. 

Regardless, seems like something that can be tackled as a followup and not block this feature from merging.

PR #424, opened 2023-03-17
Remove `tox` from e2e testing

### LINKED ISSUE #417: e2e tests don't use currently-installed versions of packages

schloerke:
Fixes #417 

--------------------------------------------

`tox` was introduced to the repo in #320

A clear article on why `tox` is useful: https://christophergs.com/python/2020/04/12/python-tox-why-use-it-and-tutorial/
tl/dr 
> * Test against different versions of Python (which would have alerted Kyle that the library hadn’t been tested against his install version).
> * Test against different dependency versions
> * Capture and run setup steps/ad hoc commands (which Kyle could have made a mistake on / not known about)
> * Isolate environment variables - By design, tox does not pass any evars from the system. Instead you are asked to explicitly declare them (which would have alerted Kyle to any environment variable requirements).
> * Do all the above across Windows / macOS / Linux (which would have saved Kyle if the issue had been due to the OS)

---------------------------------------------

#417 requests that e2e tests use the locally installed packages.  By definition, this is impossible with `tox`.

`tox` cannot install a dependency from github unless it is in a requirements file. This file is not cached. It feels like a giant hack.

I believe the `tox` advantage is minimal as we:
* Are not testing against multiple python versions locally
* Have only a few extra dependencies for e2e testing
* Do not use Environment variables
* Have consistent scripts that can be run

Disadvantage of not using tox:
* Not being able to cleanly pass arguments into the pytest command. However, this can be done with environment variables
	* This can be mitigated by using an environment variable `FILE`. Ex, test the `e2e/outputs`: `make e2e FILE=e2e/outputs`

### DIFF
--- a/.github/requirements-tox.txt
+++ b/.github/requirements-tox.txt
@@ -1,2 +1,0 @@
-https://github.com/rstudio/py-htmltools/tarball/main
-https://github.com/rstudio/py-faicons/tarball/main
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -67,7 +67,7 @@ jobs:
       - name: Run e2e tests
         if: github.event_name != 'release' && steps.install.outcome == 'success' && (success() || failure())
         run: |
-          make test-e2e
+          make e2e
 
   deploy:
     name: "Deploy to PyPI"
--- a/Makefile
+++ b/Makefile
@@ -77,9 +77,11 @@ test: ## run tests quickly with the default Python
 	python3 tests/asyncio_prevent.py
 	pytest
 
+# Default `FILE` to `e2e` if not specified
+FILE:=e2e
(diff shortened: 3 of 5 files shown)


wch:
Looks good. I'll merge, but a couple of comments:

* The top-level README should have info about how to set up a venv.
* If `make e2e FILE=e2e/async` is too cumbersome, we could just add small shell script that lets us do `e2e e2e/async`.
*

PR #427, opened 2023-03-18
Raise warning if `reactive.Effect` function returns a value

### LINKED ISSUE #425: `reactive.Effect` should warn if function returns a value
For example, if a user writes this:

```py
@reactive.Effect
def foo():
  return 123
```

It will run and the user will not receive any indication that the code is probably not doing what they expect. (It will be flagged by a static type checker like pyright, but many users don't use them.)

This should raise an informative warning.

wch:
Closes #425. 
If a user function passed to `reactive.Effect` returns a value, that is most likely a mistake. This PR causes Shiny to raise a warning in that case. For example:

```py
@reactive.Effect
def _():
    return 1
```

### DIFF
--- a/shiny/reactive/_reactives.py
+++ b/shiny/reactive/_reactives.py
@@ -531,7 +531,16 @@ class Effect_:
         with session_context(self._session):
             try:
                 with ctx():
-                    await self._fn()
+                    res = await self._fn()
+
+                    if res is not None:
+                        warnings.warn(
+                            "reactive.Effect function returned a value, but should return None; "
+                            "reactive.Effects should only be used for side effects. "
+                            "Did you mean to use reactive.Calc instead?",
+                            ReactiveWarning,
+                            stacklevel=2,
+                        )
             except SilentException:
                 # It's OK for SilentException to cause an Effect to stop running
                 pass
--- a/tests/test_reactives.py
+++ b/tests/test_reactives.py
@@ -394,6 +394,27 @@ async def test_isolate_async_prevents_dependency():
 
 
 # ======================================================================
+# Effects warn if user function returns a value
+# ======================================================================
+@pytest.mark.asyncio
+async def test_effect_no_return_value():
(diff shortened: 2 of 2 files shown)


wch:
Note that the warning message should be improved to print the line number in the user code where the return happened. This implementation simply causes Python to print out the source where the warning is raised, which always in the same place in `_reactives.py`.

```
await reactive.flush()
/Users/winston/py-shiny/shiny/reactive/_reactives.py:516: ReactiveWarning: reactive.Effect function returned a value, but should return None; reactive.Effects should only be used for side effects. Did you mean to use reactive.Calc instead?
  await self._run()
```

shiny/reactive/_reactives.py
```diff
@@ -531,7 +531,16 @@ async def _run(self) -> None:
...
+                    res = await self._fn()
+
+                    if res is not None:
+                        warnings.warn(
+                            "reactive.Effect function returned a value, but should return None; "
+                            "reactive.Effects should only be used for side effects. "
+                            "Did you mean to use reactive.Calc instead?",
+                            ReactiveWarning,
+                            stacklevel=2,
+                        )
```
jcheng5, line 543: I like this PR--can you just tell me if the warning here gives any helpful indication as to which reactive.Effect is causing the problem? If not, we could use `inspect.unwrap(foobar._fn).__code__.co_filename` and `.co_firstlineno`

wch: IIRC, the problem is that we don't know how many times the user-defined function has been wrapped, so we don't know how many levels in the stack we need to skip.

For example, if the user function is synchronous, then we wrap it once in an async function. But they may decorated it with `@reactive.event`, which would add another layer.

Now, if the user function threw an exception, I think we would be able to catch that and find the source of it, but that's not the situation here.

nealrichardson:
@jcheng5 @wch is this good to merge?

PR #437, opened 2023-03-31
Allow for htmldeps with no source file location

schloerke:
Related: `Add URL support to HTMLDependency` - https://github.com/rstudio/py-htmltools/pull/53


wch:
How about also bumping the htmltools version dependency?

wch (approved):

PR #441, opened 2023-04-04
Make random port test more robust

schloerke:
Other changes
* Make random port test more robust. 
	* GitHub Desktop uses port `9010`
	* Set the timeout to 2s
	* If a port is busy, shift the testing port range
* Use `random.randint(min, max)` for uniform port
	* Rounding _technically_ isn't uniform. Using a single method is less complicated.

### DIFF
--- a/shiny/_app.py
+++ b/shiny/_app.py
@@ -89,7 +89,7 @@ class App:
     def __init__(
         self,
         ui: Tag | TagList | Callable[[Request], Tag | TagList],
-        server: Optional[Callable[[Inputs, Outputs, Session], None]] = None,
+        server: Optional[Callable[[Inputs, Outputs, Session], None]],
         *,
         static_assets: Optional["str" | "os.PathLike[str]"] = None,
         debug: bool = False,
--- a/shiny/_app.py
+++ b/shiny/_app.py
@@ -89,7 +89,7 @@ class App:
     def __init__(
         self,
         ui: Tag | TagList | Callable[[Request], Tag | TagList],
-        server: Optional[Callable[[Inputs, Outputs, Session], None]],
+        server: Optional[Callable[[Inputs, Outputs, Session], None]] = None,
         *,
         static_assets: Optional["str" | "os.PathLike[str]"] = None,
         debug: bool = False,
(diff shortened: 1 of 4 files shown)


shiny/_app.py
```diff
@@ -89,7 +89,7 @@ def server(input: Inputs, output: Outputs, session: Session):
     def __init__(
         self,
         ui: Tag | TagList | Callable[[Request], Tag | TagList],
-        server: Optional[Callable[[Inputs, Outputs, Session], None]],
+        server: Optional[Callable[[Inputs, Outputs, Session], None]] = None,
```
wch, line 92: I'm not convinced it's a good idea to have `server=None` be the default. That could save a little typing when writing tests for UI generation, but for almost all normal uses, users want a server function, and they'll benefit from an error message if it's missing.

tests/test_utils.py
```diff
@@ -136,15 +136,44 @@ async def mutate_registrations():
     assert cb4.exec_count == 1  # Registered during previous invoke(), was called
 
 
+# Timeout within 2 seconds (not)
```
wch, line 139: What does `(not)` mean here?

tests/test_utils.py
```diff
@@ -136,15 +136,44 @@ async def mutate_registrations():
...
+    # Starting port
+    port = 9001
+    # Test `k` continguous ports
+    k = 10
+    # Number of times to try to find a port range
+    n = 100
+
+    # Find a range of `k` ports that are all available
+    for i in range(n):
+        assert i < n - 1, f"Could not find {k} continguous ports to use for testing"
```
wch, line 153: Would be a bit clearer to have this test after the `for` loop (but with `n` instead of `n-1`, I think)

tests/test_utils.py
schloerke:
```suggestion
# Timeout within 2 seconds
```

tests/test_utils.py
```diff
@@ -136,15 +136,44 @@ async def mutate_registrations():
...
 
+# Timeout within 2 seconds (not)
+@pytest.mark.timeout(2)
 def test_random_port():
     assert random_port(9000, 9000) == 9000
 
+    # Starting port
+    port = 9001
+    # Test `k` continguous ports
+    k = 10
```
wch, line 147: Would be clearer to call this `num_ports` (when reading code below that uses `k`, I had to keep scrolling back up to see what it meant)

wch (approved):

PR #442, opened 2023-04-04
Independent of dependency version, serve it if requested

schloerke:
Even if the htmldependency version is higher or lower, the HTML being sent to the user is requesting THIS dependency. Therefore, it should be available to the user independent of any previous versions of the dependency being served.

`{htmltools}` does de-duplicate dependencies and finds the highest version to return. However, dynamic UI and callable UI do not run through the same filter over time. When using callable UI functions, UI dependencies are reset
on refresh. So if a dependency makes it here, it is not necessarily the highest version served over time but is the highest version for this particular UI. Therefore, serve it must be served.

### DIFF
--- a/shiny/_app.py
+++ b/shiny/_app.py
@@ -334,8 +334,12 @@ class App:
         # the user is requesting THIS dependency. Therefore, it should be available to
         # the user independent of any previous versions of the dependency being served.
 
-        # Note: htmltools does do dependency de-duplication and finds the highest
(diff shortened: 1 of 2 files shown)


schloerke:
Needs #441 to test properly

jcheng5:
I'd like to understand this one, maybe we can talk tomorrow?

jcheng5 (approved):

PR #447, opened 2023-04-11
For streaming downloads, add `Transfer-Encoding: chunked` header

### LINKED ISSUE #443: Documentation: example of failed download actually downloads the file.
Hi and thank you for the package:) 

I have noticed that example 4 ('Failed downloads') on the documentation page of Session.download decorator: https://shiny.rstudio.com/py/api/reference/shiny.Session.download.html#shiny.Session.download
instead of raising an error and not proceeding with downloading anything as described, downloads a "failuretest.txt" file with content "helloAn internal server error occurred". It is inconsistent with the expected behavior... 

I looked into this example since I would like to be able to abort the download in case some condition is not met and proceed with the download if the condition is met, but there seems to be no way to achieve this from the example page.

wch:
Closes #443.

For streaming downloads, adding the header `Transfer-Encoding: chunked` lets the client know that it should keep receiving data until it gets a terminating chunk (with size zero). If the transfer ends before then, then the client knows there was an error.

https://stackoverflow.com/a/40282512/412655

Now with the download test app, an error in a streaming download results in this showing in Chrome:

image

### DIFF
--- a/shiny/session/_session.py
+++ b/shiny/session/_session.py
@@ -516,6 +516,12 @@ class Session(object, metaclass=SessionMeta):
 
                             wrapped_contents = wrap_content_sync()
 
+                        # In streaming downloads, we send a 200 response, but if an
+                        # error occurs in the middle of it, the client needs to know.
+                        # With chunked encoding, the client will know if an error occurs
+                        # if it does not receive a terminating (empty) chunk.
+                        headers["Transfer-Encoding"] = "chunked"
+
                         return StreamingResponse(
                             wrapped_contents,
                             200,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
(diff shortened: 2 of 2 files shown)


shiny/session/_session.py
```diff
@@ -516,6 +516,8 @@ async def wrap_content_sync() -> AsyncIterable[bytes]:
 
                             wrapped_contents = wrap_content_sync()
 
+                        headers["Transfer-Encoding"] = "chunked"
```
nstrayer, line 519: Maybe a comment here due to this being non-obvious enough that it didn't make it into the first edition of the code?

shiny/session/_session.py
```diff
@@ -516,6 +516,8 @@ async def wrap_content_sync() -> AsyncIterable[bytes]:
 
                             wrapped_contents = wrap_content_sync()
```
nstrayer, line 518:
```suggestion
# In the case of large streaming downloads using the "chunked" encoding will properly catch and abort downloads that start but then fail before finishing.
```

nstrayer (commented):
All looks good to me. Not familiar enough with the `_session.py` implementation to comment on anything there, but the use of the header seems appropriate. (Assuming there's no way this would get run on http2?)

PR #460, opened 2023-04-23
`wrap_async()` and `is_async_callable()` now allow functions with args

wch:
The use of `ParamSpec` makes it possible for these functions to work on functions that have arguments, and to pass those arguments types along to the returned function.

### DIFF
--- a/shiny/_utils.py
+++ b/shiny/_utils.py
@@ -154,7 +154,7 @@ P = ParamSpec("P")
 
 
 def wrap_async(
-    fn: Callable[P, Awaitable[T]] | Callable[P, T]
+    fn: Callable[P, T] | Callable[P, Awaitable[T]]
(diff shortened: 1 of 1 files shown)


jcheng5:
LGTM other than the failing CI?

wch:
I think the CI errors are due to some issue with the newest version of pyright. The lines of code referenced in the errors don't even contain `...`, but both of those lines do refer to types from matplotlib.

https://github.com/rstudio/py-shiny/blob/33c1ecf9412f5d55bdabe9ac5fed427a5238c7f3/shiny/render/_coordmap.py#L38

https://github.com/rstudio/py-shiny/blob/33c1ecf9412f5d55bdabe9ac5fed427a5238c7f3/shiny/render/_try_render_plot.py#L77

PR #470, opened 2023-05-02
Remove unused datatable JS library

jcheng5:
Fix https://github.com/rstudio/py-shiny/issues/468


schloerke (changes requested):
Please programmatically remove it similar to near line 33 where other legacy code is removed. 

https://github.com/rstudio/py-shiny/blob/main/scripts/htmlDependencies.R

schloerke (approved):

PR #475, opened 2023-05-05
Dashboard example

jcheng5:


wch:
@jcheng5 Do you want to keep this app?

Screenshot for future reference: 
image

cpsievert:
This doesn't seem worth keeping as a template? FWIW, we have this template now built around the tips data https://shiny.posit.co/py/templates/dashboard-tips/

Issue #477, opened 2023-05-09
How to cite Shiny for Python?


schackartk:
In the spirit of transparency and giving credit where it's due, how should I cite Shiny for Python? A BibTeX file could be very helpful.

cpsievert:
Done in #478 (thanks for the PR!)

PR #478, opened 2023-05-09
chore: Add citation

schackartk:
This pull request provides a brief suggested citation.bib file in response to #477. It may make it easier for users to cite this software.

If this is unwanted, then please feel free to close this request! Otherwise, I would be happy to update the file as preferred.


wch:
Thanks for the PR. I don't know that I should be credited to the exclusion of the rest of the team. Is there a way of crediting it to something like "Shiny Team"?

schackartk:
Of course, I can update the file with any wording you prefer. It is fairly common to use something like "The Shiny development team" (see [pandas](https://github.com/pandas-dev/pandas/blob/main/CITATION.cff)).

wch:
That would be great, thanks.

schackartk:
Hi all,

This was my first pull request suggesting a citation file. I have since been looking around a bit more and working with other projects, and it seems that a "cff" (citation file format) file is [recommended by GitHub](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files) and is probably the better way to go. With that file type, GitHub can automatically generate a BibTeX citation and provide better integration.

I will push a commit converting this file to that format.

As @nealrichardson mentioned, pandas includes the license since that's a field in cff files, so I can include that in the updated file. Any further thoughts about the year as mentioned by @wch?

Thanks for your feedback and patience!

CITATION.cff
```diff
@@ -0,0 +1,10 @@
+cff-version: 1.2.0
+title: "Shiny for Python"
+message: >-
+  If you use this software, please cite it using the metadata from this file.
+type: software
+authors:
+  - name: "The Shiny development team"
+repository-code: "https://github.com/rstudio/py-shiny"
+url: "https://shiny.rstudio.com/py/"
+license: MIT
```
wch, line 10: I've modeled this after the Pandas CITATION.cff file:

```suggestion
message: "If you use this software, please cite it as below."
authors:
  - name: "The Shiny development team"
license: MIT
license-url: "https://github.com/rstudio/py-shiny/blob/main/LICENSE"
repository-code: "https://github.com/rstudio/py-shiny"
type: software
url: "https://shiny.posit.co/py/"
```

wch (commented):
Hi @schackartk, sorry for taking so long to reply. I've added a suggestion which is modeled after the Pandas CITATION file.

PR #489, opened 2023-05-15
Updates for pyright

schloerke:


e2e/controls.py
```diff
@@ -11,7 +11,18 @@
 from playwright.sync_api import FilePayload, FloatRect, Locator, Page, Position
 from playwright.sync_api import expect as playwright_expect
 
-from shiny._typing_extensions import Literal, Protocol, TypeGuard, assert_type
+from shiny._typing_extensions import (
+    Literal,  # pyright: ignore[reportPrivateImportUsage]
```
GShotwell, line 15: Can we include a comment in this file that explains why we have that ignored?

GShotwell (approved):

PR #506, opened 2023-05-23
Fix e2e-examples

schloerke:
Introduced in #505

[Current error](https://github.com/rstudio/py-shiny/actions/runs/5050502625/jobs/9061290514?pr=505) that I can't reproduce locally:
```
INTERNALERROR> E             SystemError: AST constructor recursion depth mismatch (before=141, after=191)
INTERNALERROR> E           assert False
```

### DIFF
--- a/shiny/_main.py
+++ b/shiny/_main.py
@@ -221,11 +221,6 @@ def run_app(
 
     log_config: dict[str, Any] = copy.deepcopy(uvicorn.config.LOGGING_CONFIG)
 
-    if reload and app_dir is not None:
-        reload_dirs = [app_dir]
-    else:
-        reload_dirs = []
-
     if reload:
         if autoreload_port == 0:
             autoreload_port = _utils.random_port(host=host)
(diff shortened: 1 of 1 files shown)


schloerke:
The single failure in 3.8 is a false-positive. #507 should address this random behavior

shiny/_main.py
```diff
@@ -238,6 +233,18 @@ def run_app(
...
+        {}
+        if reload is False
+        else {
+            "reload": reload,
+            # Adding `reload_includes` param while `reload=False` produces an warning
+            # https://github.com/encode/uvicorn/blob/d43afed1cfa018a85c83094da8a2dd29f656d676/uvicorn/config.py#L298-L304
+            "reload_includes": ["*.py", "*.css", "*.js", "*.html"],
+            "reload_dirs": [app_dir] if reload and app_dir is not None else [],
+        }
+    )
```
wch, line 246: Python's inside-out syntax for inline `if-else` statements is hard to follow. Can you change it to something like this?

```suggestion
    reload_args: dict[str, bool | str | list[str]] = {}
    if reload:
        reload_dirs = []
        if app_dir is not None:
            reload_dirs = [app_dir]

        reload_args = {
            "reload": reload,
            # Adding `reload_includes` param while `reload=False` produces an warning
            # https://github.com/encode/uvicorn/blob/d43afed1cfa018a85c83094da8a2dd29f656d676/uvicorn/config.py#L298-L304
            "reload_includes": ["*.py", "*.css", "*.js", "*.html"],
            "reload_dirs": reload_dirs,
        }
```

schloerke:
Tests pass. Merging

PR #507, opened 2023-05-23
fix: Retry failing e2e-examples tests

schloerke:
I believe there is a small chance the "wait for app to load" logic will not trigger properly and time out.

While not adjusting that code, it will retry and hopefully not trigger the odd behavior. If it is truly a bug, it'll trigger all three times.


schloerke:
Timing can also be bad with the error message:

```
E       playwright._impl._api_types.Error: ReferenceError: Can't find variable: $
``

wch (approved):

PR #529, opened 2023-05-30
Convert table to web component

wch:
This converts the table to a web component. It also uses Preact instead of React, and bundles it.

Bundle size is 98kB. When It was using React, the bundle size was double that.

image

There are also updates to the build system and editor config to better support editing Typescript files.


shiny/ui/datagrid/js/package.json
```diff
@@ -4,13 +4,13 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "build": "tsc -noEmit && node build.mjs",
+    "build": "tsx build.ts",
```
jcheng5, line 7: I saw you took off `tsc -noEmit`, does `tsx build.ts` do type checking?

shiny/ui/datagrid/js/tsconfig.json
```diff
@@ -1,14 +1,15 @@
 {
   "compilerOptions": {
-    "module": "commonjs",
-    "lib": ["es5", "es6", "dom"],
+    "module": "ES2020",
+    "lib": ["ES2021", "DOM"],
```
jcheng5, line 4: What are the browser compatibility implications of this?

shiny/ui/datagrid/js/index.tsx
```diff
@@ -262,39 +262,11 @@ const roots = new WeakMap<HTMLElement, Root>();
 
 class ShinyDataGridBinding extends Shiny.OutputBinding {
   find(scope: HTMLElement | JQuery<HTMLElement>): JQuery<HTMLElement> {
-    return $(scope).find(".shiny-glide-data-grid-output");
+    return $(scope).find("shiny-glide-data-grid-output");
```
jcheng5, line 265: Should there be a difference between `<shiny-glide-data-grid-output>` tags intended for static rendering, and intended for Shiny output binding?

jcheng5: For Shiny outputs, there are three places we can take options:

1. Arguments to `output_foo()`, like `output_plot`'s `click` argument
2. Arguments to `render_foo()`, like `render.plot`'s `alt` argument
3. Attached to the object returned from the user's render code, like the `datatable()` in `DT` for R

In this case, I've been assuming it'd all be in 3 (although I'm not sure why). I think the answer does influence 👆

shiny/ui/datagrid/js/package.json
wch: Oh no, sorry. `tsx` is just a tool to run typescript code without compiling to JS.

I can add back the `tsc -noEmit`.

shiny/ui/datagrid/js/tsconfig.json
wch: If you go to https://caniuse.com/?search=es2021 and click on Feature List, then scroll all the way to the bottom and click on Date Relative, you'll see this:
image


The short story is that all major browsers have had support for all these features since early 2021.

I tried the same search for es2022, and found that Safari only got full support for es2022 features this year, which is why I went with es2021.
image

shiny/ui/datagrid/js/index.tsx
wch: > Should there be a difference between `<shiny-glide-data-grid-output>` tags intended for static rendering, and intended for Shiny output binding? (I'm asking)

I don't have strong feelings about it either way. It is nice to be able to reuse the `renderValue` code for both cases, but that could be done with class inheritance.

As for where to take options, that's a good question.

Maybe accept them in 1, and allow them to be overridden in 3? (We should be more consistent about this in general.)

shiny/ui/datagrid/js/tsconfig.json
jcheng5: Whoa, never saw that view, that's super cool. Though of course it's red/green 😠

PR #538, opened 2023-06-01
Add fast-rendering data table/data grid outputs

jcheng5:
**IN PROGRESS**

This PR introduces two distinct-but-related outputs: a data table, and a data grid.

### API

UI:

```python
ui.output_data_frame("grid")

Server:

```python
@output
@render.data_frame
def grid():
    return render.DataGrid(df(), row_selection_mode=input.selection_mode())
```

For table style, replace `render.DataGrid()` with `render.DataTable()`.

Output looks like this for grid:

image

And this for table:

image

### Common features

- [x] Designed for pd.DataFrame and objects that can be trivially converted to a pd.DataFrame (i.e. that have a `.to_pandas()` method)
- [x] Virtualized for fast rendering of lots of rows, using Tanstack Table and Tanstack Virtual
- [x] Optionally, allows rows to be selected (you can specify single or multi select); the selection state becomes available via `input` so you can use them to drive other calculations/outputs
- [x] Accessible for screen readers, keyboard, mobile
- [x] Sort by column, including multi-sort

### Data table specific

- [x] A conventional looking table. Pretty similar to [this](https://getbootstrap.com/docs/5.3/content/tables/#small-tables)
- [x] ~Pagination instead of virtual scroll~ Pushed to next release

### Data grid specific

- [x] Looks more like a spreadsheet
- [x] Gridlines
- [x] Virtual scroll instead of pagination

### For future releases

* Cell editing
* Filter rows using controls in the column header
* Find string in table (including across virtual rows)
* Simple column formatting (vert/horiz alignment, formatting of numbers and dates)
* Column/Cell/Range selection
* Sticky column(s)

### Out of scope, for now

* Paged/streaming data from the server. This might seem like a surprising omission for folks familiar with Shiny for R, since server-side DT is so commonly used. We can definitely add this in the future, but right now, we're finding that rendering is wicked fast (though you do have to pay the cost of transferring the entire data frame to the client, which will obviously be a problem beyond a certain size:bandwidth ratio).


jcheng5:
@wch I'm going to make a couple more changes, but only documentation/example/test related; the implementation should be ready for review. Thanks!

wch:
Some thoughts on the selection features. I had expected selection to work like a spreadsheet, but found that there were some surprising differences:
- Selecting many rows requires a lot of clicking -- each row needs to be clicked on. I would expect clicking and then dragging up or down to select multiple contiguous rows, at least in Grid mode.
- I would expect that if you have some rows selected, and then click on another row, it would unselect all rows, and then select the row that you just clicked on, even if it was in the previously-selected region.
- I would expect that if you click on a row, and then shift-click on another row, it would select all rows in between.
- It would be useful to have a quick way of selecting all and unselecting all. (Although unselecting all would be handled by the second point, above.)
- I wanted to copy selected rows and paste them in a document, but Cmd-C didn't do anything.


It would also be nice to be able to select regions and/or columns, but that could come later.

shiny/ui/dataframe/js/table-summary.tsx
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Create a summary
+ *
+ * @param summaryTemplate A string with "{start}", "{end}", and "{total}"
+ * @param scrollContainer Scrolling container of the table/grid
+ * @param virtualRows VirtualItem objects that might currently be visible
+ * @param thead The thead tag of the table
+ * @param nrows Number of total rows of data that exist
+ */
+export function useSummary(
```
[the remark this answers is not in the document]
wch: https://github.com/rstudio/py-shiny/assets/86978/e05101bc-787b-42b4-b4d9-c98a38141149

shiny/render/_dataframe.py
```diff
@@ -0,0 +1,320 @@
...
+    def __init__(
+        self,
+        data: object,
+        *,
+        width: Union[str, float, None] = "fit-content",
+        height: Union[str, float, None] = "500px",
+        summary: Union[bool, str] = True,
+        row_selection_mode: Union[
+            Literal["none"], Literal["single"], Literal["multi-toggle"]
+        ] = "none",
```
wch, line 29: This can be written more concisely:

```suggestion
        row_selection_mode: Literal["none", "single", "multi-toggle"] = "none",
```

shiny/render/_dataframe.py
```diff
@@ -0,0 +1,320 @@
...
+    def to_payload(self) -> object:
+        ...
+
+
+class DataGrid(AbstractTabularData):
+    def __init__(
+        self,
+        data: object,
+        *,
+        width: Union[str, float, None] = "fit-content",
```
wch, line 24: Because you have `from __future__ import annotations`, you can do:

```suggestion
        width: str | float | None = "fit-content",
```

shiny/ui/dataframe/js/build.ts
```diff
@@ -0,0 +1,27 @@
...
+async function bundle() {
+  try {
+    const options = {
+      entryPoints: ["index.tsx"],
+      bundle: true,
+      outdir: "dist",
+      minify: true,
+      sourcemap: true,
+      plugins: [sassPlugin({ type: "css-text" })],
+      target: ["safari12"],
```
wch, line 14: Any particular reason you chose `safari12` here?

shiny/render/_dataframe.py
```diff
@@ -0,0 +1,320 @@
...
+            style="table",
+        )
+        return res
+
+
+# It would be nice to specify DataGridResult to be something like:
+#   Union[pandas.DataFrame, <protocol with .to_pandas()>]
+# However, if we did that, we'd have to import pandas at load time, which adds
+# a nontrivial amount of overhead. So for now, we're just using `object`.
+DataGridResult = Union[None, object, DataGrid]
```
wch, line 185: I believe you can make this work if you add

```py
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    import pandas as pd
```

And then here:

```suggestion
DataGridResult = Union[None, "pd.DataFrame", DataGrid]
```

Note the quotes around `pd.DataFrame`. That's so that Python doesn't try to evaluate `pd.DataFrame` at load time. But the type checkers are smart enough to know that types in quotes are meant to be treated as types and not string literals.

shiny/render/_dataframe.py
```diff
@@ -0,0 +1,320 @@
...
+        data
+            A pandas `DataFrame` object, or any object that has a `.to_pandas()` method
+            (e.g., a Polars data frame or Arrow table).
+        width
+            A width for the data grid to occupy, in CSS units. The default is `fit-content`,
+            which sets the grid's width according to its contents. Set this to `100%` to use
+            the maximum available horizontal space.
+        height
+            A _maximum_ amount of vertical space for the data grid to occupy, in CSS units.
+            If there are more rows than can fit in this space, the grid will scroll.
```
wch, line 127: Is there a simple way to say `Inf`?

shiny/render/_dataframe.py
```diff
@@ -0,0 +1,320 @@
...
+
+        Parameters
+        ----------
+        data
+            A pandas `DataFrame` object, or any object that has a `.to_pandas()` method
+            (e.g., a Polars data frame or Arrow table).
+        width
+            A width for the data grid to occupy, in CSS units. The default is `fit-content`,
+            which sets the grid's width according to its contents. Set this to `100%` to use
+            the maximum available horizontal space.
```
wch, line 124: A nit: I noticed these lines are longer than the 88 chars that black wants to use. Is there a particular reason for that? They can be reflowed with Alt-Q in VS Code.

shiny/render/_dataframe.py
```diff
@@ -0,0 +1,320 @@
...
+        height
+            A _maximum_ amount of vertical space for the data grid to occupy, in CSS units.
+            If there are more rows than can fit in this space, the grid will scroll.
+        summary
+            If `True` (the default), shows a message like "Viewing rows 1 through 10 of 20"
+            below the grid when not all of the rows are being shown. If `False`, the message
+            is not displayed. You can also specify a string template to customize the
+            message, for example: `"Viendo filas {start} a {end} de {total}"`.
```
wch, line 132: Are `start`, `end`, and `total` the complete set of variables they can use?

shiny/render/_dataframe.py
```diff
@@ -0,0 +1,320 @@
...
+        row_selection_mode
+            Use `"none"` to disable row selection, `"single"` to allow a single row to be
+            selected at a time, and `"multi-toggle"` to allow multiple rows to be selected
```
wch, line 135: `multi-toggle` is a bit clunky. Any reason not to go with `multiple`?

Also, I just want to make sure we don't paint ourselves into a corner with the parameter name `row_selection_mode`. If we add region selection in the future (like a spreadsheet), can we do that without too much trouble?

Maybe instead we should have a `selection_mode` parameter, with values like `row-single` and `row-multi`?

shiny/ui/dataframe/js/build.ts
```diff
@@ -0,0 +1,27 @@
+import { build } from "esbuild";
+import { sassPlugin } from "esbuild-sass-plugin";
+import * as fs from "node:fs";
+
+async function bundle() {
+  try {
+    const options = {
```
wch, line 7: I think we need to add `format: "esm"`. esbuild's default is `iife` (which seems to work OK, though)

wch (changes requested):

wch:
Reminder: we'll need to commit the files in `dist` at some point.

shiny/ui/dataframe/js/selection.tsx
```diff
@@ -0,0 +1,180 @@
...
+      return selectedKeys;
+    },
+
+    itemHandlers() {
+      return { onMouseDown, onKeyDown };
+    },
+  };
+}
+
+const isMac = /^Mac/.test(window.navigator.platform);
```
wch, line 128: Apparently `window.navigator.platform` is deprecated. https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform

https://stackoverflow.com/questions/10527983/best-way-to-detect-mac-os-x-or-windows-computers-with-javascript-or-jquery

jcheng5:
Errors seem to break the component. In this example, the app loads with an error (due to incorrect pandas code). Selecting "cut" should render a table, but it doesn't. If you modify the input_selectize call with `selected="cut"` then it renders fine. My guess is that Shiny's default error rendering codepath is killing the React root somehow.

```python
import numpy as np
import pandas as pd
import seaborn as sns

from shiny import App, reactive, render, req, ui

df = sns.load_dataset("diamonds")

app_ui = ui.page_fluid(
    ui.input_selectize("summarize_by", "Summarize by", df.columns.to_list()),
    ui.output_data_frame("summary_data"),
    ui.output_data_frame("detail_data"),
)


    @reactive.Calc
    def summarized():
        numeric_df = df.select_dtypes(include=[np.number])
        return numeric_df.groupby(df[input.summarize_by()]).mean().reset_index()

    @output
    @render.data_frame
    def summary_data():
        return summarized()

    @output
    @render.data_frame
    def detail_data():
        ...


app = App(app_ui, server)
```

shiny/render/_dataframe.py
jcheng5: Whoa, how did I not know that?

shiny/render/_dataframe.py
jcheng5: I was imagining `multiple` would be the OS-native-like way I demoed earlier (click, Shift-Click, Ctrl-Shift-Click, etc. all working the way Finder or File Explorer or native select boxes do).

The main inherent difference between `multiple` and `multi-toggle` is that in the latter case, you can click on and click off each row independently. In the former, you can't select more than one row without a modifier key, and you also go back to having none selected without a modifier key. Maybe the `multi-toggle` is just better/more intuitive and I should forget about the other mode?

shiny/ui/dataframe/js/build.ts
jcheng5: OK, I can try that.

shiny/ui/dataframe/js/build.ts
jcheng5: Ummmm. I don't remember. I think I looked at what versions of iOS were supported by what iPhones, and therefore what version of Safari a not-too-old phone might be stuck on? Sorry, I should have documented that decision. What have you been setting it to?

shiny/ui/dataframe/js/selection.tsx
jcheng5: WHY 😠

shiny/ui/dataframe/js/table-summary.tsx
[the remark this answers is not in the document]
jcheng5: Hmmm, I'm not throttling anything intentionally. I know the Virtual library has some kind of smooth scrolling effect, I wonder if that's why? The other possibility I can think of is that the measuring of these items doesn't happen until after they're scrolled into view.

Either way, I'll make sure to fix this.

shiny/ui/dataframe/js/build.ts
wch: In stuff I've been working on, I haven't specified the `target`, and it apparently defaults to `"esnext"`. I don't know if that's a problem in practice or not.

https://esbuild.github.io/api/#target

shiny/render/_dataframe.py
jcheng5: Yes.

shiny/render/_dataframe.py
jcheng5: Re: `row_selection_mode` vs. `selection_mode`, I was thinking that there would be these different selection modes, and multiple of these could be active at the same time:

* Row: off, single, multiple
* Column: off, single, multiple
* Cell: off, single, multiple
* Range: off, on

I can think of three possibilities:

1. Separate selection mode parameters for row, column, cell, range. The downside is lots of parameters.
2. One selection mode parameter that takes a string or list. The downside is the type system doesn't help you avoid invalid combinations (like `selection_mode = ["row-single", "row-multi"]`)
3. One selection mode parameter that takes a named tuple: `selection_mode = {"row": "single", "column": "multiple"}`

We could also decide that allowing multiple simultaneous selection modes is not worth the conceptual overhead and complexity.

wch:
I just noticed that if you click on the handle in the scroll bar and move the mouse up and down, the handle doesn't move as far as the mouse cursor.

https://github.com/rstudio/py-shiny/assets/86978/ed7cc524-ea4b-43e6-ad5d-2ad168139d4d

shiny/render/_dataframe.py
jcheng5: I've changed `multi-toggle` to `multiple`. It's still `row_selection_mode` though. If you feel strongly about this, we can discuss, otherwise I'm inclined to ship it like this.

shiny/render/_dataframe.py
jcheng5: I've made the docs clearer on this point now.

jcheng5:
cc @shalutiwari

jcheng5:
@wch gave me the green light to merge over zoom.

PR #540, opened 2023-06-01
fix: add meta viewport tag

### LINKED ISSUE #535: Missing <meta name="viewport">
This line appears in the `<head>` for Shiny for R but not Shiny for Python:

```html
<meta name="viewport" content="width=device-width, initial-scale=1" />
```
 
Without this, phone browsers show everything very small. (Left is without the tag, right is with it)
Screenshots

nealrichardson:
Closes #535

### DIFF
--- a/shiny/ui/_html_dependencies.py
+++ b/shiny/ui/_html_dependencies.py
@@ -13,6 +13,7 @@ def bootstrap_deps() -> list[HTMLDependency]:
         source={"package": "shiny", "subdir": "www/shared/bootstrap/"},
         script={"src": "bootstrap.bundle.min.js"},
         stylesheet={"href": "bootstrap.min.css"},
+        meta={"name": "viewport", "content": "width=device-width, initial-scale=1"},
     )
     deps = [jquery_deps(), dep]
     return deps


wch:
Am I correct that the `<meta name="viewport">` is orthogonal to the Bootstrap dependency? If that's the case, then I think it makes sense to add a `tags.meta()` inside of the `tags.head()` here:

https://github.com/rstudio/py-shiny/blob/b874f0f0b33421d17a7dd6caac888e7b989fb445/shiny/ui/_page.py#L237

nealrichardson:
I believe you're right but I'm not confident. I just looked for where it gets added in shiny in R and tried to do the same. Happy to move it elsewhere here if you think best (and presumably do the same over there too?)

nealrichardson:
The internet confirms that this is an HTML5 thing and not a Bootstrap thing and it should probably [always](https://www.w3schools.com/css/css_rwd_viewport.asp) be there. But the place you suggested is still inside a call to `page_bootstrap()`--is that right?

wch:
>  But the place you suggested is still inside a call to `page_bootstrap()`--is that right?

That is right. Now that I've thought about it a bit more, I think it does make sense to just put it in the Bootstrap dependency like you've done, since all Bootstrap users will want it. Also, that's also how we did it on the R side and that has gone smoothly.

@nealrichardson Can you sign the CLA Assistant thing? Then feel free to merge.

wch (approved):

wch:
@schloerke I think we need to do the same for shinyswatch?

PR #547, opened 2023-06-05
chore: Replace shiny.rstudio.com with shiny.posit.co

nealrichardson:


shiny/input_handler.py
```diff
@@ -48,7 +48,7 @@ def _process_value(self, type: str, value: Any, name: str, session: Session) ->
 to the ``input`` argument of an :func:`~shiny.App`'s ``server`` function.
 
 The ``type`` is based on the ``getType()`` JavaScript method on the relevant Shiny
-input binding. See `this article <https://shiny.rstudio.com/articles/js-custom-input.html>`_
+input binding. See `this article <https://shiny.posit.co/articles/js-custom-input.html>`_
```
nealrichardson, line 51: This is a post from 2017 about doing this in the R package. IIUC the JavaScript parts would still be the same between Python and R, but the Python part would be different. 

1. Is this article still accurate even for R?
2. Assuming it should be linked here, we should note that the article is useful for the JS part, and the Python binding is analogous to the R code, right?

shiny/ui/_bootstrap.py
```diff
@@ -47,7 +47,7 @@ def row(*args: TagChild | TagAttrs, **kwargs: TagAttrValue) -> Tag:
     elements that should appear on the same line (if the browser has adequate width) and
     :func:`~shiny.ui.column` to define how much horizontal space within a 12-unit wide
     grid each on of these elements should occupy. See the [layout
-    guide](https://shiny.rstudio.com/articles/layout-guide.html>) for more context and
+    guide](https://shiny.posit.co/articles/layout-guide.html>) for more context and
```
nealrichardson, line 50: This article is about R. Is it accurate enough that it's worth linking to here, noting that the principles are the same but the function names have to be adapted to Python? Is there something better to link here? Or should we drop it? 

cc @cpsievert

shiny/input_handler.py
wch: 1. The article is still accurate for R.
2. That sounds good to me.

As an aside, the formatting on that particular page got messed up from the Jekyll->Quarto transition.
https://github.com/rstudio/shiny-dev-center/issues/240

wch (approved):

PR #565, opened 2023-06-16
Make sure fill dependency is added

wch:
This makes sure that the `fill.css` HTML dependency is always added when fill components are used in an app.

### DIFF
--- a/shiny/experimental/ui/_output.py
+++ b/shiny/experimental/ui/_output.py
@@ -217,14 +217,14 @@ def output_image(
         args.update(**format_opt_names(brush, "brush"))
 
     container = func(
-        # NEW
-        as_fill_item() if fill else None,
-        # /NEW
         id=id_resolved,
         class_="shiny-image-output",
         style=style,
         **args,
     )
+    if fill:
+        container = as_fill_item(container)
+
     return container
 
 
@@ -267,12 +267,14 @@ def output_ui(
     if not container:
         container = tags.span if inline else tags.div
     res = container(
-        # NEW
-        as_fill_item() if fill else None,
-        as_fillable_container() if fillable else None,
-        # /NEW
         {"class": "shiny-html-output"},
         id=resolve_id(id),
         **kwargs,
     )
+
+    if fillable:
+        res = as_fillable_container(res)
+    if fill:
+        res = as_fill_item(res)
+
     return res
--- a/shiny/experimental/examples/as_fill_item/app.py
+++ b/shiny/experimental/examples/as_fill_item/app.py
@@ -58,7 +58,7 @@ app_ui = ui.page_fluid(
     ui.row(
         ui.column(4, ui.div(outer0)),
         ui.column(4, ui.div(outer1)),
-        ui.column(4, ui.span(x.ui.as_fill_carrier(), outer2)),
+        ui.column(4, x.ui.as_fill_carrier(ui.span(outer2))),
     ),
 )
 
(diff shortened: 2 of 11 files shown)


shiny/experimental/ui/_output.py
```diff
@@ -217,14 +217,14 @@ def output_image(
...
-        # NEW
-        as_fill_item() if fill else None,
-        # /NEW
         id=id_resolved,
         class_="shiny-image-output",
         style=style,
         **args,
     )
+    if fill:
+        container = as_fill_item(container)
```
schloerke, line 226:
```suggestion
    # NEW
    if fill:
        container = as_fill_item(container)
    # /NEW
```

shiny/experimental/ui/_output.py
```diff
@@ -267,12 +267,14 @@ def output_ui(
...
-        # /NEW
         {"class": "shiny-html-output"},
         id=resolve_id(id),
         **kwargs,
     )
+
+    if fillable:
+        res = as_fillable_container(res)
+    if fill:
+        res = as_fill_item(res)
```
schloerke, line 278:
```suggestion
    # NEW
    if fillable:
        res = as_fillable_container(res)
    if fill:
        res = as_fill_item(res)
    # /NEW
```

schloerke (approved):

shiny/experimental/ui/_output.py
wch: What does `NEW` indicate?

wch:
For future reference: Previous to this PR, `as_fill_carrier()` could be used in two ways:

- Wrapping: `as_fill_carrier(div("Hello world"))`
- Inline: `div(as_fill_carrier(), "Hello world"))`

The function is often used conditionally. These are the two different ways to do it:

```py
# Wrapping
x = div("Hello world")
if fill:
    x = as_fill_carrier(x)

# Inline
div(
    as_fill_carrier() if fill else None,
    "Hello world"
)
```


The problem (which caused incorrect styling) is that the second usage does not include the `fill.css` HTML dependency, and it can't in Python. Well, it could work in Python, but then the user would have to use `*`, as in :

```py
div(*as_fill_carrier(), "Hello world"))

# To add it conditionally. Note the brackets around None, which are needed
# due to the precedence of *
div(
  *as_fill_carrier() if fill else [None],
  "Hello world"
)
```

One possibly confusing issue is that `as_xxx()` functions generally take an input object and return an object that is an `xxx`. In the second usage, that's not what happens -- instead, it returns an object that the parent consumes to transform itself. If we were to allow this kind of usage, we would at least want to give it a different name, so as not to be confusing about what the function does.


Because the `fill.css` HTML dependency can't be added with the inline form, this means that whoever uses this function needs to understand that part about the underlying implementation, and know to avoid the inline form when an HTML dependency needs to be added to the application. So instead of reducing the number of things that the user needs to know and think about, this abstraction (the `as_fill_carrier()` function) increases them, and it introduces the possibility of a bugs in the form of missing HTML dependencies, which actually happened, and what led us to look at this code more closely.

The wrapping usage may be a little clunkier than the inline usage when `as_fill_carrier()` is used conditionally, but if we only allow that one way of using it:
- It reduces the number of things that the user needs to know about the underlying implementation (i.e., the HTML dependency)
- It makes it so the user doesn't need to think about when it will and won't work (due to the missing HTML dependency)
- It eliminates the possibility of bugs due to the differences in behavior of the two forms of usage

So overall, if only the wrapping usage of the function is allowed, I think it simplifies the code and reduces the chances of bugs, even if the code may be a bit more verbose in some cases.

PR #570, opened 2023-06-19
Cache browsers for Playwright

wch:
This PR caches browsers for Playwright in GitHub Actions.


wch:
It turns out that this change doesn't actually speed up CI runs, because the time spent installing browsers is pretty small; most of the time is spent installing Ubuntu system dependencies.

PR #574, opened 2023-06-21
Tweak example app so Plotly plots fill their cards

jcheng5:


wch (approved):
Just curious -- would it be possible to do this all in Javascript?

cpsievert:
In theory, yes, this is how the [R binding does it](https://github.com/plotly/plotly.R/blob/0469eab/inst/htmlwidgets/plotly.js#L10-L16)

More generally, some more thought should be put into how `output_widget()` should better handle this for any widget via ipywidgets sizing API https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Layout.html

PR #582, opened 2023-06-23
More realistic file import example 

### LINKED ISSUE #567: More realistic file import example
Typically when users use file import they import the file, process it, and then send it to several renderers. Our example uses a fairly unrealistic pattern of reading in a file and sending it to an HTML table which leads people a bit astray because it's a pattern which doesn't  generalize very well. 

We should add an example which reads the data into a reactive value:
- Data is imported and sent to reactive value
- Checkboxes give people some options (row count, column count, column names)
- A table is rendered with their choices and the stats from the data

adejumoridwan:
Closes #567 
Change the file import example to display summary stats of uploaded data, such as row count, column count and column names.


wch:
This PR includes your Snowflake credentials. That doesn't seem right...

We appreciate this PR, but it's not clear to me that it is an improvement for the intended purpose of the app, which is just to illustrate how to use the file upload component. The old app just shows how to do file uploads, while the new one also adds some interactivity with checkboxes. The new version is in a sense a more realistic app, but in this case I don't think the changes are helpful for illustrating how the file upload component works.

wch:
Apologies, I looked at this too quickly and didn't see the linked issue at the top, and the discussion in that issue.

adejumoridwan:
No problem, as for the snowflake credentials, it has been removed

wch:
Great, thanks!

Just a heads up: I'm working on some review comments now but I have a meeting so it'll be a couple hours before I'll be able to submit them.

GShotwell:
Thanks, I think I should've chimed in a bit more since I think I suggested this PR. The main things I find confusing about the current example are:
- It uses conditional UI which people might not know (people probably build file-import apps before they build conditional UI apps)
- It uses `ui.HTML` output to render a table which isn't that common (most of the time people use `ui.output_table` or `ui.output_data_frame`

shiny/examples/input_file/app.py
```diff
@@ -0,0 +1,48 @@
...
+app_ui = ui.page_fluid(
+    ui.input_file("file", "File", accept=".csv"),
+    ui.input_checkbox_group(
+        "stats", "Summary Stats", choices=["Row Count", "Column Count", "Column Names"]
+    ),
+    ui.output_table("summary"),
+)
+
+
+def server(input, output, session):
```
wch, line 14: Please keep the types; they are helpful when the editor supports them (like VS Code and the shinylive online editor), and for some editing environments they are necessary for the editor to not underline things in red.

```suggestion
def server(input: Inputs, output: Outputs, session: Session):
```

shiny/examples/input_file/app.py
```diff
@@ -0,0 +1,48 @@
...
+    @reactive.Calc
+    def parsed_file():
+        file = input.file()
```
wch, line 17: Similar to above, having the type annotation can be helpful:

```suggestion
        file: list[FileInfo] = input.file1()
```

shiny/examples/input_file/app.py
```diff
@@ -0,0 +1,48 @@
...
+        # Create a new DataFrame to display the information
+        info_df = pd.DataFrame(
+            {
+                "Row Count": [row_count],
+                "Column Count": [column_count],
+                "Column Names": [column_names],
+            }
+        )
+
+        return info_df.loc[:, input.stats()]
```
wch, line 45: The reader might not have the mental model that `input.stats()` returns a list of strings, so 
It will help to have a comment explaining what's happening here:

```suggestion
        # input.stats() is a list of strings; subset the columns based on the selected
        # checkboxes
        return info_df.loc[:, input.stats()]
```

wch (changes requested):
I've added some comments which apply to both copies of the app. Thanks for your efforts!

wch (approved):
Looks good - I found that the two versions of app.py aren't quite the same so I'll get them in sync and merge.

Update: It looks like I'll have to merge and then fix in a separate commit.

PR #584, opened 2023-06-26
Annotation export example

GShotwell:
This creates a slightly more real-world brushing example where the user can visually annotate a plot, and then download the annotated data.


examples/annotation-export/app.py
```diff
@@ -0,0 +1,128 @@
...
+
+def server(input: Inputs, output: Outputs, session: Session):
+    annotation_values = reactive.Value(None)
+
+    @reactive.Calc
+    def ts_data():
+        path = Path(__file__).parent / "boulder_temp.csv"
+        dataframe = pd.read_csv(path)
+        dataframe["date"] = pd.to_datetime(dataframe["date"])
+        dataframe["annotation"] = ""
```
wch, line 43: I think the CSV reading should happen outside of the server function, at the top level of the app, so that it's read only once. Otherwise it'll be read every time any user changes their copy of `annotation_values`.

examples/annotation-export/app.py
```diff
@@ -0,0 +1,128 @@
...
+
+        out.tick_params(axis="x", rotation=30)
+        return out.get_figure()
+
+    @output
+    @render.ui
+    def annotator():
+        if input.time_series_brush() is not None:
+            selected = selected_data().copy()
+
```
wch, line 92: I think this `copy()` isn't needed.

Now that I think about it, how about updating this app to use copy on write? That should eliminate the need for using `.copy()`.
https://github.com/rstudio/py-shiny/issues/614
https://pandas.pydata.org/docs/dev/user_guide/copy_on_write.html

wch (changes requested):
I like the idea of the app but it seems to not be working for me:

https://github.com/rstudio/py-shiny/assets/86978/c9f023de-6a28-4a19-802a-aa3be30b8081

examples/annotation-export/app.py
GShotwell: I tried a few things, and it turns out that moving some date time processing up in the script means that all of the copies can be removed without needing the pandas copy-on-modify.

GShotwell:
@wch I think the app was working but maybe poorly documented. The idea was to show a table of annotated values, so if you don't add any annotations the table should be empty. I added a placeholder and heading to make that a bit clearer.

wch:
I just pushed some small formatting fixes. Note that the `br()` was added to make things look better for users that aren't using wide browser windows.

examples/annotation-export/app.py
wch: It looks like there's still a place where the `reactive.Value` `ts_data()` is being modified in place; the modified value is downloaded in `download`, where it calls `ts_data.to_csv()`. This results in the expected overall behavior for the app, but it relies on a dangerous modification-in-place. I'll make a fix.

PR #586, opened 2023-06-27
Duckdb example

GShotwell:
This creates a duckdb query app which illustrates on disk data integration, modules, and inserting/removing UIs.


wch:
Looks great!

Issue #593, opened 2023-07-05
Install fails in mac M1, missing immutable


tr31zh:
**Computer**: Mac M1
**Python**: 3.11.3
**Env**: conda

**Description**:
When installing with pip install shiny error ModuleNotFoundError: No module named immutables

**Dump**:
Collecting shinywidgets
  Using cached shinywidgets-0.2.1-py3-none-any.whl (1.8 MB)
Requirement already satisfied: ipywidgets>=7.6.5 in /opt/homebrew/anaconda3/envs/sam/lib/python3.11/site-packages (from shinywidgets) (8.0.6)
Requirement already satisfied: jupyter-core in /opt/homebrew/anaconda3/envs/sam/lib/python3.11/site-packages (from shinywidgets) (5.3.1)
Collecting shiny>=0.3.0 (from shinywidgets)
  Using cached shiny-0.4.0-py3-none-any.whl (2.2 MB)
Requirement already satisfied: python-dateutil>=2.8.2 in /opt/homebrew/anaconda3/envs/sam/lib/python3.11/site-packages (from shinywidgets) (2.8.2)
Requirement already satisfied: ipykernel>=4.5.1 in /opt/homebrew/anaconda3/envs/sam/lib/python3.11/site-packages (from ipywidgets>=7.6.5->shinywidgets) (6.23.3)
Requirement already satisfied: ipython>=6.1.0 in /opt/homebrew/anaconda3/envs/sam/lib/python3.11/site-packages (from ipywidgets>=7.6.5->shinywidgets) (8.14.0)
Requirement already satisfied: traitlets>=4.3.1 in /opt/homebrew/anaconda3/envs/sam/lib/python3.11/site-packages (from ipywidgets>=7.6.5->shinywidgets) (5.9.0)
Requirement already satisfied: widgetsnbextension~=4.0.7 in /opt/homebrew/anaconda3/envs/sam/lib/python3.11/site-packages (from ipywidgets>=7.6.5->shinywidgets) (4.0.7)
Requirement already satisfied: jupyterlab-widgets~=3.0.7 in /opt/homebrew/anaconda3/envs/sam/lib/python3.11/site-packages (from ipywidgets>=7.6.5->shinywidgets) (3.0.7)
Requirement already satisfied: six>=1.5 in /opt/homebrew/anaconda3/envs/sam/lib/python3.11/site-packages (from python-dateutil>=2.8.2->shinywidgets) (1.16.0)
[...]

PR #597, opened 2023-07-05
Remove contextvars dependency

jcheng5:
Per @nealrichardson comment in #593, this was only needed for Python 3.7, which we no longer support.


jcheng5:
Whoops he already did it

PR #598, opened 2023-07-06
Make --app-path work with app file argument

jcheng5:
Without this, `shiny run --app-dir=foo app.py` won't work even if foo/app.py exists and is a valid Shiny app. This is because the code that turns `app.py` into `app:app` for Uvicorn wasn't aware of --app-dir.

### DIFF
--- a/shiny/_main.py
+++ b/shiny/_main.py
@@ -360,18 +360,20 @@ def resolve_app(app: str, app_dir: Optional[str]) -> tuple[str, Optional[str]]:
         attr = "app"
 
     if is_file(module):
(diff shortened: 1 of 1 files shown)


jcheng5:
I don't believe this is an area where we have great (any?) automated tests.

nealrichardson:
Works as expected on my machine. 

Should we make a followup about adding tests for the CLI code paths? I don't think it would be that hard (famous last words).

jcheng5:
Agreed that tests for this area don’t seem that hard.

nealrichardson (approved):
Works on my machine. 

I just rebased on main to hopefully fix CI.

PR #607, opened 2023-07-08
Add example app on Snowflake

### LINKED ISSUE #515: Snowflake example app
There are all kinds of great uses for the Snowflake [Snowpark library](https://pypi.org/project/snowflake-snowpark-python/). It would be great to have an example app which made use of that integration for two main reasons:

1) It's a really good example of why an R shiny user might be interested in the Python library. 
2) There are great Stremlit apps which make use of Snowflake, but I believe they all use Snowflake packages which are available to Shiny. It's possible that Shiny might do a better job because reactive execution might minimize calls to expensive Snowflake operations.

adejumoridwan:
Closes #515 
Added an example that integrates snowflake and python shiny using the snowpack library.


GShotwell:
Thanks so much for this, we're looking into how we can host apps which communicate with Snowflake.

PR #610, opened 2023-07-10
Move dataframe JavaScript files into /js subdirectory

wch:
This PR moves files from `shiny/ui/dataframe/js` to `/js`, and slightly reorganizes the directory structure. It moves the .ts files from `/js` to `/js/dataframe`.

This will set things up to add other components.


jcheng5:
There's probably a conversation we'll need to have at some point about whether these components should all be individual bundles or a single big one? But I'm happy to defer that until later.

jcheng5 (approved):

wch:
> There's probably a conversation we'll need to have at some point about whether these components should all be individual bundles or a single big one? But I'm happy to defer that until later.

I was planning on making a separate bundle for upcoming components, because the dataframe.js file is about 100KB and probably doesn't share many dependencies with other components. But I could be persuaded either way.

PR #612, opened 2023-07-11
Add classification component

wch:
This component is still a work in progress, so this PR is a placeholder for now.


js/ml/classification-label.ts
```diff
@@ -0,0 +1,120 @@
...
+      flex-direction: row;
+      align-items: baseline;
+    }
+
+    .dashed-line {
+      flex: 1 1 0%;
+      border-bottom: 1px dashed #888;
+      margin-left: 0.5em;
+      margin-right: 0.5em;
+    }
```
jcheng5, line 48: Were you planning to make these rules use CSS variables? And/or default to existing CSS variables?

Issue #611, opened 2023-07-10
Warnings with matplotlib 3.7.2


wch:
When running the penguins app with the latest version of matplotlib, 3.7.2, it raises the following warnings:

```
$ shiny run examples/penguins/app.py 
...
/Users/winston/Projects/py-shiny/shiny/render/_try_render_plot.py:47: UserWarning: The figure layout has changed to tight
  plt.tight_layout()  # pyright: ignore[reportUnknownMemberType]
```

The warning does not occur with 3.7.1.

Looking at the [matplotlib 3.7.2 release notes](https://matplotlib.org/3.7.2/users/github_stats.html), it might have something to do with this?
https://github.com/matplotlib/matplotlib/pull/25624

wch:
This seems to happen with the Penguins app only when marginal plots are enabled.

GShotwell:
Any objection to just squashing the warning?

tacaswell:
I think this is a regression it Matplotlib between 3.7.1 and 3.7.2.

wch:
@tacaswell Thanks for finding the bug, and weighing in here.

tacaswell:
You should note I also put the bug in 🐑 .

GShotwell:
Closed via https://github.com/rstudio/py-shiny/pull/615

PR #615, opened 2023-07-12
Silence figure layout warning on matplotlib 3.7.2

GShotwell:
It seems like there's a matplotlib regression which causes our CI to fail [Issue #611](https://github.com/rstudio/py-shiny/issues/611). Since I was trying to debug this issue on another branch I went ahead and silenced that warning on the penguins app.

### DIFF
--- a/examples/penguins/app.py
+++ b/examples/penguins/app.py
@@ -14,7 +14,7 @@ from colors import bg_palette, palette
 import shiny.experimental as x
 from shiny import App, Inputs, Outputs, Session, reactive, render, req, ui
 
-# There is a matplotlib bug which causes pyright failures
+# There is a matplotlib bug which causes CI failures
 # see https://github.com/rstudio/py-shiny/issues/611#issuecomment-1632866419
 if matplotlib.__version__ == "3.7.2":
     warnings.filterwarnings(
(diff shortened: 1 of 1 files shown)


examples/penguins/app.py
```diff
@@ -12,6 +14,13 @@
 import shiny.experimental as x
 from shiny import App, Inputs, Outputs, Session, reactive, render, req, ui
 
+# There is a matplotlib bug which causes pyright failures
```
wch, line 17: Small nit: the warning actually has nothing todo with pyright - it just happens to be on a line where there is a pyright comment.

wch (approved):
Looks good to me, besides the one small comment.

I think it makes sense to suppress this warning given that we don’t know when the next matplotlib release (which will hopefully fix the issue) will be.

examples/penguins/app.py
GShotwell: Oh right, thanks for the clarification

schloerke:
Moved warning check into the CI test that was originally failing: https://github.com/rstudio/py-shiny/pull/627/commits/93033fde00208f5806b7364fec9e37666657b9f3

PR #640, opened 2023-07-25
Fix jumpy cursor in numeric filter

jcheng5:
This change makes the numeric inputs "uncontrolled" from a React perspective. Not sure why the controlled approach was causing the jumpy cursor.

Also didn't realize that numeric inputs only support integers by default.


nstrayer (approved):
Looks good to me.  The joys of react forms.

PR #649, opened 2023-07-28
Allow ui.update_slider to handle non-numeric values

jcheng5:
Fixes #648

- [x] Test that value can actually be list
- [x] Add changelog entry

PR #650, opened 2023-07-29
Model score example

### LINKED ISSUE #638: Create dashboard example
We need a kind of classic dashboard example with cards, value boxes, etc.

jcheng5:
Depends on #649

Fixes #638 

- [x] ~Performance is oddly slow since 6c8f665, run profiler~ Previously, we were syncing each new Plotly figure's `.data` tuple with the existing widget. This was pretty fast. Then I added updating of each figure's `layout`, which is pretty slow. I made that optional and disabled it in our app--now it's fast again.


examples/model-score/app.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        Returns a Plotly Figure visualization. Streams new data to the Plotly widget in
+        the browser whenever filtered_df() updates, and completely recreates the figure
+        when filtered_model_names() changes (see recreate_key=... above).
+        """
+        fig = px.line(
+            filtered_df(),
+            x="time",
+            y="score",
+            labels=dict(score="auc"),
```
GShotwell, line 225: After thinking about it I'm not sure we should use AUC for this. While AUC can vary from 0-1, it's pretty weird to have a number lower than 0.5 because that indicates that the model is worse than guessing. Maybe we should change these to "accuracy".

examples/model-score/app.py
```diff
@@ -0,0 +1,305 @@
...
+    tbl["time"] = tbl["timestamp"].dt.strftime("%H:%M:%S")
+
+    return tbl
+
+
+model_colors = {
+    "model_1": "#7fc97f",
+    "model_2": "#beaed4",
+    "model_3": "#fdc086",
+    "model_4": "#ffff99",
```
GShotwell, line 78: This color is quite faint and may not be accessible, can we change it to something bolder?

examples/model-score/app.py
```diff
@@ -0,0 +1,305 @@
...
+            ),
+            ui.input_radio_buttons(
+                "timeframe",
+                "Timeframe",
+                ["Latest", "Specific timeframe"],
+                selected="Latest",
+            ),
+            ui.panel_conditional(
+                "input.timeframe === 'Latest'",
+                ui.input_selectize(
```
GShotwell, line 100: I'm not sure that people will view this app long enough to get value out of the longer time intervals, would it be better to make this a numeric input with 5s steps?

examples/model-score/app.py
```diff
@@ -0,0 +1,305 @@
...
+            line=dict(color=THRESHOLD_LOW_COLOR, width=2),
+        )
+        fig.add_hline(
+            THRESHOLD_MID,
+            line_dash="dash",
+            line=dict(color=THRESHOLD_MID_COLOR, width=2),
+        )
+
+        fig.update_yaxes(range=[0, 1], fixedrange=True)
+        fig.update_xaxes(fixedrange=True, tickangle=60, dtick="M5")
```
GShotwell, line 247: When switching to static view you get a very crowded x axis. Setting tick marks directly is a bit less noisy.

```suggestion
        fig.update_xaxes(fixedrange=True, tickangle=60, nticks=6)
```

examples/model-score/app.py
```diff
@@ -0,0 +1,305 @@
...
+            color="model",
+            color_discrete_map=model_colors,
+            # The default for render_mode is "auto", which switches between
+            # type="scatter" and type="scattergl" depending on the number of data
+            # points. Switching that value breaks streaming updates, as the type
+            # property is read-only. Setting it to "webgl" keeps the type consistent.
+            render_mode="webgl",
+        )
+
+        fig.add_hline(
```
GShotwell, line 235: I don't like that the threshold lines are stronger than the score lines. Can we do something like:
- Make the background a bit lighter
- Turn the alpha down on the threshold lines

examples/model-score/plotly_streaming.py
```diff
@@ -0,0 +1,100 @@
...
+    if hasattr(obj, "to_list") and callable(obj.to_list):
+        return obj.to_list()
+    if hasattr(obj, "tolist") and callable(obj.tolist):
+        return obj.tolist()
+    if hasattr(obj, "to_dict") and callable(obj.to_dict):
+        return obj.to_dict()
+    raise TypeError(f"Object of type {type(obj)} is not JSON serializable")
+
+
+def render_plotly_streaming(
```
GShotwell, line 29: This is awesome, I wonder if we could do something similar with matplotlib.

GShotwell (changes requested):
Overall looks great, I think it would be good to do a bit of work on the plot to make the model lines stand out a bit more.

examples/model-score/app.py
jcheng5: I didn't think people would actually sit there and wait, but I did think it was representative of what people might do in real life. I can put some other intervals in there though.

examples/model-score/app.py
jcheng5: OK awesome, I wasn't sure how to do that.

examples/model-score/plotly_streaming.py
jcheng5: That would be super helpful, the matplotlib rendering time definitely limits its usefulness in realtime apps.

examples/model-score/app.py
jcheng5: Now using px.colors.qualitative.D3 (see https://plotly.com/python/discrete-color/#color-sequences-in-plotly-express for other options)

examples/model-score/app.py
jcheng5: I got the best results by accepting the default.

examples/model-score/app.py
jcheng5: How's this? (I switch to a simpler Plotly theme, the gridlines and background are both gone)

image

GShotwell: That looks great to me.

examples/model-score/plotly_streaming.py
GShotwell: I raised this issue https://github.com/rstudio/py-shiny/issues/653

PR #668, opened 2023-08-08
Wrap bare value box value in `p` tags

### LINKED ISSUE #646: value_box value is too small
```python
x.ui.value_box("model_1", "0.98")
```

image

This is different than the R version of this function. The reason is because in Python, "0.98" is going in without an enclosing tag, while in R, a bare string is wrapped in `<p></p>`.

schloerke:
Fixes #646
- [x] Changelog entry


schloerke:
Failing test is related to https://github.com/rstudio/py-shiny/pull/675 (and is unrelated to this PR).

Merging

PR #677, opened 2023-08-11
Make data frame selection return row numbers, not pandas index value

### LINKED ISSUE #676: Data frame selected rows intermittently returning None Type
When selecting the fourth (not the 1-3rd) entry of the [data_frame example](https://shinylive.io/py/examples/#data-frame-grid) the row selection produces a `None` value which causes an error. Weirdly this seems to be specific to that particular entry and persists you reorder or filter the data.

jcheng5:
Fixes #676 

The code in js/dataframe/index.tsx uses the row number ([0, 1, 2, ...]) as the row.id and <td> tag's data-key attribute. That's fine until it's time to send the selection back to Shiny, at which time we need to convert from the row number to the entry in `df.index`.

Actually, I suppose it's a question worth asking: what would you expect `input.grid_selected_rows()` to return: row numbers (to be used with `df.iloc`) or Pandas index values (to be used with `df.loc`)? I guess either is defensible? (cc @nealrichardson @machow @chendaniely)

---

**Update 2023-08-15 13:32-700:** Before this PR, the code was just wrong--using a row's row number to look up the `keyToIndex` map, which was actually not `key -> index` (where "index" is referring to the pandas index) but `str(index) -> index`. Hence the `None` being returned from the repro case in #676 (there were row numbers that did not exist in the pandas index). In other words, the only way the reported selection would be correct is if the pandas index is the set of integers `0..n` (which, to be fair, is probably the case with most pandas data frames, as it's the default).


jcheng5:
@andrie would love your opinion on 👆, it looked like your reprex had `df["col"][input.grid_selected_rows()]`

jcheng5:
Actually maybe using Pandas index values is not that realistic. They can theoretically be made up of almost any kind of Python object, can't they? I'm not sure we can successfully round-trip e.g. Python datetime values through JSON, and if they come back wrong, they won't be usable with `df.loc`.

andrie:
> @andrie would love your opinion on 👆, it looked like your reprex had `df["col"][input.grid_selected_rows()]` which implies row numbers I think.

I don't have a strong opinion on this.  I adapted the documented example at https://shinylive.io/py/examples/#interactively-excluding-data and tried to make minimal modifications:

``` python
    @reactive.Calc
    def filtered_df():
        selected_idx = list(req(input.summary_data_selected_rows()))
        countries = summary_df["country"][selected_idx]
        # Filter data for selected countries
        return df[df["country"].isin(countries)]
```

jcheng5:
After sleeping on it, I think row numbers are the way to go. They're also less pandas-specific, if we want more native support of, I don't know, DuckDB or Ibis in the future.

shiny/render/_dataframe.py
```diff
@@ -207,6 +207,10 @@ def to_payload(self) -> object:
 
 
 def serialize_pandas_df(df: "pd.DataFrame") -> dict[str, Any]:
+    # Currently, we don't make use of the index; drop it so we don't error trying to
+    # serialize it or something
+    df = df.reset_index(drop=True)
```
wch, line 212: Does it matter that this could alter the data frame that the user has passed in?

Perhaps when we do Pandas data frame operations, we should enable copy on write. This can be done in blocks of code like this:

```python
with pd.option_context('mode.copy_on_write', True):
    # Do operations here
    ...
```

(It would be dangerous to set the global option `pd.options.mode.copy_on_write = True`.)

wch (approved):
Looks good, but please see my comment about CoW operations when working with Pandas data frames.

shiny/render/_dataframe.py
wch: It looks like CoW was first available in Pandas 1.5.0, which was released about a year ago.
https://pandas.pydata.org/docs/dev/user_guide/copy_on_write.html

jcheng5: `reset_index` has an argument `inplace=False`. I assumed this means I'm not mutating the original?

jcheng5: Here's evidence: https://github.com/pandas-dev/pandas/blob/0f437949513225922d851e9581723d82120684a6/pandas/core/frame.py#L6151-L6154

jcheng5:
Rebased against master--had to do this interactively because the `e2e` folder changed to `tests/e2e` on main since this PR started.

shiny/render/_dataframe.py
wch: Sounds good - I was able to confirm by running the example from [here](https://pandas.pydata.org/docs/dev/user_guide/copy_on_write.html#description) in shinylive, and it does indeed make a copy. It's pretty confusing to figure out when Pandas makes a copy and when it doesn't.

```
>>> df = pd.DataFrame({"foo": [1, 2, 3], "bar": [4, 5, 6]})
>>> df2 = df.reset_index(drop=True)
>>> df2.iloc[0, 0] = 100
>>> df
   foo  bar
0    1    4
1    2    5
2    3    6
>>> df2
   foo  bar
0  100    4
1    2    5
2    3    6
```

PR #678, opened 2023-08-15
Fix pyright error

jcheng5:
Seems to have been introduce by new pandas type stubs (2.0.3)


wch:
Thanks. Can you also update shiny/api-examples/input_file/app.py ? It has the same contents as shiny/examples/input_file/app.py.

jcheng5:
Oh, wait, wait. `shiny/examples/input_file` shouldn't exist. If we want to make it a top-level example, it should be `examples/input_file`. (Do we?)

PR #719, opened 2023-08-30
Beginnings of typed accessors for client data and outputs

jcheng5:
This will be necessary for plotly automatic resizing, I believe

### DIFF
--- a/shiny/session/_session.py
+++ b/shiny/session/_session.py
@@ -138,6 +138,39 @@ class SessionMeta(type):
         return isinstance(__instance, SessionProxy)
 
 
+class OutputInfo:
+    _input: Inputs
+    _id: Id
+
+    def __init__(self, input: Inputs, id: Id):
+        self._input = input
+        self._id = id
+
+    # TODO: Consider what to do if the outputs are not present right now. Should we
+    # allow a default value? Cause by default, it will throw a silent error.
+
+    def width(self):
+        return self._input[f".clientdata_output_{self._id}_width"]()
+
+    def height(self):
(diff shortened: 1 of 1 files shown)


shiny/session/_session.py
```diff
@@ -138,6 +138,73 @@ def __instancecheck__(self, __instance: Any) -> bool:
...
+
+    def __init__(self, input: Inputs, id: Id):
+        self._input = input
+        self._id = id
+
+    # TODO: Consider what to do if the outputs are not present right now. Should we
+    # allow a default value? Cause by default, it will throw a silent error.
+
+    def width(self):
+        return self._input[f".clientdata_output_{self._id}_width"]()
```
wch, line 153: Instead of implementing individual functions, it might be better to implement a `__getattr__` method. That would provide a place to handle outputs that don't exist yet.

Another useful thing with a `__getattr__` method is that it could return `reactive.Value` object itself, instead of only providing a way to invoke `()` on the object. Although... that could also be accomplished `@property` decorator -- and using that decorator on each individual item would also allow more specific typing, like `Value[bool]`.

jcheng5: I don't like the `__getattr__` due to the inability to type or document each value (right?). The `@property` could work though. I don't love that `reactive.Value` exposes `.set()` that doesn't actually work, but the other methods on `Value` could be helpful.

cpsievert:
Clientdata is now available (wish I had seen this before!) https://shiny.posit.co/py/api/express/session.ClientData.html

Issue #708, opened 2023-08-28
Remove `session` argument from methods that utilize `require_active_session()`


schloerke:
Having the `session` is good for resolving the correct DOM `id` (such as when in modules), and for sending messages to the browser.

However, some update methods to not utilize the session being sent in: 
* https://github.com/rstudio/py-shiny/blob/0bc5ad07bac8dd418d27a848f0234ef6d79a2150/shiny/ui/_input_update.py#L226
* https://github.com/rstudio/py-shiny/blob/0bc5ad07bac8dd418d27a848f0234ef6d79a2150/shiny/ui/_input_update.py#L278

In addition, the session object is something that users shouldn't really be setting.

--------------------------

Proposal:
* Remove `session` from API
* Internally use `session = require_active_session(None)` to retrieve the session
* Show examples of using `with shiny.session.session_context(session=):` to set the context

Doing this would naturally resolve the two situations above and make the API cleaner.

schloerke:
For the rare case that someone is trying to update checkboxgroup choices or radio button choices outside of the module that their created in, the proposal above will fix that bug.

Current (broken) Situation:
```python
# from within "main" session
ui.update_checkbox_group("my_chk_group", choices=new_choices, session = my_module_session)
```

Current workaround / future solution:
```
# from within "main" session
with session.session_context(my_module_session):
	ui.update_checkbox_group("my_chk_group", choices=new_choices)
```

schloerke:
Reprex app for bug behavior:

[...]

PR #720, opened 2023-08-30
bug(module): Verify & add module support for experimental/main ui methods

### LINKED ISSUE #696: Use `resolve_id(id)` in `experimental` to resolve any `id` variables
Please make sure this works in modules--`resolve_id` may need to be called.

_Originally posted by @jcheng5 in https://github.com/rstudio/py-shiny/issues/680#issuecomment-1686915111_
            

This needs to apply to all of experimental and is out of scope for #680. So I made a new issue.

### LINKED ISSUE #712: Render plot type input type should be `object`, not `ImgData | None`
Introduced in #621

Previous: https://github.com/posit-dev/py-shiny/commit/da69ea055aad666c449a58676bdaa5bd4bb0850c#diff-3848a3fa2e11daf6f4de6a2be47908128df1ca342c8da4ccfc3692ecc2d53c88L190
```
RenderPlotFunc = Callable[[], object]
RenderPlotFuncAsync = Callable[[], Awaitable[object]]
```
which defines `IT` as `object`.

Current:
```
async def PlotTransformer(
    _meta: TransformerMetadata,
    _fn: ValueFn[ImgData | None],
    *,
    alt: Optional[str] = None,
    **kwargs: object,
) -> ImgData | None:
```
Which defines `IT` as `ImgData | None`

schloerke:
Related: #708 (comment)
Fixes #712
Fixes #696 

E2E test app added has a main session and two module sessions that have inputs and outputs. It makes sure that one module does not update the other module or the main session.


shiny/experimental/ui/_sidebar.py
```diff
@@ -207,6 +208,9 @@ def sidebar(
     if id is None and open != "always":
         # but always provide id when collapsible for accessibility reasons
         id = f"bslib-sidebar-{random.randint(1000, 10000)}"
+    else:
+        if id:
+            id = resolve_id(id)
```
wch, line 213: The structure of this conditional feels weird to me. How about this? I think it should result int he same logic:

```py
if id is not None:
    ...
elif open != "always":
    ...
```

But also, now that I think about it, shouldn't `id` always go through `resolve_id()`, even if it started out as `None`?

shiny/ui/_input_file.py
```diff
@@ -82,11 +82,12 @@ def input_file(
...
         accept = [accept]
 
+    resolved_id = resolve_id(id)
     btn_file = span(
         button_label,
         tags.input(
-            id=resolve_id(id),
-            name=id,
+            id=resolved_id,
+            name=resolved_id,
```
wch, line 90: I just want to make sure I understand this part... What is the purpose of `name` here?

wch (approved):
Just a couple of small things. Otherwise it looks good.

shiny/experimental/ui/_sidebar.py
schloerke: new 210 is needed and I like where it's at. But the else can be removed and we'll just use `resolve_id(id)`

shiny/ui/_input_file.py
schloerke: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name
> A string specifying a name for the input control. This name is submitted along with the control's value when the form data is submitted.
> 
> Consider the name a required attribute (even though it's not). If an input has no name specified, or name is empty, the input's value is not submitted with the form! 

--------------

The behavior also matches rstudio/shiny's implementation.
https://github.com/rstudio/shiny/blob/e87f942e893cc91905e0fdbe8a44daaf1e39020f/R/input-file.R#L103-L105


---------------

Testing it in the module app, there is no difference in the knowledge server-side.

* Upload file in regular session
* Upload file in module (`mod1`) session
```
INFO:     127.0.0.1:51695 - "POST /session/f9ee6d72ecd5c99fdc0078a0d2bd264ba87f869488fff0a88aadc516fcceaccf/upload/d9e9ab0a32d36f3b60ac991e?w= HTTP/1.1" 200 OK
[{'name': 'app.py', 'size': 18316, 'type': 'text/x-python-script', 'datapath': '/var/folders/gg/lzk06xmx6x3c0_9vqq1lcl9w0000gp/T/fileupload-ip_bz31f/tmp4bifgl2r/0.py'}]
INFO:     127.0.0.1:51708 - "POST /session/f9ee6d72ecd5c99fdc0078a0d2bd264ba87f869488fff0a88aadc516fcceaccf/upload/6447aa7becc96806f2cd7ba8?w= HTTP/1.1" 200 OK
[{'name': 'test_0696_resolve_id.py', 'size': 9194, 'type': 'text/x-python-script', 'datapath': '/var/folders/gg/lzk06xmx6x3c0_9vqq1lcl9w0000gp/T/fileupload-ip_bz31f/tmp4ymau36m/0.py'}]
```

PR #747, opened 2023-10-04
Self outputting renderers

jcheng5:
- [ ] TODO: Fix `@output_transformers` so that `@output_transformers` works again; currently `@output_transformers()` must be used.

### DIFF
--- a/shiny/render/transformer/_transformer.py
+++ b/shiny/render/transformer/_transformer.py
@@ -37,6 +37,7 @@ from typing import (
 
 if TYPE_CHECKING:
     from ...session import Session
+    from htmltools import TagChild
 
 from ..._docstring import add_example
 from ..._typing_extensions import Concatenate, ParamSpec
@@ -219,6 +220,7 @@ class OutputRenderer(Generic[OT], ABC):
         value_fn: ValueFn[IT],
         transform_fn: TransformFn[IT, P, OT],
         params: TransformerParams[P],
+        default_output: Callable[[str], TagChild] | None = None,
     ) -> None:
         """
         Parameters
@@ -233,6 +235,10 @@ class OutputRenderer(Generic[OT], ABC):
             function), then the function should execute synchronously.
         params
             App-provided parameters for the transform function (`transform_fn`).
+        default_output
+            Optional function that takes an `output_id` string and returns a Shiny UI
+            object that can be used to display the output. This allows render functions
+            to respond to `_repr_html_` method calls in environments like Jupyter.
 
         """
 
@@ -249,6 +255,7 @@ class OutputRenderer(Generic[OT], ABC):
         self._value_fn = value_fn
         self._transformer = transform_fn
         self._params = params
+        self.default_output = default_output
 
     def _set_metadata(self, session: Session, name: str) -> None:
         """
@@ -293,6 +300,24 @@ class OutputRenderer(Generic[OT], ABC):
         )
         return ret
 
+    def _repr_htmltools_(self) -> str | None:
+        import json
+
+        import htmltools
+
+        if self.default_output is None:
+            return None
+        res = htmltools.TagList(self.default_output(self.__name__)).render()
+        # TODO: Include deps too
+        return json.dumps(dict(html=res["html"]))
+
+    def _repr_html_(self) -> str | None:
+        import htmltools
+
+        if self.default_output is None:
+            return None
+        return htmltools.TagList(self.default_output(self.__name__))._repr_html_()
+
 
 # Using a second class to help clarify that it is of a particular type
 class OutputRendererSync(OutputRenderer[OT]):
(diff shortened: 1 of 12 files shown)


shiny/render/transformer/_transformer.py
```diff
@@ -220,6 +221,7 @@ def __init__(
         value_fn: ValueFn[IT],
         transform_fn: TransformFn[IT, P, OT],
         params: TransformerParams[P],
+        default_output: Callable[[str], TagChild] | None = None,
```
wch, line 224: I don't feel super strongly about this, but the name `default_output` makes me think that it's referring to an output, rather than a function that generates a corresponding output UI component. What do you think about calling it `default_output_ui`?

shiny/render/transformer/_transformer.py
```diff
@@ -294,6 +300,24 @@ async def _run(self) -> OT:
         )
         return ret
 
+    def _repr_htmltools_(self) -> str | None:
```
wch, line 303: Is this function used anywhere?

shiny/render/transformer/_transformer.py
```diff
@@ -294,6 +300,24 @@ async def _run(self) -> OT:
...
+        import json
+
+        import htmltools
+
+        if self.default_output is None:
+            return None
+        res = htmltools.TagList(self.default_output(self.__name__)).render()
+        # TODO: Include deps too
+        return json.dumps(dict(html=res["html"]))
```
wch, line 312: This looks similar to the functionality in https://github.com/posit-dev/py-htmltools/pull/61

wch (commented):
PR looks good overall -- I just have a few small questions and comments.

shiny/render/transformer/_transformer.py
jcheng5: Sure, that sounds good

shiny/render/transformer/_transformer.py
jcheng5: Hmmm, that's interesting. I think we're coming at this from two different ways because you're targeting Quarto and I'm targeting IPython. And I guess Shiny's `output_ui` counts as a third approach.

Have you considered using a `<html-dependency>` custom element? It hadn't occurred to me before, but maybe that could be a unifying abstraction?

shiny/render/transformer/_transformer.py
jcheng5: Yeah, my JupyterLab extension registers it as a new format.

shiny/render/transformer/_transformer.py
jcheng5: BTW, in a similar vein, I have been using a `<shiny-bind>` custom element to wrap "HTML that possibly contains Shiny inputs/outputs" instead of calling `Shiny.bindAll`/`Shiny.unbindAll`, and it has worked great so far.

The way I've done it, it only binds during `connectedCallback` and unbinds during `disconnectedCallback` so it still isn't great when code is arbitrarily mutating stuff.

shiny/render/transformer/_transformer.py
jcheng5: Never mind, removed this and handle it a different way now

shiny/render/transformer/_transformer.py
```diff
@@ -551,20 +609,20 @@ def output_transformer(
     function, and any keyword arguments supplied to the output tranformer decorator:
 
     * The first parameter to the handler function has the class
-      :class:`~shiny.render.transformer.TransformerMetadata` and is typically called
-      `_meta`. This information gives context the to the handler while trying to
-      resolve the app-supplied value function (typically called `_fn`).
+    :class:`~shiny.render.transformer.TransformerMetadata` and is typically called
+    `_meta`. This information gives context the to the handler while trying to
+    resolve the app-supplied value function (typically called `_fn`).
```
wch, line 614: Just curious: did you do this, or did the text editor do it automatically?

tests/e2e/default-render-ui/app.py
```diff
@@ -0,0 +1,17 @@
+from shiny import App, Inputs, Outputs, Session, render, ui
+
+app_ui = ui.page_fluid(ui.output_ui("dynamic_ui"))
+
+
+def server(input: Inputs, output: Outputs, session: Session):
+    @output
```
wch, line 7: This can be removed, right?
```suggestion
```

wch (approved):

shiny/render/transformer/_transformer.py
jcheng5: Oh, text editor. I didn't even realize.

wch: Weird, mine doesn't do that when I press Alt-Q. It leaves the indentation however it was before (either in line with the `*` or indented two spaces).

jcheng5: Mine too. No idea how that happened then

jcheng5: Fixed

PR #753, opened 2023-10-06
api!: Move most methods from `shiny.experimental.ui` into `shiny.ui`; Deprecate most of `shiny.experimental.ui`

schloerke:
Goal:
Move almost everything (if not everything), out of `shiny.experimental.ui` and into `shiny.ui`. Rearrange function locations as necessary to be pythonic. Deprecate every previously existing method in `shiny.experiemental.ui` that was ported. 

Method to not port:
- [x] Valuebox's showcase methods. This API will change in the near future
- [x] Card's `card_body()` method. It should either be necessary or not. Currently going with `not`, therefore skipping it from the public API.

Overarching Tasks:
- [x] Port from `shiny.experimental.ui` to `shiny.ui`
	- [x] Add deprecated methods for `shiny.experimental.ui`
- [x] Move api-examples from `./shiny/experimental/api-examples` to `./shiny/api-examples`
- [x] Update e2e tests
- [x] Do not use "deprecated experimental methods"
- Documentation
	- [x] Update method documentation to not use experimental
	- [x] Update quartodoc's config to expose newly ported methods
	- [x] Remove experimental methods from quartodoc's config

Followup PRs:
- Rename accordion methods (See https://github.com/posit-dev/py-shiny/issues/730)


wch (changes requested):

shiny/ui/_sidebar.py
```diff
@@ -326,32 +337,61 @@ def layout_sidebar(
 
     See Also
```
wch, line 338: `layout_sidebar` should drop `sidebar` parameter. It should inspect `*args` for the Sidebar object.

shiny/ui/_sidebar.py
```diff
@@ -416,7 +456,7 @@ def layout_sidebar(
     return CardItem(res)
 
 
-# TODO-maindocs; @add_example()
+@add_example()
 def toggle_sidebar(
```
wch, line 460: Change to `update_sidebar`.

shiny/ui/fill/_fill.py
```diff
@@ -0,0 +1,266 @@
...
+    * :func:`~shiny.ui.fill.is_fill_carrier`
+    * :func:`~shiny.ui.fill.is_fill_item`
+    * :func:`~shiny.ui.fill.is_fillable_container`
+    """
+    as_fillable_container(tag)
+    as_fill_item(tag)
+    return tag
+
+
+def remove_all_fill(
```
wch, line 117: Do we use this? Same for `is_fill` methods.

If no, maybe we should not export them.

wch (changes requested):

shiny/ui/__init__.py
```diff
@@ -113,44 +162,79 @@
...
+    "card_footer",
+    # _accordion
+    "AccordionPanel",
+    "accordion",
+    "accordion_panel",
+    "accordion_panel_open",
+    "accordion_panel_close",
+    "accordion_panel_insert",
+    "accordion_panel_remove",
+    "accordion_panel_set",
```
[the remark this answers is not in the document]
schloerke: Comment added in accordion code. Resolving

shiny/ui/_sidebar.py
cpsievert: I'd prefer to see `sidebar` as an (optional) keyword parameter, similar in spirit to what we have on the R side:

```r
layout_sidebar <- function(..., sidebar = NULL) {}
``` 

```python
def layout_sidebar(*args, sidebar = None):
  pass
```

I'm aware that this implies that the sidebar definition would have to come after the main content (e.g., `layout_sidebar("Main", sidebar = "Side")`, but here's some considerations that make me feel OK about that tradeoff:

1. It doesn't feel obviously wrong to me for the main content to come first.
2. Given there is a preference for sidebar first, this is fits into a class of problem that isn't isolated to `layout_sidebar()`. `page_*()` functions (e.g., `title`, etc) and anything else that wants to prioritize a collection of UI elements 1st and (optional) "parent" parameters 2nd (e.g., `layout_column_wrap(card(), card(), width=200)`) will also suffer from this problem, so it at least fits an "existing pattern" that'll hopefully feel familiar already.
3. With your suggestion (inspecting `*args`), it seems like it'd be relatively easy for a user to wrap a `sidebar()` in some other object and encounter an error or undefined behavior.
4. With your suggestion (inspecting `*args`), it becomes harder to align the R<->Python APIs. It's not obvious to me how we'd add support for this usage on R side (e.g., if we were to inspect `...` for a `sidebar()`, what should happen if a `sidebar` argument is also used?). That might be made easier by making a breaking change, but I'd _really_ hate to do that in order to fix something that isn't really a problem in R.

Also, more generally, if and when API changes are being proposed when porting R -> Python, @gadenbuie and I would appreciate being involved in those discussions/decisions.

schloerke: Addressing 3. ... the dev can return a `Sidebar` class.
Addressing 4. ... I don't believe bslib needs to change their API. There can be deviations. It is ok if they do not have the same exact interface for everything. There are fundamental difference in how we can express arguments in the two languages. Ex: The split between `TagAttrs` in an `*args` and `TagAttrValue` in `**kwargs` vs just `...` in R.

-----------------------

You and I talked about it yesterday and we couldn't decide on a straight path forward. I brought all motivations we had for the different approaches in conversation with @wch that afternoon. 

We'll discuss as a team about this method and a handful of others on Friday.

------------------------

In the interest of moving forward, I'm continuing with the current API and merging the PR.

PR #761, opened 2023-10-12
Run tests on Python 3.12

wch:
This change makes tests run on Python 3.12 on GitHub Actions.

My earlier attempt at this resulted in failures due to a problems building a package that shiny's tests depend on; this PR is here so that we can wait until that package is fixed and then merge.


wch:
Closing in favor of #1168.

PR #767, opened 2023-10-17
Add support for express mode apps

wch:
To install:

```py
# First uninstall old version
pip uninstall shiny

pip install git+https://github.com/posit-dev/py-shiny.git@flat-mode#egg=shiny
```

Express apps can be run with:

```
shiny run examples/express/app.py
```

You can also run it with an ASGI server like `uvicorn`, like so:

```
SHINY_EXPRESS_APP_FILE="examples/express/app.py" uvicorn shiny.express.app:app
```

This works because `shiny.express.app:app` is an ASGI application, and so ASGI servers like `uvicorn` will be able to use it. However, it is part of the shiny package and can't be user-modified, so the the way you tell it which file to use for the flat app source is by setting `SHINY_EXPRESS_APP_FILE`.


jcheng5:
I don't feel comfortable enough with the `@no_default_ui` decorator to push it to you branch, but here's the implementation:

```python
def no_default_ui(output_renderer: OutputRenderer[OT]) -> OutputRenderer[OT]:
    output_renderer.default_ui = lambda x: ""
    return output_renderer
```

Bool-on-OutputRenderer route is equally easy.

I put it in `shiny/render/transformer/_transformer.py` but then made it available from `shiny/render/__init__.py`.

js/page-output/page-output.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    if (this.originalBodyTagAttrs === null) {
+      // On the first run, store el's attributes so that on later runs we can clear
+      // any added attributes and reset this element to its original state.
+      this.originalBodyTagAttrs = Array.from(el.attributes);
+    } else {
+      // This is a later run. Reset attributes to their inital state.
+      for (const attr of this.originalBodyTagAttrs) {
+        el.setAttribute(attr.name, attr.value);
```
jcheng5, line 32: Need to also clear added attributes

js/page-output/page-output.ts
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+
+    let content = typeof data === "string" ? data : data.html;
```
jcheng5, line 36: data will never be a string

js/page-output/page-output.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    // Parse the HTML
+    const parser = new DOMParser();
+    const doc = parser.parseFromString(content, "text/html");
```
jcheng5, line 40: This might be a little slow but whatever

shiny/express/app.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import annotations
+
+from .._app import App
+from ._run import wrap_express_app
+
+app: App
+
+
+def __getattr__(name: str):
```
jcheng5, line 9: Overkill?

shiny/express/__init__.py
```diff
@@ -1,8 +1,74 @@
...
 
 __all__ = (
+    "input",
+    "output",
+    "session",
+    "is_express_app",
     "output_args",
     "suspend_display",
+    "wrap_express_app",
+    "app",
```
jcheng5, line 18: This seems private

shiny/express/_run.py
```diff
@@ -0,0 +1,243 @@
...
+
+    old_cm.__exit__(BaseException, None, None)
+    cm.__enter__()
+    _top_level_recall_context_manager = cm
+    _top_level_recall_context_manager_has_been_replaced = True
+
+    return old_cm
+
+
+class DetectShinyExpressVisitor(ast.NodeVisitor):
```
jcheng5, line 220: Consider rejecting if shiny.App() call is found

shiny/express/_run.py
```diff
@@ -0,0 +1,243 @@
...
+
+            traceback.print_exception(*sys.exc_info())
+            raise
+
+    app = App(app_ui, express_server)
+
+    return app
+
+
+def is_express_app(app: str, app_dir: str | None) -> bool:
```
jcheng5, line 74: Maybe have a way out as well--environment variable or something to force one mode or another.

shiny/express/__init__.py
```diff
@@ -1,8 +1,74 @@
...
+session: Session
+
+
+# Note that users should use `from shiny.express import input` instead of `from shiny
+# import express` and acces via `express.input`. The former provides a static value for
+# `input`, but the latter is dynamic -- every time `express.input` is accessed, it
+# returns the input for the current session. This will work in the vast majority of
+# cases, but when it fails, it will be very confusing.
+def __getattr__(name: str) -> object:
+    if name == "input":
```
jcheng5, line 35: Consider detecting if input, output, or session are requested more than once from the same scope. (using `inspect`?)

shiny/express/_run.py
```diff
@@ -0,0 +1,243 @@
...
+                "is not set."
+            )
+        file = Path(os.getcwd()) / app_file
+
+    # TODO: title and lang
+    app_ui = ui.page_output("__page__")
+
+    def express_server(input: Inputs, output: Outputs, session: Session):
+        try:
+            dyn_ui = run_express(file)
```
jcheng5, line 57:
```suggestion
    app_ui = run_express(file)

    def express_server(input: Inputs, output: Outputs, session: Session):
        try:
            dyn_ui = run_express(file)
```

jcheng5: The reasons for suggesting this (during live discussion):

1. HTML dependencies load with the page load which leads to smoother rendering (less fouc)

Downside:

1. Process startup is slightly slower

jcheng5: Oops, also would need to get rid of render.ui below

shiny/express/_run.py
```diff
@@ -0,0 +1,243 @@
...
+    ui_result: Tag | TagList = TagList()
+
+    def set_result(x: object):
+        nonlocal ui_result
+        ui_result = cast(Tag, x)
+
+    sys.displayhook = set_result
+
+    reset_top_level_recall_context_manager()
+    get_top_level_recall_context_manager().__enter__()
```
jcheng5, line 128: This should use `with`

shiny/express/_run.py
```diff
@@ -0,0 +1,243 @@
...
+    tree = DisplayFuncsTransformer().visit(tree)
+    tree = ast.fix_missing_locations(tree)
+
+    ui_result: Tag | TagList = TagList()
+
+    def set_result(x: object):
+        nonlocal ui_result
+        ui_result = cast(Tag, x)
+
+    sys.displayhook = set_result
```
jcheng5, line 125: Probably this too

shiny/express/_run.py
```diff
@@ -0,0 +1,243 @@
...
+
+    reset_top_level_recall_context_manager()
+    get_top_level_recall_context_manager().__enter__()
+
+    file_path = str(file.resolve())
```
jcheng5, line 130: Maybe require this to be absolute rather than resolving it here? Or not, I don't know.

shiny/express/_run.py
```diff
@@ -0,0 +1,243 @@
...
+    var_context: dict[str, object] = {
+        "__file__": file_path,
+        display_decorator_func_name: _display_decorator_function_def,
+    }
+
+    # Execute each top-level node in the AST
+    for node in tree.body:
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
+            exec(
+                compile(ast.Module([node], type_ignores=[]), file_path, "exec"),
```
jcheng5, line 141: AST loading, transformation, and compiling could be a separate step that we only do once. Then separate step for running each code block, which we do every time.

shiny/express/_run.py
```diff
@@ -0,0 +1,243 @@
...
+                var_context,
+                var_context,
+            )
+
+    get_top_level_recall_context_manager().__exit__(None, None, None)
+    return ui_result
+
+
+_top_level_recall_context_manager: RecallContextManager[Tag]
+_top_level_recall_context_manager_has_been_replaced = False
```
jcheng5, line 157: When you use a `with layout.foo()` function in the wrong setting, you get an unhelpful error message

shiny/express/_run.py
```diff
@@ -0,0 +1,243 @@
...
+                compile(ast.Interactive([node], type_ignores=[]), file_path, "single"),
+                var_context,
+                var_context,
+            )
+
+    get_top_level_recall_context_manager().__exit__(None, None, None)
+    return ui_result
+
+
+_top_level_recall_context_manager: RecallContextManager[Tag]
```
jcheng5, line 156: Consider making all this state/logic an object/methods

shiny/express/_recall_context.py
```diff
@@ -0,0 +1,78 @@
...
+            self.args.append(HTML(value._repr_html_()))  # pyright: ignore
+        else:
+            # We should NOT end up here for objects that were `def`ed, because they
+            # would already have been filtered out by _display_decorator_function_def().
+            # This is only for other kinds of expressions, the kind which would normally
+            # be printed at the console.
+            if value is not None:
+                self.args.append(tags.pre(repr(value)))
+
+    def __enter__(self) -> None:
```
jcheng5, line 48: The global context manager object could accumulate the results, and then only decide to create the page at the end.

shiny/express/_recall_context.py
```diff
@@ -0,0 +1,78 @@
...
+        if self.default_page is not None:
+            from . import _run
+
+            _run.replace_top_level_recall_context_manager(self.default_page)
+
+        self._prev_displayhook = sys.displayhook
+        # Collect each of the "printed" values in the args list.
+        sys.displayhook = self.append_arg
```
jcheng5, line 56: We should consider just replacing `sys.displayhook` once, at process start time, with a smart version of our own. And then have that smart version look at a different variable, backed by a contextvar, to decide where things actually go. This would let us keep "displayhooks" straight even when using async logic.

PR #772, opened 2023-10-19
Update value_box; Update to bootstrap 5.3; Update htmldeps

### LINKED ISSUE #782: Selectize & other 3rd party CSS should be re-compiled for Bootstrap 5
py-shiny currently copies over CSS (and JS) from (R) shiny (which is pre-compiled with Bootstrap 3 in mind).

In the build script, we do [re-compile CSS for ionRangeSlider](https://github.com/posit-dev/py-shiny/blob/e502a3f44fe6bd6854b733e1f61f35bc5dd55eb2/scripts/htmlDependencies.R#L129C9-L143) to target Bootstrap 5 (+ our styling tweaks). We should be doing a similar thing for `shiny.css`, `input_selectize()`, `input_date()`, etc...

schloerke:
Fixes #782

Packages:
- [x] htmltools
- [x] shiny
- [x] bslib
	- [x] new file structure
	- [x] Use new deps
	- [x] Updates for bootstrap 5.3


shiny/ui/_input_check_radio.py
```diff
@@ -124,19 +124,26 @@ def input_switch(
...
+        value,
+        width=width,
+    )
 
 
 def _input_checkbox(
     id: str,
     label: TagChild,
-    class_: str = "form-check",
+    class_: str = "bslib-input-checkbox",
```
wch, line 139: Should this class be on `input_checkbox`?

Also, why does this function exist? If it's simple to do, maybe `input_checkbox` should use this one. Or maybe we should rename this.

shiny/ui/_layout.py
```diff
@@ -78,6 +87,26 @@ def layout_column_wrap(
     """
     attrs, children = consolidate_attrs(*args, class_=class_, **kwargs)
 
+    if isinstance(width, MISSING_TYPE):
```
wch, line 90: Please add a note about when we can get rid of this.

wch (approved):

PR #775, opened 2023-10-20
Remove output from template app

gshotwell:
So we don't forget to remove the `@output` from the default app before the next release. I also added a `panel_title` instead of an `h2` so that the app shows up in the browser tab.


schloerke (approved):
Thank you!

wch (approved):

Issue #765, opened 2023-10-16
Option `--reload_dirs` have no effect


bryan-brancotte:
Hi,

When looking at `reload_dirs` variable, I think that the value provided by the user is not used : `setup_hot_reload` is called without `reload_dirs` [here](https://github.com/posit-dev/py-shiny/blob/main/shiny/_main.py#L283), and a few line later, reload_dirs is defined to an empty list at https://github.com/posit-dev/py-shiny/blob/54101ac4cdc38dec74a34a5e618f79316e1041d3/shiny/_main.py#L287

Thanks for your time

jcheng5:
You're right, looks like there was a bad merge here. @gshotwell can you please look at this as part of the overall uvicorn options PR?

PR #780, opened 2023-10-24
Kwargs to uvicorn run

gshotwell:
This PR does three things:
- Add `reload-excludes` argument to shiny run, for completeness with `reload-includes`
- Fixes [#765 ](https://github.com/posit-dev/py-shiny/issues/765)
- Pass kwargs to `uvicorn.run()` 

I have tested the following cases locally, but we decided that it wasn't worth adding automated tests: 
- Adding an `reload-dir` causes the app to reload when changes occur in that dir's files
- Setting `shiny run --reload --reload-includes="*.py" app.py` causes app to only reload when `.py` files change
- `shiny run --reload --reload-includes="*.py" --reload-excludes="test.py"  app.py` causes the app to reload when `app.py` is changed, but not when `test.py` is changed. 

I couldn't figure out a way to pass `**kwargs` to the command line, but I still think it's a useful safety valve to include.


shiny/_main.py
```diff
@@ -95,6 +96,13 @@ def main() -> None:
     help="File glob(s) to indicate which files should be monitored for changes. Defaults"
     f' to "{",".join(RELOAD_INCLUDES_DEFAULT)}".',
 )
+@click.option(
+    "--reload-excludes",
+    "reload_excludes",
+    default=",".join(RELOAD_INCLUDES_DEFAULT),
+    help="File glob(s) to indicate which files should be excluded from file monitoring. Defaults"
+    f' to "{",".join(RELOAD_INCLUDES_DEFAULT)}".',
```
schloerke, line 104:
```suggestion
    default=",".join(RELOAD_EXCLUDES_DEFAULT),
    help="File glob(s) to indicate which files should be excluded from file monitoring. Defaults"
    f' to "{",".join(RELOAD_EXCLUDES_DEFAULT)}".',
```

shiny/_main.py
```diff
@@ -172,11 +185,13 @@ def run_app(
     reload: bool = False,
     reload_dirs: Optional[list[str]] = None,
     reload_includes: list[str] | tuple[str, ...] = RELOAD_INCLUDES_DEFAULT,
+    reload_excludes: list[str] | tuple[str, ...] = RELOAD_EXCLUDES_DEFAULT,
     ws_max_size: int = 16777216,
     log_level: Optional[str] = None,
     app_dir: Optional[str] = ".",
     factory: bool = False,
     launch_browser: bool = False,
+    **kwargs: Dict[str, Any],
```
schloerke, line 194: This feels like we should make it a named argument of type `Dict[str, object]`. 

Ex:
```python
    unicorn_args: Dict[str, object]
```

I don't mind the name being explicit. By having a parameter, it can also give us opportunity for better typing down the road.

shiny/_main.py
```diff
@@ -172,11 +185,13 @@ def run_app(
     reload: bool = False,
```
schloerke, line 185: Because we are adding arguments to the middle of the positional args, I am inclined to add a `*` arg after `port`. Or even after `app`?

The followup args seem like they should always be named.

shiny/_main.py
```diff
@@ -207,6 +222,9 @@ def run_app(
     reload_includes
         List or tuple of file globs to indicate which files should be monitored for
         changes.
+    reload_excludes
+        List or tuple of file globas to indicate which files should be excluded from
```
schloerke, line 226:
```suggestion
        List or tuple of file globs to indicate which files should be excluded from
```

shiny/_main.py
```diff
@@ -207,6 +222,9 @@ def run_app(
     reload_includes
         List or tuple of file globs to indicate which files should be monitored for
         changes.
```
schloerke, line 224:
```suggestion
        changes. Can be combined with `reload_excludes`.
```

shiny/_main.py
```diff
@@ -261,6 +282,8 @@ def run_app(
 
     if reload_dirs is None:
         reload_dirs = []
+        if app_dir is not None:
+            reload_dirs = [app_dir]
```
schloerke, line 286: We should do a check below where we are calling `reload_dirs.append(app_dir)` to make sure `app_dir` is not already added. 

I can't suggest in this comment, so I'll put it here as code.

```python
        if app_dir and app_dir not in reload_dirs:
            reload_dirs.append(app_dir)
```

shiny/_main.py
schloerke: If we are not changing the `kwargs` to a named arg, then the typing should be updated to represent the type of each element.

```suggestion
    **kwargs: object,
```

shiny/_main.py
```diff
@@ -141,13 +149,16 @@ def run(
     reload: bool,
     reload_dirs: tuple[str, ...],
     reload_includes: str,
+    reload_excludes: str,
     ws_max_size: int,
     log_level: str,
     app_dir: str,
     factory: bool,
     launch_browser: bool,
+    **kwargs: Dict[str, Any],
```
schloerke, line 158:
```suggestion
    **kwargs: object,
```

schloerke (changes requested):

shiny/_main.py
gshotwell: Right, and realistically this is always called from the command line so they'll be named in that case anyway.

shiny/_main.py
gshotwell: The thought here is that it would let the user call something like this without wrapping the arguments in a dictionary. 

`shiny.run(app, uvicorn_arg1 = "foo", uvicorn_arg2 = "bar")` 

I'm happy with whatever, but since `shiny.run` is a pretty thin wrapper around `uvicorn.run` my intuition is that `*kwargs` is a bit better.

gshotwell: Resolved on Slack, we're going with `*kwargs`

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [UNRELEASED]
 
 ### New features
-
+* `shiny run` now takes `reload-includes` and `reload-excludes` to allow you to define which files trigger a reload.
+* `shiny.run` now passes keyword arguments to `uvicorn.run`
```
schloerke, line 13:
```suggestion
* `shiny run` now takes `reload-includes` and `reload-excludes` to allow you to define which files trigger a reload (#780).
* `shiny.run` now passes keyword arguments to `uvicorn.run` (#780).
```

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [UNRELEASED]
```
schloerke, line 10: Can we add a `### Breaking changes` section with a bullet on how `shiny.run()`'s app,host,port are positional args and everything else is now a keyword arg?

CHANGELOG.md
```diff
@@ -59,6 +60,9 @@ Methods still under consideration in `shiny.experimental.ui`:
 * `card(wrapper=)`, `card_body()`, `card_image()`, `card_header()`
 
 
+### Bug fixes
+* `shiny run` now respects the user provided `reload-dir` argument (#765)
```
schloerke, line 64:
```suggestion
* `shiny run` now respects the user provided `reload-dir` argument (#765).
```

CHANGELOG.md
```diff
@@ -59,6 +60,9 @@ Methods still under consideration in `shiny.experimental.ui`:
 * `card(wrapper=)`, `card_body()`, `card_image()`, `card_header()`
 
 
+### Bug fixes
```
schloerke, line 63: Can this section be combined with `### Bug fixes` below?

shiny/_main.py
```diff
@@ -261,10 +284,12 @@ def run_app(
...
     if reload_dirs is None:
         reload_dirs = []
+        if app_dir is not None:
+            reload_dirs = [app_dir]
 
     if reload:
         # Always watch the app_dir
-        if app_dir:
+        if app_dir and app_dir not in reload_dirs:
             reload_dirs.append(app_dir)
```
schloerke, line 293: `app_dir` is added to `reload_dirs` above iff `reload_dirs` is its default value and `app_dir` is supplied.
So, we should remove this enhancement.

```suggestion
```

schloerke (approved):
LGTM! (pending suggestions and changelog adjustments)
(Please squash merge when checks pass 🥳 )

dbkegley:
> I couldn't figure out a way to pass **kwargs to the command line, but I still think it's a useful safety valve to include.

Coming in just a little bit late on this one, but I was just looking for a way to pass `--root-path` to uvicorn and I saw that kwargs were added to `uvicorn.run()`.

I'm exposing my app behind a reverse proxy and `request.url_for` is returning the incorrect path.

https://github.com/encode/starlette/issues/604

Is there a workaround for passing in kwargs to `shiny run` or should I look at creating a uvicorn config file?

PR #786, opened 2023-10-26
Add output_args and suspend_display decorators

jcheng5:


shiny/express/_output.py
```diff
@@ -0,0 +1,125 @@
...
+    "suspend_display",
+)
+
+OT = TypeVar("OT")
+P = ParamSpec("P")
+R = TypeVar("R")
+CallableT = TypeVar("CallableT", bound=Callable[..., object])
+
+
+def output_args(
```
wch, line 22: No need to do this immediately since this PR is time-sensitive, but it will be helpful to have example apps.

shiny/render/transformer/_transformer.py
```diff
@@ -264,6 +264,8 @@ def __init__(
         self._transformer = transform_fn
         self._params = params
         self.default_ui = default_ui
+        self.default_ui_args: tuple[object, ...] = cast(tuple[object, ...], [])
```
wch, line 267: I think this `cast` is doing something different from what you want. The actual stored object is a list but you're telling the type checker to see it as a tuple.

It should be one of:

```suggestion
        self.default_ui_args: tuple[object, ...] = tuple()
```

```suggestion
        self.default_ui_args: tuple[object, ...] = ()
```

shiny/express/_output.py
```diff
@@ -0,0 +1,125 @@
...
+    if fn is None:
+        return suspend_display_ctxmgr()
+
+    # Special case for OutputRenderer; when we decorate those, we just mean "don't
+    # display yourself"
+    if isinstance(fn, OutputRenderer):
+        fn.default_ui = null_ui
+        return cast(Callable[P, R], fn)
+
+    return suspend_display_ctxmgr()(fn)
```
wch, line 107: Cool, I didn't realize you could use context managers this way.

shiny/render/transformer/_transformer.py
```diff
@@ -342,9 +344,16 @@ def _render_default(self) -> TagList | Tag | MetadataNode | str:
 
         params = tuple(inspect.signature(self.default_ui).parameters.values())
         if len(params) > 0 and params[0].name == "_params":
-            return self.default_ui(self._params.kwargs, self.__name__)  # type: ignore
+            return self.default_ui(
+                self._params.kwargs,
```
wch, line 348: I'm having a bit of trouble picturing this function call. Is it always true that `default_ui` can take a dictionary as the first argument?

I would have imagined that you would merge `self.default_ui_kwargs,` and `self._params.kwargs` and then pass the result in (with `**`, of course).

wch (approved):
Generally looks good, though I have a few comments/questions.

shiny/express/_output.py
jcheng5: Let's do that in the flat-mode branch, cause you can't really do anything interesting with them until then, unless you have quarto

shiny/render/transformer/_transformer.py
jcheng5: The conditional just before this is checking if the first param is named `_params`, that's a signal from the `default_ui` function to us that it wants the dictionary as the first argument, and inside of that function it will then do the rewriting.

Just writing that out felt bad, so I refactored it to be much more straightforward. Thanks for the feedback

PR #792, opened 2023-10-30
Fix default width/height with implicit plot output

jcheng5:
Previously, default width/height value from render.plot (None) was passed to implicit plot output. With this fix, use MISSING as the default parameter value to make it clear that the default value is not set and shouldn't be passed through.

### DIFF
--- a/shiny/render/_render.py
+++ b/shiny/render/_render.py
@@ -33,7 +33,7 @@ if TYPE_CHECKING:
 from .. import _utils
 from .. import ui as _ui
 from .._namespaces import ResolvedId
-from ..types import ImgData
+from ..types import MISSING, MISSING_TYPE, ImgData
 from ._try_render_plot import (
     PlotSizeInfo,
     try_render_matplotlib,
@@ -115,8 +115,8 @@ async def PlotTransformer(
     _fn: ValueFn[object],
     *,
     alt: Optional[str] = None,
-    width: Optional[float] = None,
-    height: Optional[float] = None,
+    width: Optional[float] | MISSING_TYPE = MISSING,
+    height: Optional[float] | MISSING_TYPE = MISSING,
     **kwargs: object,
 ) -> ImgData | None:
     is_userfn_async = is_async_callable(_fn)
@@ -124,6 +124,11 @@ async def PlotTransformer(
     session = _meta.session
 
     inputs = session.root_scope().input
+
+    if width is MISSING:
+        width = None
+    if height is MISSING:
+        height = None
 
     # We don't have enough information at this point to decide what size the plot should
     # be. This is because the user's plotting code itself may express an opinion about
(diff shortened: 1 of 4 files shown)


shiny/render/transformer/_transformer.py
```diff
@@ -800,33 +803,3 @@ async def resolve_value_fn(value_fn: ValueFn[IT]) -> IT:
 
 
 R = TypeVar("R")
-
-
-def decorator_args_passthrough(
```
jcheng5, line 805: This was an unused function, from an older implementation that I moved away from after @wch code review feedback.

tests/e2e/plot-sizing/test_plot_sizing.py
```diff
@@ -52,16 +52,3 @@ def test_output_image_kitchen(page: Page, local_app: ShinyAppProc) -> None:
                 tolerance += 2
             assert abs(rect["width"] - 300) <= tolerance
             assert abs(rect["height"] - 200) <= tolerance
-
-
-def test_decorator_passthrough_size():
```
jcheng5, line 57: Moved this from e2e test to regular pytest, since it didn't actually use any playwright logic.

tests/pytest/test_plot_sizing.py
```diff
@@ -0,0 +1,64 @@
...
+    def foo():
+        ...
+
+    rendered = str(foo.tagify())
+    assert "1280px" in rendered
+    assert "960px" in rendered
+    assert rendered == str(ui.output_plot("foo", width=1280, height=960))
+
+
+def test_decorator_plot_default():
```
jcheng5, line 19: This is the test case that reflects the issue the PR is meant to fix.

shiny/render/_render.py
```diff
@@ -115,8 +115,8 @@ async def PlotTransformer(
     _fn: ValueFn[object],
     *,
     alt: Optional[str] = None,
-    width: Optional[float] = None,
-    height: Optional[float] = None,
+    width: Optional[float] | MISSING_TYPE = MISSING,
+    height: Optional[float] | MISSING_TYPE = MISSING,
```
wch, line 119: Is `None` still a valid value?

If so, the intent would be clearer with `float | None | MISSING_TYPE`. Also, do the docs explain what `None` means?

```suggestion
    width: float | None | MISSING_TYPE = MISSING,
    height: float | None  | MISSING_TYPE = MISSING,
```

If not, take out the `Optional`. 

```suggestion
    width: float | MISSING_TYPE = MISSING,
    height: float | MISSING_TYPE = MISSING,
```

jcheng5: I think the difference only matters when you're using the implicit output. `None` means the implicit output shouldn't have any width/height attribute, `MISSING` means the implicit output should use its default width/height attribute. I can't really document that until we launch express mode, maybe for now I just say None or MISSING mean the same?

wch (approved):

PR #793, opened 2023-10-30
feat(api(session)): Start process of making `session` arguments removed from API

schloerke:
Related to #790

* Update `session` to have type `MISSING_TYPE` in functions where it could be resolved by using `with session_context(sess):`
* Remove `@output` and `session` from `server` definitions in examples (where possible)

- [x] Changelog entry

## App examples

Before:
```python

app_ui = ui.page_fluid(
    ui.input_slider("controller", "Controller", min=0, max=1, value=0, step=1),
    ui.input_checkbox("inCheckbox", "Input checkbox"),
    ui.output_text_verbatim("txt", placeholder = True),
)


    @reactive.Effect
    def _():
        # True if controller is odd, False if even.
        x_even = input.controller() % 2 == 1
        ui.update_checkbox("inCheckbox", value=x_even)

    @output
    @render.text
    def txt():
        return str(input.controller())

app = App(app_ui, server)
```

After:
```python
from shiny import App, Inputs, reactive, render, ui

app_ui = ui.page_fluid(
    ui.input_slider("controller", "Controller", min=0, max=1, value=0, step=1),
    ui.input_checkbox("inCheckbox", "Input checkbox"),
    ui.output_text_verbatim("txt", placeholder=True),
)


def server(input: Inputs):
    @reactive.Effect
    def _():
        # True if controller is odd, False if even.
        x_even = input.controller() % 2 == 1
        ui.update_checkbox("inCheckbox", value=x_even)

    @render.text
    def txt():
        return str(input.controller())


app = App(app_ui, server)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `@render.data_frame`'s input value `input.<ID>_column_filter()` has been deprecated. Please use `<ID>.filter()` to retrieve the same information. (#1374)
 
+* Many `shiny.ui` methods (and `reactive.invalidate_later`) have deprecated their `session` argument. This rarely used argument would pass the `Session` into the method, but it is now considered an anti-pattern. If a custom `session` value is needed, please execute your code within a session context like `with shiny.session.session_context(MY_SESSION):`.  (#793)
+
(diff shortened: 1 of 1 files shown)


shiny/reactive/_core.py
```diff
@@ -327,12 +333,19 @@ def invalidate_later(
...
 
         # If no session is provided, autodetect the current session (this
         # could be None if outside of a session).
-        session = get_current_session()
+        cur_session = get_current_session()
+    else:
+        # Display warning that they should use `session_context()`
+        session_type_warning()
+        # Use current session value
+        cur_session = session
```
schloerke, line 348: This is the only section that did not use `active_session = require_active_session()` cadence as an active session isn't required given the code. 

I'd argue a session should be required to call `invalidate_later(delay)`

PR #796, opened 2023-10-30
Keep version number to 3 components

jcheng5:
Quarto uses semver rules to parse our version number, and chokes if it contains four elements


wch:
I believe this won't be necessary because Quarto changed its version check to just use the first two numbers.

https://github.com/quarto-dev/quarto-cli/commit/6f7cc03544f5e96bb7008fe586c4af74d8e66664

PR #800, opened 2023-11-03
Un-pin pyright version and fix type issues

wch:


shiny/_autoreload.py
```diff
@@ -19,17 +19,6 @@
...
-
-if not is_pyodide:
-    # Warning: `import websockets` alone doesn't work because the websockets package uses
-    # a lazy loading technique which is opaque to mypy and pyright.
-    # https://github.com/aaugustin/websockets/issues/940#issuecomment-874012438
-    from websockets.client import connect
-    from websockets.server import serve, WebSocketServerProtocol
-    import websockets.exceptions
-    import websockets.datastructures
-
```
wch, line 32: I moved these imports into the functions where they're actually used. This allowed us to get rid of the `is_pyodide` check. 

Also, with the most recent version of websockets, 12.0, we can switch back to `import websockets` instead of importing the individual objects. https://github.com/python-websockets/websockets/issues/940#issuecomment-1742118059

PR #812, opened 2023-11-16
Custom component templates

nstrayer:
Flesh out workflow for custom templates and modularize logic for template building a tiny bit.

### DIFF
--- a/shiny/_template_utils.py
+++ b/shiny/_template_utils.py
@@ -1,0 +1,31 @@
+import shutil
+import sys
+from pathlib import Path
+
+# import questionary
+# from questionary import Choice
+
+
+def copyTemplateFiles(dest: str, template: str):
+    app_dir = Path(dest)
+    template_dir = Path(__file__).parent / "templates" / template
+    duplicate_files = [
+        (app_dir / file.name).exists() for file in template_dir.iterdir()
+    ]
+
+    if any(duplicate_files):
+        print(
+            f"Error: Can't create new files because the following files already exist in the destination directory: {duplicate_files}"
+        )
+        sys.exit(1)
+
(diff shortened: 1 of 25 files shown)


shiny/_template_utils.py
```diff
@@ -0,0 +1,150 @@
...
+    # Define the control flow for the top level menu
+    # This is simple now but will get more complicated as we add more templates
+    flow_dispatch = {
+        "cancel": lambda: sys.exit(),
+        "js-component": lambda: jsComponentQuestions(),
+        "default": lambda: appTemplateQuestions(template),
+    }
+
+    # Call the appropriate function based on the user's choice
+    flow_dispatch.get(template, flow_dispatch["default"])()
```
nstrayer, line 41: I'm curious about people's thoughts on this. It feels nice and extensible but also may be trying to be too clever. @schloerke @gshotwell

shiny/_template_utils.py
```diff
@@ -0,0 +1,150 @@
...
+    print("- Run `npm run build` to build the component")
+    print("- Install package locally with `pip install -e .`")
+    print("- Open and run the example app in the `example-app` directory")
+
+
+def choicesArray(choices: list[tuple[str, str]]):
+    """
+    Convert tuple of key-value pairs for Questionary questions to a list of Choice objects
+    """
+    return [Choice(name, value=value) for name, value in choices]
```
nstrayer, line 121: This may be unnecessary abstraction but I found it annoying to always have to type that list comprehension out

shiny/_template_utils.py
gshotwell: Seems sensible to me

gshotwell (approved):

PR #813, opened 2023-11-16
Shiny templates

gshotwell:
# New dependency
This PR adds a dependency to the [questionary package](https://github.com/tmbo/questionary). I chose this dependency for a few reasons:
- It's fairly mature (five years old, on version 2.0)
- We're mostly going to want fairly simple CLI interfaces and so a high level package like questionary is better than the  lower-level [prompt-toolkit](https://github.com/prompt-toolkit/python-prompt-toolkit)
- It more popular than [PyInquirer](https://github.com/CITGuru/PyInquirer)
- It appears to be professionally maintained by a company called [Rasa](https://rasa.com/about/) and has type annotation and CI

# Description
This adds a templating process to `shiny create`  there are currently two types of templates:

### Shiny application templates
These are based on common requests from Discord, and are meant to provide various levels of complexity.
- Basic app: Same as our current `shiny create`
- Express app: Our hello world app with shiny express
- Dashboard: a basic dashboard with a sidebar, value boxes and cards. This is meant to show people the new UI components, as well as how to use `reactive.Calc` to share data between output components
- Multi-page app: A full production app. This illustrates how to use modules, how to pass arguments to modules, and how to add your own CSS styling. It is also the only template which is type-annotated

### Javascript component templates
These are templates for people to build their own shiny extension packages using Javascript. They're developed by @nstrayer and have three types::
- Input component
- Output component
- React component

# TODOs
- [x] Confirm that we want to take this dependency
- [x] Confirm that these are the templates we want
- [x] Add READMEs to the templates describing what they are and where to learn more about the functionality.
- [x] QA review (make sure this works on windows, Workbench, other editors)
- [ ] ~Add back and cancel buttons to basic app file path entry~
- [ ] ~Provide user feedback when NPM install happens~
- [x] Prevent special characters for component names
- [x] Make sure that component name is pypi compatitable
- [x] Dashboard app, max slider value should be 5000 (to prevent empty filters in dataframe)
- [x] Add node_modules to .gitignore?
- [x] Validate to avoid `"my-component"`
- [x] Update CHANGELOG.md


jcheng5:
Am I doing this wrong? (newlines added for readability)

```
(.venv) ➜  py-shiny-templates git:(shiny-templates) shiny create

? Enter destination directory: ./foobar

  File "/Users/jcheng/Development/posit-dev/py-shiny-templates/.venv/bin/shiny", line 33, in <module>
    sys.exit(load_entry_point('shiny', 'console_scripts', 'shiny')())
  File "/Users/jcheng/Development/posit-dev/py-shiny-templates/.venv/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/Users/jcheng/Development/posit-dev/py-shiny-templates/.venv/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/Users/jcheng/Development/posit-dev/py-shiny-templates/.venv/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/jcheng/Development/posit-dev/py-shiny-templates/.venv/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/jcheng/Development/posit-dev/py-shiny-templates/.venv/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/Users/jcheng/Development/posit-dev/py-shiny-templates/shiny/_main.py", line 470, in create
    template_query()
  File "/Users/jcheng/Development/posit-dev/py-shiny-templates/shiny/_template_utils.py", line 57, in template_query
    flow_dispatch.get(template, flow_dispatch["default"])()
  File "/Users/jcheng/Development/posit-dev/py-shiny-templates/shiny/_template_utils.py", line 53, in <lambda>
    "default": lambda: appTemplateQuestions(template),
  File "/Users/jcheng/Development/posit-dev/py-shiny-templates/shiny/_template_utils.py", line 67, in appTemplateQuestions
    app_dir = copyTemplateFiles(appdir, template)
  File "/Users/jcheng/Development/posit-dev/py-shiny-templates/shiny/_template_utils.py", line 146, in copyTemplateFiles
    duplicate_files = [
  File "/Users/jcheng/Development/posit-dev/py-shiny-templates/shiny/_template_utils.py", line 146, in <listcomp>
    duplicate_files = [
  File "/Users/jcheng/.pyenv/versions/3.10.6/lib/python3.10/pathlib.py", line 1017, in iterdir
    for name in self._accessor.listdir(self):
FileNotFoundError: [Errno 2] No such file or directory: '/Users/jcheng/Development/posit-dev/py-shiny-templates/shiny/templates/start'
```

jcheng5:
I don't have a problem with questionary, seems like a pretty minimal dependency. I think if we get much more complicated than it is now, we might want to move to [Textual](https://textual.textualize.io/) instead.

jcheng5:
Do you feel strongly about these being bundled in the shiny package itself, versus living in a git repo and being cloned? 

Reasons for not bundling:

1. I'm a little more sensitive to package size bloat because of shinylive--and I'd eventually like us to have dozens of templates
2. Would be nice to change templates on a totally different schedule from shiny

Reasons for bundling:

1. Offline use. But for the authoring case I'm not sure how important that is.
2. Easier to ensure that `shiny create` always makes an app that is compatible with your exact version of Shiny.

We can also bundle now and unbundle later.

jcheng5:
Regarding the templates themselves:

**Basic/Express:** I'd much prefer these be one "Basic app" choice, with a description that indicates you should choose this if you want to start from scratch (which is very different from "Basic" in the sense that "I don't know Shiny that well and need something basic"). And then after you choose it, then it asks if you want Express or regular syntax, with a basic description (really a reminder, hopefully) of the difference.

**Dashboard:** Fantastic. I think instead of two plots we should have one plot and one data_frame. We are going to want multiple variations on this showing off different features; we'll for sure want a variant with plotly and ipyleaflet/folium, we'll want ones that show common layouts, etc. Inspiration: https://quarto.org/docs/dashboards/examples/

**Multi-page app with modules:** The second page showed an error for me.

jcheng5:
Other notes:
1. Can you make `Cancel` and `Back` options visually distinct, like `[Cancel]` and `[Back]` maybe?
2. Hitting Ctrl+C gave incorrect behavior, differing depending which step I was on but never the right thing (cleanly exit with a non-zero exit code).
3. "What do you want to name your component?" should not accept an empty answer. The `E.g. 'my-component-name'` made me think I could hit Enter and it would default to my-component-name, instead it led to a downstream error.
4. In my case, the installing of NPM dependencies failed (because I use nvm) which is fine but it then said it succeeded, which is less fine.
```
? Do you want to install js dependencies now? Yes
Installing NPM deps now...
An error occurred while installing NPM dependencies: [Errno 2] No such file or directory: 'npm'
Successfully created foo component package!
Next steps:
- Run `cd foo` to change into the new directory
- Run `npm run build` to build the component
- Install package locally with `pip install -e .`
- Open and run the example app in the `example-app` directory
```

jcheng5:
We'll have a complementary web page on the website that will let users browse these visually. We should think about what action we expect users to take after they select a template from the website. Do we give them a `git clone` command to copy-and-paste, or a `shiny create -t dashboard-plotly-drilldown` type command?

jcheng5:
After we release this we need to be sure to update the places on the Shiny website where we tell people to run `shiny create my_app`. That doesn't work anymore (maybe it should?) and probably we want to tell them to go through the wizard anyway.

shiny/_template_utils.py
```diff
@@ -0,0 +1,166 @@
...
+    # Define the control flow for the top level menu
+    # This is simple now but will get more complicated as we add more templates
+    flow_dispatch = {
+        "cancel": lambda: sys.exit(),
+        "js-component": lambda: jsComponentQuestions(),
+        "default": lambda: appTemplateQuestions(template),
+    }
+
+    # Call the appropriate function based on the user's choice
+    flow_dispatch.get(template, flow_dispatch["default"])()
```
jcheng5, line 57: This feels like unnecessary complexity even given the comment, but maybe I'm missing something about how we're expecting this to evolve.

```suggestion
    # Define the control flow for the top level menu
    if template is None or template == "cancel":
        sys.exit(1)
    elif template == "js-component":
        jsComponentQuestions()
        return
    else:
        appTemplateQuestions(template)
```

jcheng5, line 57: If we don't go the way I suggested, at the very least, I don't like `"default"` being one of the `flow_dispatch` keys as it will never be the value of `template`.

```suggestion
    # Define the control flow for the top level menu
    # This is simple now but will get more complicated as we add more templates
    flow_dispatch = {
        "cancel": lambda: sys.exit(),
        "js-component": lambda: jsComponentQuestions(),
    }

    # Call the appropriate function based on the user's choice
    flow_dispatch.get(template, lambda: appTemplateQuestions(template))()
```

shiny/_template_utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def appTemplateQuestions(template: str):
+    appdir = questionary.path(
+        "Enter destination directory:",
+        default="./",
```
jcheng5, line 63: Should use `f".{os.path.sep}"` here for portability

shiny/_template_utils.py
nstrayer: Very fair. The problem of dispatching to a given action based on a string is one I always want to reach for a `switch` statement with but, as I hinted at in the comment: it's most likely overkill here. I think I had some idea of a future where the branches got complex and we used this dispatch object to control navigation and backing up etc based on a path of "keys" used to get to a given step.

nstrayer: Definitely. Was a poor thought out key and also indicative of my desire to strong-arm a switch statement into a language that doesnt have one.

shiny/_template_utils.py
nstrayer: https://github.com/posit-dev/py-shiny/pull/813/commits/9f6076f62213f7195a200a2db6a167eb54a7fcf9

nstrayer:
> Other notes:
> ...
> 3. "What do you want to name your component?" should not accept an empty answer. The `E.g. 'my-component-name'` made me think I could hit Enter and it would default to my-component-name, instead it led to a downstream error.
> ...

Fixed this by setting up a proper Validation class. Pretty cool and powerful. https://questionary.readthedocs.io/en/stable/pages/advanced.html#validation

https://github.com/posit-dev/py-shiny/pull/813/commits/fe537337b84cdedf69e9e0bb6eeb4a3a0cf04e0f

nstrayer:
> Can you make Cancel and Back options visually distinct, like [Cancel] and [Back] maybe?

Done in [20d1b37](https://github.com/posit-dev/py-shiny/pull/813/commits/20d1b372d30946889ddeebbc197f20bb35bd2b15). Added a back arrow to the Back option and the requested boxing for Cancel. 
image

nstrayer:
> 4. In my case, the installing of NPM dependencies failed (because I use nvm) which is fine but it then said it succeeded, which is less fine.

Fixed in [0564b0b](https://github.com/posit-dev/py-shiny/pull/813/commits/0564b0b7672ba249e58ab3cc63bec56502427680) by adding better messages. We've already written the template to the user's disk at this point so the template stays but we warn them they may need to install NPM or node etc..

```python
...
   if success_intalling_deps:
        print("Successfully installed NPM dependencies.")
    else:
        print(
            "Error: Failed to install NPM dependencies. You may need to install Node/npm. After doing this you can try installing manually."
        )
 ...
```

jcheng5:
Re: Ctrl+C, handling None is great for now. If the UX gets a lot more complicated in the future and Ctrl+C always means “exit”, we also have the option of telling questionary to let the KeyboardInterrupt bubble up and we just catch it at the top level (I saw this in their docs, it’s called “unsafe” mode). No need to do this now though, just wanted to note it for the future.

gshotwell:
Based on Joe's comments I did the following:
- Remove `express app` I think we should change the basic app to an express app, but should do that at a later date when we really role out shiny express
- Updated dashboard to display a table

I think we should include  these templates with the package just so that we can test that they work and avoid version mismatch issues. It might make sense to have github-based templates in the future, we could add those with another submenu. Another option would be to have a `shiny-templates` package that just has a ton of templates.

nealrichardson:
> We'll have a complementary web page on the website that will let users browse these visually. We should think about what action we expect users to take after they select a template from the website. Do we give them a `git clone` command to copy-and-paste, or a `shiny create -t dashboard-plotly-drilldown` type command?

+1 for a CLI flag like this to provide a template name so you don't have to do this interactively. (Maybe you've already done this, I haven't looked.) You could even drop the interactive package dependency altogether if either `shiny create --template` with no template specified, or with `--help` or something, listed the available templates.

wch:
One thing I noticed is that `import shiny` causes questionary to be loaded. In shinylive, this results in about 770Kb of data to be loaded in the browser, for something that can't be used in that environment. I'll add a couple of suggestions about how to avoid that.

shiny/_template_utils.py
```diff
@@ -0,0 +1,200 @@
...
+def build_path_string(*path: str):
+    """
+    Build a path string that is valid for the current OS
+    """
+    # If no args are provided we should add an empty path to the list
+    # so that the result is properly formed. E.g. "./" on Unix systems
+    prefix = f".{os.path.sep}"
+    if len(path) == 0:
+        return prefix
+    return f"{prefix}{str(Path(*path))}"
```
wch, line 171: I think this can just be:

```suggestion
    return os.path.join(".", *path)
```


The existing behavior:

```
>>> shiny._template_utils.build_path_string()
'./'
>>> shiny._template_utils.build_path_string("a")
'./a'
>>> shiny._template_utils.build_path_string("a", "b")
'./a/b'
>>> shiny._template_utils.build_path_string("a/b", "c")
'./a/b/c'
```

The one thing I see above that's a bit inconsistent is that when there's no input, returns `./` with a trailing slash, whereas with all the other cases, it doesn't have a trailing slash.


With my suggested change:

```
>>> shiny._template_utils.build_path_string()
'.'
>>> shiny._template_utils.build_path_string("a")
'./a'
>>> shiny._template_utils.build_path_string("a", "b")
'./a/b'
>>> shiny._template_utils.build_path_string("a/b", "c")
'./a/b/c'
```

The only difference in output is the `.` vs `./`, but I think that's actually more internally consistent and will lead to more predictable behavior (e.g. if you append a path string and forget to use a leading slash, it will always result in an error, instead of only in some cases).

wch (changes requested):

wch:
Overall, this looks good to me. Just a few small comments and suggestions.

gshotwell:
@nealrichardson I added a template argument, but it turns out it's super awkward to get nice looking help/discovery text because it resizes to the terminal window. My plan for this is that we will point people to the website from the help text when we get that up and running. I think that should be okay because mostly we expect people to discover the templates through either the prompts or the website.  

Here's what it looks like now:

```
Usage: shiny create [OPTIONS]

  Create a Shiny application from a template.

  Create an app based on a template. You will be prompted with a number of
  application types, as well as the destination folder. If you don't provide
  a destination folder, it will be created in the current working directory
  based on the template name.

  After creating the application, you use `shiny run`:

      shiny run APPDIR/app.py --reload

Options:
  -t, --template [basic-app|dashboard|multi-page|js-component|js-input|js-output|js-react]
                                  Choose a template for your new application.
  --help                          Show this message and exit.
  ```

jcheng5 (approved):

PR #816, opened 2023-11-17
Use path to app as part of entrypoint for Express apps

wch:
For running Express apps, this PR makes it so that setting an environment variable is no longer needed for uvicorn to launch the app; instead, if the app path is:

```
path/to/app.py
```

then this sets the entrypoint to

```
shiny.express.app:path_2f_to_2f_app_2e_py
```

This will make it possible to deploy on shinyapps.io.

The escaping rule is this:
- For all characters that are _not_ numbers or letters escape them by replacing with `_<hex>_`, where `<hex>` is the hexadecimal representation of the character.
- In addition, if the first character is a number, escape it the same way. This is because variable names in Python can't start with a number. For example, `0app_2e_py` is not a valid name, and so can't be used as part of an entrypoint, as in  `shiny.express.app:0app_2e_py`.

In my testing, I found that uvicorn actually would accept other characters, like if you run:

```py
uvicorn shiny.express.app:path/to/app.py
```

However, `gunicorn` would error if you run:

```py
gunicorn shiny.express.app:path/to/app.py
```

because in `gunicorn` [tries to parse](https://github.com/benoitc/gunicorn/blob/02d3dd8b7ae28816a37623077388dbcfdbfd4379/gunicorn/util.py#L387-L393) the part after the `:` as an `ast.Name` or `ast.Call`, and `path/to/app.py` contains characters that are invalid for either of those types.

I decided to play it safe and make it general by escaping to valid Python variable names.

### DIFF
--- a/shiny/_main.py
+++ b/shiny/_main.py
@@ -18,10 +18,11 @@ import uvicorn
 import uvicorn.config
 
 import shiny
-from shiny.express import is_express_app
 
 from . import _autoreload, _hostenv, _static, _utils
 from ._typing_extensions import NotRequired, TypedDict
+from .express import is_express_app
+from .express._utils import escape_to_var_name
 
 
 @click.group("main")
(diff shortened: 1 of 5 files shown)


jcheng5:
This is impressively pragmatic 🎩

PR #819, opened 2023-11-21
Update to latest CSS/JS assets

cpsievert:
Updates to sync up with the recent bslib v0.6.1, shiny v1.8.0, and htmltools v0.5.7 releases


schloerke (approved):
LGTM. Please update the script file to display where the packages are being downloaded from.

PR #822, opened 2023-11-27
Add lowercase `reactive.value`, `calc`, `effect`

### LINKED ISSUE #814: Should add lowercase aliases for `reactive.value`, `reactive.calc`, `reactive.effect`
There are technical reasons for having the capitalized names `reactive.Value`, `reactive.Calc`, an `reactive.Effect`, but from a user perspective, it is confusing. So we should add lowercase aliases for them.

wch:
This closes #814.

Many users have wondered why `reactive.Value`, `.Calc`, and `.Effect` are capitalized. There are technical reasons for this, but from a user perspective it makes more sense to just use lowercase names.

This PR changes those names to lowercase, and adds aliases so that the old capitalized names still work. 

This PR does not change any examples because if users on a previous version of shiny were to use the example code, it wouldn't work for them, and it would be unclear to them why. In a future release, we should update the examples to use the new lowercase names.

### DIFF
--- a/shiny/reactive/__init__.py
+++ b/shiny/reactive/__init__.py
@@ -8,10 +8,13 @@ from ._core import (  # noqa: F401
 )
 from ._poll import poll, file_reader
 from ._reactives import (  # noqa: F401
+    value,
     Value,
+    calc,
     Calc,
     Calc_,  # pyright: ignore[reportUnusedImport]
     CalcAsync_,  # pyright: ignore[reportUnusedImport]
+    effect,
     Effect,
     Effect_,  # pyright: ignore[reportUnusedImport]
     event,
@@ -26,8 +29,11 @@ __all__ = (
     "on_flushed",
     "poll",
     "file_reader",
+    "value",
     "Value",
+    "calc",
     "Calc",
+    "effect",
     "Effect",
     "event",
 )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### New features
 
+* Closed #814: The classes `reactive.Value`, `reactive.Calc`, and `reactive.Effect` have been changed to have lowercase names: `reactive.value`, `reactive.calc`, and `reactive.effect`. The old capitalized names are now aliases to the new lowercase names, so existing code will continue to work. The examples have not been changed yet, but will be changed in a future release. (#822)
+
 ### Bug fixes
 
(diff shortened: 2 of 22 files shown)


shiny/reactive/__init__.py
```diff
@@ -26,8 +29,11 @@
     "on_flushed",
     "poll",
     "file_reader",
+    "value",
     "Value",
+    "calc",
     "Calc",
+    "effect",
     "Effect",
```
schloerke, line 37: Should the old names be removed and pyright ignored above?

```suggestion
    "calc",
    "effect",
```

shiny/reactive/__init__.py
```diff
@@ -8,10 +8,13 @@
...
 from ._poll import poll, file_reader
 from ._reactives import (  # noqa: F401
+    value,
     Value,
+    calc,
     Calc,
     Calc_,  # pyright: ignore[reportUnusedImport]
     CalcAsync_,  # pyright: ignore[reportUnusedImport]
+    effect,
     Effect,
```
schloerke, line 18: (Continuation of comment below)

```suggestion
    Value,  # pyright: ignore[reportUnusedImport]
    calc,
    Calc,  # pyright: ignore[reportUnusedImport]
    Calc_,  # pyright: ignore[reportUnusedImport]
    CalcAsync_,  # pyright: ignore[reportUnusedImport]
    effect,
    Effect,  # pyright: ignore[reportUnusedImport]
```

shiny/reactive/__init__.py
wch: I think we need to keep the old names for backward compatibility, at least for the time being. If we were to remove them now, then a lot of user code would get flagged by pyright.

jcheng5:
I think we have consensus that `@reactive.calc` and `@reactive.effect`, as decorators, should be lower-cased.

I'm a little surprised to see `reactive.value()` being lower-cased as well, as that's just a plain old constructor?

jcheng5:
And IF we do really want `reactive.value` to be lower-cased, I think we should still have `class Value` and then create a `value` alias (or even make a wrapper function `def value(x: T) -> reactive.Value[T]`) so that it still makes sense to have `reactive.Value` for type annotations.

wch:
I think we should provide a lowercase version of `value`, because if we don't, I think we'll get the same questions about "why is this thing capitalized?" But I don't feel strongly about it.

I think it makes sense to keep the `Value` class, and then make `value` an alias to it.

So in short, I can roll back the `Value`->`value` changes, and then just add `value` as an alias. Then if we decide we don't like the alias, we can take it out, and it'll be just a one-line change.

jcheng5 (approved):

PR #828, opened 2023-11-28
Add fixes for type stub generation

wch:
When shiny type stubs are generated by pyright, and they are used for types (instead of using the shiny source code directly), we have some typing errors, which are visible in shinylive:

image

This is because some of the .pyi files generated by pyright lack necessary type information. This PR fixes most cases of missing types in the .pyi files, but not all.

@schloerke in the future we should have an automated test which runs pyright on the generated .pyi files, as well as the examples (but using the shiny type stubs instead of shiny sources directly).

******

To illustrate what's going on, here's the relevant part of [`OutputTransformer`](https://github.com/posit-dev/py-shiny/blob/d541388a2c82750bcbae4c2f8731447cb37ee71a/shiny/render/transformer/_transformer.py#L541-L604) from _transformer.py before the change:

```py
class OutputTransformer(Generic[IT, OT, P]):
    def __init__(
        self,
        fn: OutputTransformerFn[IT, P, OT],
    ) -> None:
        self._fn = fn
        self.ValueFn = ValueFn[IT]
        self.OutputRenderer = OutputRenderer[OT]
        self.OutputRendererDecorator = OutputRendererDecorator[IT, OT]
```

This is the content of the resulting _transformer.pyi file:

```py
class OutputTransformer(Generic[IT, OT, P]):
    def __init__(self, fn: OutputTransformerFn[IT, P, OT]) -> None:
        ...
```

Notice that it loses all information about `self._fn`, `self.ValueFn`, `self.OutputRenderer`, and `self.OutputRendererDecorator`.

This is what the [`OutputTransformer`](https://github.com/posit-dev/py-shiny/blob/c9415525a756e7d9c8691f419bd33752d0dad402/shiny/render/transformer/_transformer.py#L542-L610) looks like after the change:

```py
class OutputTransformer(Generic[IT, OT, P]):
    fn: OutputTransformerFn[IT, P, OT]
    ValueFn: Type[ValueFn[IT]]
    OutputRenderer: Type[OutputRenderer[OT]]
    OutputRendererDecorator: Type[OutputRendererDecorator[IT, OT]]

    def __init__(
        self,
        fn: OutputTransformerFn[IT, P, OT],
    ) -> None:
        self._fn = fn
        self.ValueFn = ValueFn[IT]
        self.OutputRenderer = OutputRenderer[OT]
        self.OutputRendererDecorator = OutputRendererDecorator[IT, OT]
```

The class-level variables like `fn` and `OutputRenderer` aren't actually used by any of the code, but pyright uses them for generating type stubs. This is the resulting _transformer.pyi file:

```py
class OutputTransformer(Generic[IT, OT, P]):
    fn: OutputTransformerFn[IT, P, OT]
    ValueFn: Type[ValueFn[IT]]
    OutputRenderer: Type[OutputRenderer[OT]]
    OutputRendererDecorator: Type[OutputRendererDecorator[IT, OT]]

    def __init__(self, fn: OutputTransformerFn[IT, P, OT]) -> None:
        ...
```

That change makes the red underlines go away in shinylive.

### DIFF
--- a/shiny/_app.py
+++ b/shiny/_app.py
@@ -166,7 +166,7 @@ class App:
                 cast("Tag | TagList", ui), lib_prefix=self.lib_prefix
             )
 
-    def init_starlette_app(self):
+    def init_starlette_app(self) -> starlette.applications.Starlette:
         routes: list[starlette.routing.BaseRoute] = [
             starlette.routing.WebSocketRoute("/websocket/", self._on_connect_cb),
             starlette.routing.Route("/", self._on_root_request_cb, methods=["GET"]),
@@ -400,7 +400,7 @@ class App:
         return rendered
 
 
-def is_uifunc(x: Path | Tag | TagList | Callable[[Request], Tag | TagList]):
+def is_uifunc(x: Path | Tag | TagList | Callable[[Request], Tag | TagList]) -> bool:
     if (
         isinstance(x, Path)
         or isinstance(x, Tag)
--- a/shiny/_namespaces.py
+++ b/shiny/_namespaces.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import re
 from contextlib import contextmanager
 from contextvars import ContextVar, Token
-from typing import Pattern, Union, overload
+from typing import Generator, Pattern, Union, overload
 
 
 class ResolvedId(str):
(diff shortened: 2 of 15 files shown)


wch:
Update: there still seems to be red underlining when I build shinylive with this change:

image

I'm not sure if it's because shinylive is stuck on an old version of pyright, or if it's because it's not actually fixed here. Note that if I load up the same app in VS Code and use the shiny type stubs (instead of shiny source code), then it does _not_ underline the `@output`:

image

App source:

```py
from shiny import App, render, ui, Outputs, Session, Inputs

app_ui = ui.page_fluid(
    ui.output_text("txt"),
)

    @output
    @render.text()
    def txt():
        return "boo"

app = App(app_ui, server)
```

PR #833, opened 2023-11-29
chore(pyright): Fix typing error within accordion

schloerke:


schloerke:
Build tests pass. Merging early.

PR #842, opened 2023-12-05
Use development version of shiny for Connect deployments.

gshotwell:
The previous deployment tests were not actually testing the development version of pyshiny on Connect. The way Connect deployment works is:

1) Check if there's a requirements.txt file, if not, generate a new one from the active environment with `pip freeze`
2) Install packages from that file in a virtual environment on Connect

So when we had a requirments.txt file which specified `shiny` it would install shiny from pypi which is to say that our deployment tests were always testing one version back.

### DIFF
--- a/tests/deploys/apps/plotly_app/requirements.txt
+++ b/tests/deploys/apps/plotly_app/requirements.txt
@@ -2,4 +2,4 @@ pandas
 plotly
 git+https://github.com/posit-dev/py-shiny.git#egg=shiny
 git+https://github.com/posit-dev/py-htmltools#egg=htmltools
-shinywidgets
+git+https://github.com/posit-dev/py-shinywidgets#egg=shinywidgets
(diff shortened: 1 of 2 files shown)


tests/deploys/apps/plotly_app/requirements.txt
```diff
@@ -1,4 +1,5 @@
 pandas
 plotly
-shiny
+git+https://github.com/posit-dev/py-shiny.git#egg=shiny
+git+https://github.com/posit-dev/py-htmltools#egg=htmltools
 shinywidgets
```
cpsievert, line 5:
```suggestion
git+https://github.com/posit-dev/py-shinywidgets#egg=shinywidgets
```

karangattu (approved):
LGTM after you add Carson's suggested changes. Thanks

gshotwell:
@karangattu  I have no memory of this PR, but looks like a dropped the ball on merging it, has it been superceded?

PR #847, opened 2023-12-05
Allow express' `layout.sidebar` to handle inline layouts

### LINKED ISSUE #841: Shiny express, placing UI elements above the sidebar context manager causes error.
It appears that when you place a UI element above the sidebar layout context manager it causes the app to error. 

```
import numpy as np
from shiny import ui, render
from shiny.express import input, layout

np.random.seed(0)
data = np.random.randn(10000)

ui.h1("test") # causes the app to fail

with layout.sidebar():
    ui.input_slider("num_bins", "Number of Bins", min=1, max=50, value=30)

@render.plot
def plot():
    plt.hist(data, bins=input.num_bins())
```

```
    raise ValueError(
ValueError: `layout_sidebar()` is not being supplied with a `sidebar()` object. Please supply a `sidebar()` object to `layout_sidebar(sidebar)`.
```

### LINKED ISSUE #846: Consider making `page_fillable()` the default for `shiny.express`
This way, top-level fill items like plots, cards, or widgets, will fill the page how you might expect them to, for example:

```python
from shiny import render, ui
from shiny.express import input, layout

# This could go away
# layout.set_page(layout.page_fillable())

@render.plot
def hist():
    import seaborn as sns
    df = sns.load_dataset("penguins")
    fig, ax = plt.subplots()
    sns.histplot(df, x="bill_length_mm", hue="species", ax=ax)
    return fig
```

Another benefit of doing this is that we might not have to change the page layout for a top-level `with layout.sidebar():`

Screenshot 2023-12-05 at 3 35 26 PM

BTW, @wch, the last line in that screenshot should be `layout.set_page(layout.page_fixed())`

cpsievert:
Closes #841
Closes #846 

This PR does 3 things:

1. Changes express' default page to `page_fillable()` (see #846 for motivation)
2. Changes `express.layout.sidebar()` to be an interface to `layout_sidebar()`, not `sidebar()`.
3. In order for 2 to work sensibly, updates `layout_sidebar()` to accept a `TagChild` for it's first argument (`sidebar`) and wraps that value in a `sidebar()` if need be.

As a result of 2 and 3, we gain the following:

* Avoid problems like #841, where if `with layout.sidebar()` appears anywhere but the first top-level component, you'll get a error / undefined behavior.
* Somewhat relatedly, we no longer need to special case the page layout for `with layout.sidebar()` in a meaningful way
   * Using a `with layout.sidebar()` as a single, top-level, component still fills the page sensibly
* `with layout.sidebar()` now works sensibly as an inline sidebar layout, just like `ui.layout_sidebar()` does.

The main downside to this approach is that it may be surprising that code like this produces "Thing 1" in the sidebar, but "Thing 2"/"Thing 3" in the main region:

```python
from shiny import ui

with layout.sidebar():
    ui.div("Thing 1")
    ui.div("Thing 2")
    ui.div("Thing 3")
```

And to place multiple elements in the sidebar, you need an explicit `ui.sidebar()`

```python
from shiny import ui

with layout.sidebar():
    ui.sidebar(
        ui.div("Thing 1"), 
        ui.div("Thing 2")
    )
    ui.div("Thing 3")
```

That seems reasonable though, especially considering it seems as though we'll be needing this sort of pattern for other components, like cards:

```python
from shiny import ui, render

with layout.card():
    ui.card_header("Title")
    @render.plot
     def foo():
          ...
```

### DIFF
--- a/shiny/express/_run.py
+++ b/shiny/express/_run.py
@@ -19,7 +19,7 @@ from .display_decorator._node_transformers import (
 
 __all__ = ("wrap_express_app",)
 
-_DEFAULT_PAGE_FUNCTION = ui.page_fluid
+_DEFAULT_PAGE_FUNCTION = ui.page_fillable
 
 
 def wrap_express_app(file: Path) -> App:
--- a/shiny/express/layout.py
+++ b/shiny/express/layout.py
@@ -56,59 +56,42 @@ pre = wrap_recall_context_manager(htmltools.pre)
 # ======================================================================================
 def sidebar(
     *,
-    width: CssUnit = 250,
-    position: Literal["left", "right"] = "left",
-    open: Literal["desktop", "open", "closed", "always"] = "always",
-    id: Optional[str] = None,
-    title: TagChild | str = None,
+    fillable: bool = True,
+    fill: bool = True,
     bg: Optional[str] = None,
     fg: Optional[str] = None,
-    class_: Optional[str] = None,  # TODO-future; Consider using `**kwargs` instead
-    max_height_mobile: Optional[str | float] = "auto",
+    border: Optional[bool] = None,
+    border_radius: Optional[bool] = None,
+    border_color: Optional[str] = None,
     gap: Optional[CssUnit] = None,
     padding: Optional[CssUnit | list[CssUnit]] = None,
-) -> RecallContextManager[ui.Sidebar]:
+    height: Optional[CssUnit] = None,
+    **kwargs: TagAttrValue,
+) -> RecallContextManager[ui.CardItem]:
     """
-    Sidebar element
+    A sidebar layout.
 
-    Create a collapsing sidebar layout. This function wraps :func:`~shiny.ui.sidebar`.
+    Create a collapsing sidebar layout. This function wraps :func:`~shiny.ui.layout_sidebar`.
 
     Parameters
     ----------
-    width
-        A valid CSS unit used for the width of the sidebar.
-    position
-        Where the sidebar should appear relative to the main content.
-    open
-        The initial state of the sidebar. It can be `"desktop"` (the sidebar starts open
-        on desktop screen, closed on mobile), `"open"` or `True` (the sidebar starts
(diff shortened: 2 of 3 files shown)


schloerke:
I like us pushing back on the assumption that `shiny.express` do not need to be implemented 1:1 with `shiny.ui`.

> 2. Changes `express.layout.sidebar()` to be an interface to layout_sidebar(), not sidebar().

I disagree with this change, but I like it's intent of not requiring `layout.layout_sidebar()` for the top level.

I propose that `layout.page_fillable()` (or any layout method that can work with a sidebar, e.g. `card()` or `layout_sidebar()`) yank any top level `Sidebar` object to be implemented within `ui.layout_sidebar()` and all remaining top-level items be put into the main content. `layout.layout_sidebar()` would continue to exist as it does on `main` branch.

Ex:
```python
from shiny import ui

ui.div("Thing 0")

with layout.sidebar():
    ui.div("Thing 1")
    ui.div("Thing 2")

ui.div("Thing 3")

with layout.card():
  ui.card_header("header content")
  with layout.sidebar():
    ui.div("sidebar content")
  ui.div("card main content")
```
... would produce a `ui.page_sidebar()` whose sidebar contains `Thing 1` and `Thing 2`, and the main content has `Thing 0`, `Thing 3`, and the card (who has it's own sidebar of `sidebar content`) containing `header content` and `card main content`.

`layout.card()` could have `sidebar=MISSING` in `**kwargs` while still supporting it as a child. (Typing it out, I'd be happy even if we dropped any key-word sidebar parameter and only took it from an upgraded child.)

Assertions (for all methods who handle a sidebar) would need to be made that at most one sidebar object is supplied when upgrading the tags.


> 3. In order for 2 to work sensibly, updates layout_sidebar() to accept a TagChild for it's first argument (sidebar) and wraps that value in a `sidebar()` if need be.

I believe this situation should also throw a warning / message if no Sidebar object is found in it's children. (And then as you said, wrap the first child in a `sidebar()`.)

cpsievert:
Closing in favor of #847 (and #850).

PR #852, opened 2023-12-06
Pass express' sidebar title arg to the page title

cpsievert:
With this change, this code will yield the following:

```python

with layout.sidebar(title="My title"):
    "Sidebar"

"Main content"
```

Screenshot 2023-12-06 at 10 33 03 AM

Instead of 

Screenshot 2023-12-06 at 10 34 29 AM

(The latter design is intended more for "inline" sidebar layouts, which `with layout.sidebar()` doesn't support)


cpsievert:
Closing this in favor of another (incoming) PR

PR #853, opened 2023-12-06
Add `layout_sidebar()` to express to make inline sidebar layouts possible

cpsievert:
The current design of `shiny.express.layout.sidebar()` doesn't allow for inline sidebar layouts. This would make it a possibility with:

```python
from shiny import ui

with layout.layout_sidebar():
    ui.sidebar(
        ui.div("Thing 1"), 
        ui.div("Thing 2")
    )
    ui.div("Thing 3")
```

### TODO

Consider the direction proposed by @schloerke https://github.com/posit-dev/py-shiny/pull/847#issuecomment-1843104848 (which would remove the need to have a `layout_sidebar()`, and would seemingly also allow for inline layouts). I'm not sure we currently have a technical foundation to easily accommodate for this though

### DIFF
--- a/shiny/express/layout.py
+++ b/shiny/express/layout.py
@@ -18,6 +18,7 @@ __all__ = (
     "span",
     "pre",
     "sidebar",
+    "layout_sidebar",
     "layout_column_wrap",
     "column",
     "row",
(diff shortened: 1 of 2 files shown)


cpsievert:
Closing since `layout_sidebar()` was added somewhat recently

PR #861, opened 2023-12-06
Make sure the fill HTML dependency comes for the ride with `page_fillable()`

### LINKED ISSUE #860: Spacing between elements within `page_fillable()` is off when no fill items are present
For example, you get no spacing between these elements. The underlying cause is that `page_fillable()` currently doesn't bring in the underlying fill.css 

```python
from shiny import App, ui

app_ui = ui.page_fillable(
    ui.input_slider("slider", label="Slider", min=0, max=100, value=50),
    ui.p("Paragraph 1", class_="bg-dark"),
    ui.p("Paragraph 2", class_="bg-dark")
)

app = App(app_ui, None)
```

Screenshot 2023-12-06 at 4 58 21 PM

cpsievert:
Closes #860

### DIFF
--- a/shiny/ui/_page.py
+++ b/shiny/ui/_page.py
@@ -36,7 +36,7 @@ from ._sidebar import Sidebar, layout_sidebar
 from ._tag import consolidate_attrs
 from ._utils import get_window_title
 from .css import CssUnit, as_css_padding, as_css_unit
-from .fill._fill import FILLABLE_CONTAINTER_ATTRS
+from .fill._fill import as_fillable_container
 
 
 def page_sidebar(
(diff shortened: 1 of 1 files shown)


shiny/ui/_page.py
```diff
@@ -314,6 +309,15 @@ def page_fillable(
         lang=lang,
     )
 
+    # page returns a <html> tag, but we need to make the <body> fillable
+    body = page.children[1]
+    if not isinstance(body, Tag) or body.name != "body":
+        raise ValueError("Expected a <body> tag")
+
+    as_fillable_container(body)
```
wch, line 317: This function is being called for its side effect, modifying the object in place. I think in all other instances in the code base, we call it for its return value.

The function definition has a question of how it should work:
https://github.com/posit-dev/py-shiny/blob/ddfd7628608ee1bdc834110cd35bae5efb78c3f0/shiny/ui/fill/_fill.py#L33-L40

I think that for clarity and consistency, that function should only be used for its return value, and shouldn't alter the object in place. I'll make a follow-up PR that changes that function, and also alters the usage here to match.

wch (approved):

PR #868, opened 2023-12-07
Change express' `navset_tab()` -> `navset()` and `navset_card_tab()` -> `navset_card()` with `type = "underline"` default

### LINKED ISSUE #867: Prefer `navset_underline()` to `navset_tab()` in express
Currently we export `navset_card_tab()` and `navset_tab()`, but `navset_card_underline()` and `navset_underline()` are a better fit for the overall design of the shiny preset

cpsievert:
Closes #867

### Note

This does make a breaking change of swapping out:

* `shiny.express.layout.navset_tab()` for `shiny.express.layout.navset()`
* `shiny.express.layout.navset_card_tab()` for `shiny.express.layout.navset_card()`

It also makes `type = "underline"` the default, but tabs can still be achieved via `with navset(type = "tab")`


wch:
Just curious: why do we want to get rid of `navset_tab` and `navset_card_tab`?

cpsievert:
The underline variants are a much better match design wise for the shiny preset stuff (i.e., the new bootstrap.css/shiny.css defaults) we've been doing

cpsievert:
@wch here's a PR to reflect the change in the your gh-pages doc https://github.com/wch/shiny_express_doc/pull/2

wch:
Hang on, I don't think we should just go and make a different API for express and classic shiny here. We need to discuss these changes.

PR #885, opened 2023-12-09
Print a warning message when running Express app

wch:
Because the Express API is still evolving, we want to make it clear to users that the API could change in a way that breaks their app. The idea is that this warning would be in the next release of Shiny, but _not_ in the release after that.

The warning looks something like this:

```
WARNING:root:Detected Shiny Express app. please note that Shiny Express is still in development and the API is subject to change!
INFO:     Started server process [8187]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:52623 (Press CTRL+C to quit)
```

I looked into making it so that it only warns once when you call `shiny run --reload`, but that ended up being pretty complicated to implement, because every time the app reloads, it starts a new Python process, so the state would have to be kept somewhere else, like on disk. I think it's OK to not do that since this is only a temporary thing for people who are using experimental features.

### DIFF
--- a/shiny/express/_run.py
+++ b/shiny/express/_run.py
@@ -36,8 +36,7 @@ def wrap_express_app(file: Path) -> App:
     :
         A `shiny.App` object.
     """
-
-    logging.warning(
+    logging.getLogger("uvicorn.error").warning(
         "Detected Shiny Express app. please note that Shiny Express is still in "
         "development and the API is subject to change!"
     )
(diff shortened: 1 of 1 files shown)


jcheng5:
You might be able to get rid of the `root:` by doing `logging.getLogger("uvicorn.error").warning(...)`

jcheng5 (approved):

wch:
That worked great. With that change it looks like:

```
WARNING:  Detected Shiny Express app. please note that Shiny Express is still in development and the API is subject to change!
INFO:     Started server process [10489]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
```

PR #887, opened 2023-12-09
Fix failing express tests

karangattu:
1. Fix the `pyright` errors since recent code changes in Shiny express from `navset_tab()` ->` navset()` and `navset_card_tab()` -> `navset_card()`
2. install main version of `rsconnect-python` instead of pypi since pypi version does not yet support shiny express tests and we want to test shiny express tests.
3. Modify name of the test since it was copied from another test and not changed to reflect the test file name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,13 +94,16 @@ install-playwright:
 install-trcli:
 	which trcli || pip install trcli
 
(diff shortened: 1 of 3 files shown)


wch:
Thank you!

Issue #866, opened 2023-12-07
Add a helper for setting page arguments


cpsievert:
As highlighted in #852, this unfortunate code is required to get a page title with the sidebar layout.

```python
from shiny.express import layout

layout.set_page(layout.page_sidebar(title="Page title"))
```

And we have a similar problem if you want to change page-level settings like `gap`, `padding`, `lang`, etc.

I think it could make sense to have both a `.set_title()` and a lower-level `.set_page_parameters()`

wch:
Some options and comments:

1. `layout.set_page_sidebar(title="Page title")`: One would never actually use `with layout.page_sidebar(title="Page title")`, so it might make sense to get rid of that function and replace it with `layout.set_page_sidebar(title="Page title")` (and same for the other page functions). Drawbacks: the user still needs to know what kind of page function is used by default.
2. `layout.set_title("Page title")`: This is nice and simple, but doesn't really make sense to extend to all the arguments for `page_xx` functions.
3. `layout.set_page_parameters(title="Page title")`: This is more general than `.set_title()`, but we can't offer the user much support for knowing what parameters are available. So in the end, other than very common parameters like `title`,  the user still would need to know which `page_xx` function they are using, and look up the available parameters there.

To me, there's not an obvious winner here.

cpsievert:
We now have `shiny.express.ui.page_opts()` which gets at the main idea here

PR #893, opened 2023-12-11
Smarter, lazier, and more complete page default/api for express

cpsievert:
This PR currently just introduces a `TopLevelRecallContextManager()`, which inspects it's args to determine a smarter page (i.e., `page_sidebar()`/`page_navbar()`/something else) based on the presence of top-level `layout.sidebar()`/`layout.nav()`/other components.
  * As result, this PR makes multi-page apps (with/without a sidebar) possible. It also closes https://github.com/posit-dev/py-shiny/issues/841.

This new approach isn't compatible with the more immediate/imperative `layout.set_page(layout.page_*())` API, so I'm removing it for now. This does bring up, however, more questions as to how to the page API (https://github.com/posit-dev/py-shiny/issues/866) should look. Originally I was thinking `.set_page_*()` + `.set_title()` would be sensible, but on second thought, a `set_page_preferences()` is probably more ideal. There you could not only control page `kwargs` (e.g., `title`, `lang`, etc), but also parameters related to default page container decision (e.g., `fillable`, `wider` (i.e., `page_fluid()` vs `page_fixed()`), etc))

Also note that, we still currently have the issue of `layout.sidebar()` in a non-top-level context.  I think, ideally, a non-top-level `RecallContextManager` could handle this by doing essentially the same as `TopLevelRecallContextManager` (i.e., `wrapping` any `Sidebar`(s) into a `layout_sidebar()`)


wch:
Things the user needs to control:

- Title
- Fillable
- Width
- (Maybe) If inference is wrong, explicit way of setting page function


Possible interface for this:

- `set_page_xxx`
- `set_title`, `set_fillable`, `set_width`
- `set_page_preferences(*, title, fillable, width, **kwargs)`

wch:
I've added the following functions:

```
set_page_title()
set_page_lang()

# The two below affect heuristics for which page function is used.
set_page_fillable()
set_page_wider()
```

If the user wants to explicitly choose which page function is used, there are the following:

```
use_page_fixed()
use_page_fluid()
use_page_fillable()
use_page_sidebar()
use_page_navbar()
```


All of these functions live in `express._page`, but I haven't imported them into `express.layout`, because that will introduce merge conflicts with #904. I think we should merge these two PRs and then export the functions here from `express.ui`.

shiny/express/_page.py
```diff
@@ -0,0 +1,334 @@
...
+    nNavs = len(navs)
+    nSidebars = len(sidebars)
+
+    if _page_fn is None:
+        if nNavs == 0:
+            if nSidebars == 0:
+                if _fillable:
+                    _page_fn = (
+                        ui.page_fillable
+                    )  # pyright: ignore[reportGeneralTypeIssues]
```
cpsievert, line 61: Not that we absolutely need this now, but I think we could support `wider=F` when `fillable=T` by including  a `{"class": "container"}`?

shiny/express/_page.py
```diff
@@ -0,0 +1,334 @@
...
+from ._recall_context import RecallContextManager
+from ._run import get_top_level_recall_context_manager
+
+__all__ = (
+    "page_auto",
+    "page_auto_cm",
+    "set_page_title",
+    "set_page_lang",
+    "set_page_fillable",
+    "set_page_wider",
```
cpsievert, line 21: It still seems to me that we should have a singular `set_page_preferences()` function, rather than one function per page parameter. I can see two benefits of doing so:

* API surface doesn't need to expand for other parameters we want to bless (i.e., should we be including `set_page_gap`/`set_page_padding` for a `page_fillable()` context?)
* With this current API, it's not very clear that you're really setting _preferences_ (at least for `fillable`/`wider`) that may or may not effect the end result

shiny/express/_page.py
```diff
@@ -0,0 +1,334 @@
...
+)
+
+
+def page_auto_cm() -> RecallContextManager[Tag]:
+    return RecallContextManager(
+        page_auto,
+        kwargs={
+            "_page_fn": None,
+            "_fillable": False,
+            "_wider": False,
```
cpsievert, line 36: I don't feel strongly, but seems like slightly better naming?

```suggestion
            "_restrict_width": True,
```

shiny/express/_page.py
wch: I agree that a function like `set_page_preferences()` could make sense for things like `fillable` and `wider`, but the title feels different to me. It's not a preference like the others, and it's such a common thing to do that this feels really verbose:

```py
layout.set_page_preferences(title = "My Title")
```

wch: Maybe:

```
layout.page_opts(
  title = "My title",
  lang = "en",
  filllable = True,
  full_width = True,
)
```

wch: This could also work in Shiny Core:

```py
ui.page(
  title = "My title",
  lang = "en",
  filllable = True,
  full_width = True,
)
```

wch:
From a conversation with @cpsievert, @gadenbuie, @jcheng5:

Things we can do now:
- Add `page_opts()` function. Make all args keyword only (first arg is `*`).
- Make `page_auto()` the default page. Also add that function to core UI.
- Add `set_page_fn()`

    ```py
    layout.set_page_fn(
        lambda *args: ui.page_navbar(*args, ...)
    )
    
    # or
    
    def my_page_fn(*args):
        return ui.page_navbar(*args, ...)
    
    layout.set_page_fn(my_page_fn)
    ```

Things we can do possibly in the future:
- Maybe add `page_opts(mode = "navbar", ...)` or `page_opts_navbar(...)`
- Allow passing `*kwargs` to `page_opts()`? How to distinguish between HTML attributes and proper args?

wch:
I've retargeted this PR against `express-layout-to-ui` (#904) to make the changes make more sense.

I implemented the changes as described in the previous comment:

* Add `page_opts()` function. Make all args keyword only (first arg is `*`).

* Make `page_auto()` the default page. Also add that function to core UI.

* However, for the third item, I made a change. Instead of adding a function `set_page_fn()`, I added a `page_fn` parameter to `page_opts()`. It would be used like this:

    ```py
    page_opts(page_fn=shiny.ui.page_fillable)
    ```

    I did this because the parameters to `page_opts()` basically map to `page_auto()` parameters, and `page_auto()` already accepts `_page_fn`.

shiny/express/_page.py
jcheng5: This ended up being `_full_width`

shiny/ui/_page.py
```diff
@@ -443,6 +444,99 @@ def page_bootstrap(
...
+    Parameters
+    ----------
+    *args
+        UI elements. These are used to determine which page function to use, and they
+        are also passed along to that page function.
+    _page_fn
+        The page function to use. If ``None`` (the default), will automatically choose
+        one based on the arguments provided.
+    _fillable
+        This has an effect only if there are no sidebars or top-level navs. If ``True``,
```
jcheng5, line 472: `page_navbar` and `page_sidebar` both take a `fillable` argument, we could pass it through

shiny/ui/_page.py
```diff
@@ -443,6 +444,99 @@ def page_bootstrap(
     )
 
 
+def page_auto(
+    *args: TagChild | TagAttrs,
+    _page_fn: Callable[..., Tag] | None = None,
+    _fillable: bool = False,
+    _full_width: bool = False,
+    **kwargs: object,
```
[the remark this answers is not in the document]
jcheng5: Or don't take `_page_fn`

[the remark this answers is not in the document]
wch: I considered taking out `page_fn`, but I think we should keep it because it's useful when, in Express, you call `page_opts(page_fn=...)`. The `page_opts` function can just pass it through to this function.

Issue #854, opened 2023-12-06
Consider renaming `shiny.express.layout`


cpsievert:
The name `layout` seems to be over-optimizing for `with layout.sidebar()`. It gets awkward with things like `layout.layout_*()`, `layout.page_*()`, etc.

I think I like `ux` as an alternative. Here are some pros/cons I can think of:

#### Pros

* Short and similar to `ui`, which helps with the mental model of "use `ux` as an alternative of `ui` when you want to use `@render.*` decorators as children"
* Serves as a more general home for UI things that might not be `layout` related.

### Cons

* It's maybe too similar to `ui` (hurts readability), but the fact that `ux` things will always start `with ux` will provide a clear enough distinction.

gadenbuie:
I also like `ux`. Where UI refers to the visual elements and design of a website (the user interface), UX encompasses the overall experience of users while using a website (the user experience). These terms are also aligned with the difference between `shiny.ui` and `shiny.express.ux`: the Shiny Express version of UI isn't just an interface but can also include outputs. It's directionally aligned with the UX terminology and hints that Shiny Express elements can also wrap the outputs.

I also see a lot of value in being able to mix Express and core UI elements, so I see the closeness as useful to be able to switch between implementations if you realize you need one or the other.

```py
from shiny import ui
from shiny.express import layout as ux

with ux.div():
  "something in a div"

ui.div("something also in a div")
```

gshotwell:
My sense of what's important for express is the following:
1) Make Express succinct and easy to use
2) Make it easy to move from Express to Classic
3) Clearly differentiate Express and Classic

It seems that mixing Classic containers and Express containers is a bit of an edge case in that world, and so we shouldn't optimize for that use if it undermines the other goals. 

I really don't like the ui/ux because it's so easy to mess up when copying and pasting, and it'll make our documentation very confusing. I would have trouble distinguishing these two blocks of code, so a new learner would probably have more trouble. 

```
from shiny import ui
from shiny.express import layout as ux

with ui.div():
  "something in a div

ux.div("something also in a div")
```

```
from shiny import ui
from shiny.express import layout as ux

with ux.div():
  "something in a div"

ui.div("something also in a div")
```

[...]

PR #904, opened 2023-12-13
Change `express.layout` to `express.ui`

wch:
This PR changes the name of `express.layout` to `express.ui`, as described in https://github.com/posit-dev/py-shiny/issues/854#issuecomment-1848011932.

It also adds a test to make sure that all items from `shiny.ui` are accounted for in `shiny.express.ui`, and vice versa. For both of these modules, either a corresponding item must exist in both modules, or it must be listed explicitly in a variable `_known_missing`.

There are a few remaining API questions:
- For some functions, like `card_header()`, it's not clear to me whether they should be context managers or regular functions.
- `shiny.ui.page_navbar()` takes a named argument `sidebar`, which comes after the `*args`. The `express` version of this function also takes a named argument `sidebar`, but the only way to use it is with something like this, which is really awkward:

    ```py
    s = shiny.ui.sidebar(....)
    with shiny.express.ui.page_navbar(sidebar=s):
        ....
    ```
    
    It would be much better if we could pass in a sidebar object just as a `*arg`, and it would discover it and place it correctly.


cpsievert:
@wch fwiw #893 solves that sidebar issue for `page_navbar()`, meaning you can do something like this and it "just works" (i.e., it infers a `page_navbar()` context is relevant, and puts the sidebar in the keyword arg)

```python
with ui.sidebar():
   ...

with ui.nav("Page 1"):
    "Page 1 content"

with ui.nav("Page 1"):
    "Page 2 content"
```

We still have that (sidebar kwarg) problem for `navset_card()` though.

shiny/express/ui/__init__.py
```diff
@@ -0,0 +1,300 @@
...
+    "page_sidebar",
+    "page_navbar",
+)
+
+
+_known_missing = {
+    # Items from shiny.ui that don't have a counterpart in shiny.express.ui
+    "shiny.ui": (
+        "nav_control",
+        "nav_menu",
```
cpsievert, line 269: It seems fairly obvious to me that these should be ctx managers

shiny/express/ui/__init__.py
```diff
@@ -0,0 +1,300 @@
...
+        "nav_spacer",
```
cpsievert, line 270: This should never have any children, so I think it's probably a regular function?

shiny/express/ui/__init__.py
```diff
@@ -0,0 +1,300 @@
...
+        "navset_card_pill",
+        "navset_card_tab",
+        "navset_card_underline",
+        "navset_hidden",
+        "navset_pill",
+        "navset_pill_card",
+        "navset_pill_list",
+        "navset_tab",
+        "navset_tab_card",
+        "navset_underline",
```
cpsievert, line 281: Are we not including these because express has `navset()` and `navset_card()` now? 

For context, when I added `navset()`/`navset_card()`, the primary motivation was coming trying to keep `shiny.express.ui`'s API minimal (I still think that could be a valuable thing, but would come at the cost of needing two UI imports).

Anyway, assuming we go in this direction (of `shiny.express.ui` basically having everything that `shiny.ui` has), I think I'd be favor of removing `navset()`/`navset_card()` and exposing all these as context managers.

shiny/express/ui/__init__.py
```diff
@@ -0,0 +1,300 @@
...
+        "page_bootstrap",
+        "panel_main",
+        "panel_sidebar",
+        "panel_title",
+        "page_output",
```
cpsievert, line 286: This should come after `page_bootstrap()`?

shiny/express/ui/__init__.py
wch: I don't have strong opinions about this either way. I'm fine with removing `navset()` and `navset_card()` and adding the others as context managers.

wch: I replaced `navset` and `navset_card` with the specific components in 364cbad6.

shiny/express/_recall_context.py
```diff
@@ -54,6 +54,19 @@ def __exit__(
...
+        if callable(getattr(res, "tagify", None)):
+            return res.tagify()  # pyright: ignore
+        if callable(getattr(res, "_repr_html_", None)):
+            return res._repr_html_()  # pyright: ignore
+
+        raise RuntimeError(
+            "RecallContextManager was used without `with`. When used this way, the "
+            "result must have a .tagify() or ._repr_html_() method, but it does not."
+        )
+
```
wch, line 69: Note this `tagify` method on `RecallContextManager`.

shiny/express/layout.py
```diff
@@ -1,904 +1,14 @@
...
-    "nav_panel",
-    "page_fluid",
-    "page_fixed",
-    "page_fillable",
-    "page_sidebar",
+warnings.warn(
+    "shiny.express.layout has been deprecated and renamed to shiny.express.ui. "
+    "Please import shiny.express.ui instead of shiny.express.layout.",
+    ImportWarning,
+    stacklevel=2,
```
wch, line 9: `express.layout` now raises this warning if it is imported.

jcheng5:
I would lean context manager for `card_header`, if I had to choose one. Would it be crazy to support both, like the tags do? I know we can't do this for containers that have nontrivial logic relating to their children.

shiny/express/_recall_context.py
jcheng5: Is there a reason this is stricter than `__exit__`? Oh, is it because sys.displayhook will do something similar?

jcheng5: Never mind, I understand

shiny/express/_recall_context.py
```diff
@@ -54,6 +54,19 @@ def __exit__(
             sys.displayhook(res)
         return False
 
+    def tagify(self) -> Tag | TagList | MetadataNode | str:
+        res = self.fn(*self.args, **self.kwargs)
+
+        if callable(getattr(res, "tagify", None)):
+            return res.tagify()  # pyright: ignore
+        if callable(getattr(res, "_repr_html_", None)):
+            return res._repr_html_()  # pyright: ignore
```
jcheng5, line 63: It would be nice if someday we could ask htmltools if an object is outputtable

PR #905, opened 2023-12-13
Express: don't automatically put strings in a `<pre>` block

### LINKED ISSUE #872: Express string output are quoted and wrapped in a code block
I find this behavior surprising (both the quoting and wrapping of strings into a code block):

```python
import shiny.express

"Hello"
"express"
```

Screenshot 2023-12-07 at 1 22 24 PM

I would have expected a result closer to:

```python
from shiny import ui
import shiny.express

ui.div("Hello")
ui.div("express")
```

Screenshot 2023-12-07 at 1 24 55 PM

Especially considering it's not terribly hard to opt-into code blocks (and I think we've been considering adding a more official one)?

````python
from shiny import ui
import shiny.express

ui.markdown(
    """
```
Hello
```

```
express
```
"""
)
````

But maybe there's additional rationale for this as a default that I'm not aware of?

wch:
Closes #872.

This PR makes it so strings (and other types of objects) are not automatically wrapped in a `<pre>` block, as described in https://github.com/posit-dev/py-shiny/issues/872#issuecomment-1852630725.

image

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,7 @@ install_requires =
     starlette>=0.17.1
     websockets>=10.0
     python-multipart
-    htmltools>=0.4.1.9001
+    htmltools @ git+https://github.com/posit-dev/py-htmltools.git
     click>=8.1.4
     markdown-it-py>=1.1.0
     # This is needed for markdown-it-py. Without it, when loading shiny/ui/_markdown.py,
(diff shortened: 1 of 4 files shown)


jcheng5:
LGTM except that tests are failing?

wch:
One notable change in behavior is that values are just passed directly to context manager UI component. Here's an example where the values are passed to the `page_fixed` function that encompasses the content of the .py file.


image

PR #908, opened 2023-12-13
Add output to React template

gshotwell:
When testing the react template I found I needed to do some research to figure out how to add an output. This adds a react output as well as an input. The idea here is that it's easier to delete the boilerplate than to add it.


nstrayer (approved):
Everything looks good to me! 
I feel like eventually we may want to make the templates more dynamic where we dynamically add input or output code to the template, use typescript or not, etc. based on questions but this is good for now!

PR #913, opened 2023-12-15
Make output transformer execute using async/await syntax

schloerke:
Simplifies transformer experience and removes [major] caveat of being able to use an non-async (but created as async) implementation.

Questions:
- [ ] Name for TranformerMetadata key for the value_fn being originally sync/async. Different mechanism?


schloerke:
Included within #964. Closing

PR #920, opened 2023-12-18
feat: Allow for `App` `server=` to take `input` only

schloerke:
Related: #793


shiny/_app.py
```diff
@@ -446,3 +452,29 @@ def file_response_handler(req: Request) -> FileResponse:
...
+        # Only has 1 parameter, ignore output, session
+        server(input)
+
+    return _server
+
+
+def has_only_input(
+    x: Callable[[Inputs], None] | Callable[[Inputs, Outputs, Session], None]
+) -> TypeGuard[Callable[[Inputs], None]]:
+    return len(signature(x).parameters) == 1
```
wch, line 474: I understand that the `TypeGuard` does provide some help for the type checker here, but still, I think it's clearer if the `len(...) == 1` check is inlined in the `elif` statements above, and then a `cast` is used inside the `elif` blocks. As it is, I find myself jumping up and down in the code to understand what it's doing.

It would be nice if Python type checking were smarter and didn't require either of these workarounds.

shiny/_app.py
```diff
@@ -446,3 +452,29 @@ def file_response_handler(req: Request) -> FileResponse:
             file_response_handler,
             name="shiny-app-static-assets-" + mount_point,
         )
+
+
+def noop_server(input: Inputs, output: Outputs, session: Session) -> None:
+    pass
+
+
+def ignore_output_session_server(
```
wch, line 461:
```suggestion
def wrap_server_fn_with_output_session(
```

shiny/_app.py
```diff
@@ -446,3 +452,29 @@ def file_response_handler(req: Request) -> FileResponse:
             file_response_handler,
             name="shiny-app-static-assets-" + mount_point,
         )
+
+
+def noop_server(input: Inputs, output: Outputs, session: Session) -> None:
```
wch, line 457: For clarity:

```suggestion
def noop_server_fn(input: Inputs, output: Outputs, session: Session) -> None:
```

PR #927, opened 2023-12-19
Minor improvements to palmerpenguins dashboard

cpsievert:
Makes small improvements the UI (filling layout + titles + images) and makes the server code a bit more readable.

### DIFF
--- a/shiny/templates/app-templates/dashboard/app-core.py
+++ b/shiny/templates/app-templates/dashboard/app-core.py
@@ -5,14 +5,25 @@ import seaborn as sns
 
 from shiny import App, Inputs, Outputs, Session, reactive, render, ui
 
-sns.set_theme(style="white")
-df = pd.read_csv(Path(__file__).parent / "penguins.csv", na_values="NA")
 species = ["Adelie", "Gentoo", "Chinstrap"]
 
+species_images = {
+    "Adelie": "https://upload.wikimedia.org/wikipedia/commons/thumb/e/e3/Hope_Bay-2016-Trinity_Peninsula%E2%80%93Ad%C3%A9lie_penguin_%28Pygoscelis_adeliae%29_04.jpg/1280px-Hope_Bay-2016-Trinity_Peninsula%E2%80%93Ad%C3%A9lie_penguin_%28Pygoscelis_adeliae%29_04.jpg",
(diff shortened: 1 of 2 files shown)


shiny/templates/app-templates/dashboard/app-core.py
```diff
@@ -5,94 +5,101 @@
 
 from shiny import App, Inputs, Outputs, Session, reactive, render, ui
 
-sns.set_theme(style="white")
-df = pd.read_csv(Path(__file__).parent / "penguins.csv", na_values="NA")
 species = ["Adelie", "Gentoo", "Chinstrap"]
 
+species_images = {
```
gshotwell, line 10: I kind of wanted to make the dashboard pretty minimal, so I'm not sure about adding images. My experience is that images in value boxes is a bit rare. Also we probably don't want to link out to web sites that might change.

gshotwell (changes requested):
This mostly seems good to me, I don't love including images here but if you feel strongly that they should be there I'm happy to include them. Maybe we should add a template with value box images on the github template site instead of including them in the package.

shiny/templates/app-templates/dashboard/app-core.py
jcheng5: I agree, especially for a template.

shiny/templates/app-templates/dashboard/app-core.py
```diff
@@ -5,94 +5,101 @@
...
+    "Adelie": "https://upload.wikimedia.org/wikipedia/commons/thumb/e/e3/Hope_Bay-2016-Trinity_Peninsula%E2%80%93Ad%C3%A9lie_penguin_%28Pygoscelis_adeliae%29_04.jpg/1280px-Hope_Bay-2016-Trinity_Peninsula%E2%80%93Ad%C3%A9lie_penguin_%28Pygoscelis_adeliae%29_04.jpg",
+    "Gentoo": "https://upload.wikimedia.org/wikipedia/commons/thumb/0/00/Brown_Bluff-2016-Tabarin_Peninsula%E2%80%93Gentoo_penguin_%28Pygoscelis_papua%29_03.jpg/1202px-Brown_Bluff-2016-Tabarin_Peninsula%E2%80%93Gentoo_penguin_%28Pygoscelis_papua%29_03.jpg",
+    "Chinstrap": "https://upload.wikimedia.org/wikipedia/commons/0/08/South_Shetland-2016-Deception_Island%E2%80%93Chinstrap_penguin_%28Pygoscelis_antarctica%29_04.jpg",
+}
 
-def make_value_box(penguin):
+
+def penguin_box(x):
```
jcheng5, line 17: Can you rename this back to `make_value_box`? Just to minimize the work the user will have to do when they adapt this to their own use case.

cpsievert:
Closing in favor of #1273

PR #936, opened 2023-12-22
Update `output_transformer` to work with docs. Add `output_transformer_simple`

schloerke:
WIP, but very exciting that pylance supports passing through documentation if the decorator returns a single `Callable` type (not a callable class, union of `Callable`, or overloaded function).

Therefore, if we can return a single Callable, the docs can be transferred through and the pkg author does not need to type out overloads manually.


schloerke:
Closing in favor of #964

PR #949, opened 2023-12-28
Include `tooltip()` and `popover()` in express

### LINKED ISSUE #946: `shiny.express.ui` is missing `tooltip` and `popover`
Which is especially weird given it has `ui.update_*()`

cpsievert:
Closes #946

### DIFF
--- a/shiny/express/ui/__init__.py
+++ b/shiny/express/ui/__init__.py
@@ -94,6 +94,8 @@ from ...ui import (
     modal_remove,
     notification_show,
     notification_remove,
+    tooltip,
+    popover,
     nav_spacer,
     output_plot,
     output_image,
@@ -269,6 +271,8 @@ __all__ = (
     "panel_conditional",
     "panel_fixed",
     "panel_absolute",
+    "popover",
+    "tooltip",
     # Imports from ._page
     "page_opts",
 )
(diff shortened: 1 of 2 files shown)


wch:
I wasn't sure what API we want for these functions. With this PR, these functions need to be used in the non-context-manager way:

```py
from shiny.express import ui

ui.tooltip(
    ui.input_action_button("btn", "A button"),
    "A message",
    id="btn_tooltip",
)
```

This is quite different from the other context manager stuff in Express. Also, if the user wants to have nested content for either the trigger or tooltip/popover, they will need to have a separate import of `shiny.ui` and use those functions.


Here are a few possible ways of doing this with context managers. Note that in all cases, the kwargs (like `id`) needs to be passed in immediately.

```py
# (1) Passing values straight through
with ui.tooltip(id="btn_tooltip"):
    ui.input_action_button("btn", "A button")
    "A message"
    

# (2) Taking the `trigger` argument:
with ui.tooltip(ui.input_action_button("btn", "A button"), id="btn_tooltip"):
    "A message"


# (3) Taking the *args:
with ui.tooltip("A message", id="btn_tooltip"):
    ui.input_action_button("btn", "A button")
```

I have to say that I don't really love any of these, but I think (1) is the least weird.

With (1), it's weird that the kwargs have to go to the into the function call immediately, but then the unnamed args come later, after the `trigger`.

With (2), it feels strange to me that the `trigger` are passed in alongside the kwargs, and then the unnamed args come later. The reason it feels weird is because the kwargs are conceptually more closely associated with the unnamed args, than with the `trigger`. It is also limiting if you want a large `trigger`.

(3) feels better to me than (2), but is limiting in a similar way, if there's a lot of content in the tooltip. This is even more true for popovers. 

For both (2) and (3), it is not possible to have nested content in the args passed to the function, unless a separate import is made for `shiny.ui`.

jcheng5:
Easily option 1 for me, personally.

cpsievert:
Agreed, I'll update the PR accordingly

wch (approved):

PR #950, opened 2023-12-28
Consistently pass positional arguments from express context managers to their functional component signature

### LINKED ISSUE #947: Issues with positional arguments and `shiny.express.ui` components
### Problem

I can see two issues with how `shiny.express.ui` handles positional arguments:

1. Many functions don't pass along positional arguments (i.e., they have `*` instead of `*args` in their signature). This means you can't currently use things like `ui.card()` the "functional" way:

```python
from shiny.express import ui

ui.card("foo", "bar")
```

```python
TypeError: card() takes 0 positional arguments but 2 were given
```

2. Some functions have named positional arguments, which forces those arguments to be specified when used the "context-manager" way. For example, since `value_box()` has `title`/`value` as named positional args, usage _must_ look like this:

```python

with value_box("Title", "Value"):
    pass
```

...which is unfortunate, especially when you want to a dynamically rendered value. 

### Proposal

I think we can solve both issues by taking inspiration from how `card_header()` currently works. That is, if the underlying UI function takes any positional arguments, we _always_ start the express function with `*args` and pass those args to the underlying UI function

https://github.com/posit-dev/py-shiny/blob/b61452d565edf0e8895f5fcf80874dac3d1a689a/shiny/express/ui/_cm_components.py#L455-L487

This does lead to the question of whether functions like `nav_panel()` should be `(title, *args, **kwargs)` or `(*args, **kwargs)`, but it seems pretty sensible to me (for consistency sake) that we go in the latter direction.

cpsievert:
Closes #947

This PR adds positional arguments to all `shiny.express.ui` context managers that wrap a `shiny.ui` function with positional arguments. This primarily means that valid `shiny.ui` code also works for `shiny.express.ui`. For example, you can now do:

```python
from shiny.express import ui

ui.card("Hello", "card")
```

instead of being forced into:

```python
from shiny.express import ui

with ui.card():
  "Hello"
  "card"
```

I can see at least a handful of nice things that come out of this:

1. We have the option to write more documentation in a express-mode agnostic way.
2. Users have the option to write more UI code that is express-mode agnostic.
3. Less confusion with users wondering why something like `ui.input_action_button("foo", "bar")` works, but `ui.card("foo", "bar")` doesn't.
4. More consistency with express' context managers handling of positional arguments.
    * Currently `card_header()` allows positional arguments because it's not clear whether you want to use it as a context manager. I think you could make a similar argument for a lot of other UI components.
5. Mixing positional arguments and children seems desirable in some situations like `ui.value_box()`, for example:

```python
from shiny import render
from shiny.express import ui

with ui.value_box("Title 3"):
    @render.ui
    def value_3():
        return "Value 3"
```

## TODO

- [ ] Double-check types on `*args`
- [ ] Document/updates for breaking change to `nav_panel()`, `accordion_panel()`, and `nav_menu()`
- [ ] Unit tests

### DIFF
--- a/shiny/express/ui/__init__.py
+++ b/shiny/express/ui/__init__.py
@@ -94,6 +94,8 @@ from ...ui import (
     modal_remove,
     notification_show,
     notification_remove,
+    tooltip,
+    popover,
     nav_spacer,
     output_plot,
     output_image,
--- a/shiny/express/ui/__init__.py
+++ b/shiny/express/ui/__init__.py
@@ -94,8 +94,6 @@ from ...ui import (
     modal_remove,
     notification_show,
     notification_remove,
-    tooltip,
-    popover,
     nav_spacer,
     output_plot,
     output_image,
--- a/shiny/express/ui/_cm_components.py
+++ b/shiny/express/ui/_cm_components.py
@@ -39,7 +39,7 @@ __all__ = (
 # Shiny layout components
 # ======================================================================================
 def sidebar(
-    *,
+    *args: TagChild | TagAttrs,
     width: CssUnit = 250,
     position: Literal["left", "right"] = "left",
     open: Literal["desktop", "open", "closed", "always"] = "always",
@@ -59,6 +59,8 @@ def sidebar(
 
     Parameters
     ----------
+    *args
+        Positional arguments for :func:`~shiny.ui.sidebar`.
     width
         A valid CSS unit used for the width of the sidebar.
     position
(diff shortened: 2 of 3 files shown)


cpsievert:
Closing since, after discussing with Joe and Winston, I was reminded that there are several issues with this approach:

* Handling errors gets difficult/tricky
* We can't pass both positional args and children to the context manager, which is important for things `navset_*()` that do transformation of those values.

PR #967, opened 2024-01-04
Add `render.download`

wch:
This PR adds `render.download`. This is a proof-of-concept; the final implementation will have to be done after #964 is merged, and will be significantly different. But the app-developer-facing API will be mostly the same.

The existing API is to use `session.download`, like this:

```py
# UI contains:
ui.download_button("dl", "Download File")

# Server function contains:
@session.download(filename="hello.txt")
def dl():
    yield "Hello, world!\n"
```

This is quite a bit different from all other outputs, because it relies on the `session` object. This also doesn't work for Express apps, because there isn't a `session` object available. (I'll note here that it would be possible to add a `session` object that could be imported from `shiny.express`, but the change in this PR is a better one.)

With this PR, the app uses `render.download` instead of `session.download`:

```py
# UI contains:
ui.download_button("dl", "Download File")

# Server function contains:
@render.download(filename="hello.txt")
def dl():
    yield "Hello, world!\n"
```

I think this makes more sense from the app author's perspective.

In an Express app, the code would look like this (the button is automatically inserted in the UI):

```py
@render.download(filename="hello.txt")
def dl():
    yield "Hello, world!\n"
```

However, there is currently one limitation with this: it would be nice to be able to pass in a label for the button, but it doesn't work, because of where the function is defined for the button. @schloerke With the changes in #964, would it be possible to pass the label to the `ui.download_button` function?

### DIFF
--- a/shiny/render/_render.py
+++ b/shiny/render/_render.py
@@ -6,15 +6,18 @@ __all__ = (
     "image",
     "table",
     "ui",
+    "download",
 )
 
 import base64
 import os
 import sys
 import typing
+import urllib.parse
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     Literal,
     Optional,
     Protocol,
@@ -24,7 +27,7 @@ from typing import (
     runtime_checkable,
 )
 
-from htmltools import TagChild
+from htmltools import Tag, TagChild
 
 if TYPE_CHECKING:
     from ..session._utils import RenderedDeps
@@ -33,6 +36,7 @@ if TYPE_CHECKING:
 from .. import _utils
 from .. import ui as _ui
 from .._namespaces import ResolvedId
+from ..session._session import DownloadHandler, DownloadInfo
 from ..types import MISSING, MISSING_TYPE, ImgData
 from ._try_render_plot import (
     PlotSizeInfo,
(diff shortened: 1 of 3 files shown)


shiny/render/_render.py
```diff
@@ -560,3 +564,97 @@ def ui(
...
+
+
+@output_transformer(
+    default_ui=download_button_wrapped
+)  # pyright: ignore[reportGeneralTypeIssues]
+async def DownloadTransformer(
+    _meta: TransformerMetadata,
+    _fn: DownloadHandler,
+    *,
+    button_label: Optional[TagChild] = None,
```
wch, line 585: The `button_label` gets passed here, but it currently isn't used, and doesn't end up in the UI. It would need to somehow go to the call to `download_button_wrapped()`.

schloerke: This should work great in #964

jcheng5, line 585: I think this would work today with:
```suggestion
@output_transformer(
    default_ui=ui.download_button,
    default_ui_passthrough_args=["label"],
)  # pyright: ignore[reportGeneralTypeIssues]
async def DownloadTransformer(
    _meta: TransformerMetadata,
    _fn: DownloadHandler,
    *,
    label: Optional[TagChild] = "Download",
```
Definitely will be easier with Barret's new class-based approach.

wch:
Closing in favor of #977.

PR #977, opened 2024-01-06
Add `render.download`, take 2

### LINKED ISSUE #971: Consider replacing `output_args()`
The `output_args()` function (added in #786) does the following: In Express, it allows passing arguments to the output function that's associated with a render function. For example, with the upcoming `render.download()` function (#967), it would look something like this:

```py
@output_args(label="Download file")
@render.download(filename="hello.txt")
def dl:
    yield f"Hello, world! n = {input.n()}"
```

One major drawback is that the function signature accepts just `*args` and `**kwargs`, so it can't give any help for argument completion or typing. In the example above, you just have to know that the output UI function is `download_button()`, and look up the parameter names (and types) for that function.

Currently, if you want finer control and completion support, you can use `@suspend_display` and call the output UI function directly:

```py
@suspend_display
@render.download(filename="hello.txt")
def dl:
    yield f"Hello, world! n = {input.n()}"

ui.download_button("dl", label="Download file")
```

The downside of this is that when the user just wants to customize a single parameter value, the code becomes a lot more verbose. And it's conceptually a bit of a leap to turn off the automatic UI, and then manually create it.

For the case of `render_plot()`, we've done something different: we added the output arguments to the `render.plot()` function. If we were to use that strategy here, it would look like this:

```py
@render.download(filename="hello.txt", label="Download file")
def dl:
    yield f"Hello, world! n = {input.n()}"
```

This is concise, but it has the following drawbacks:

- It's not immediately obvious to the user which arguments are for the output UI, and which are for the rendering step.
- Those arguments are also present for Shiny Core, which may be confusing.

Another way we could go is to have a method for setting UI output arguments, that chains off the renderer object:

```py
@render.download(filename="hello.txt").ui_args(label="Download file")
def dl:
    yield f"Hello, world! n = {input.n()}"
```

wch:
This PR adds `render.download`, and supersedes #967.

The existing API is to use `session.download`, like this:

```py
# UI contains:
ui.download_button("dl", "Download File")

# Server function contains:
@session.download(filename="hello.txt")
def dl():
    yield "Hello, world!\n"
```

This is quite a bit different from all other outputs, because it relies on the `session` object. This also doesn't work for Express apps, because there isn't a `session` object available. (I'll note here that it would be possible to add a `session` object that could be imported from `shiny.express`, but the change in this PR is a better one.)

With this PR, the app uses `render.download` instead of `session.download`:

```py
# UI contains:
ui.download_button("dl", "Download File")

# Server function contains:
@render.download(filename="hello.txt")
def dl():
    yield "Hello, world!\n"
```

I think this makes more sense from the app author's perspective.

In an Express app, the code would look like this (the button is automatically inserted in the UI):

```py
@render.download(filename="hello.txt")
def dl():
    yield "Hello, world!\n"
```

With Express, the label can also be set with `label`:

```py
@render.download(filename="hello.txt", label="Download!")
def dl():
    yield "Hello, world!\n"
```

Currently, `label` is the only parameter for `download_button` that's exposed in `render.download`.  There are others, like `icon` and `class_` which are relatively commonly used -- maybe it's best to wait until we resolve #971 before we decide what to do with these other parameters.

### DIFF
--- a/shiny/api-examples/download/app.py
+++ b/shiny/api-examples/download/app.py
@@ -7,7 +7,7 @@ from typing import Any
 import matplotlib.pyplot as plt
 import numpy as np
 
-from shiny import App, Inputs, Outputs, Session, ui
+from shiny import App, Inputs, Outputs, Session, render, ui
 
 
 def make_example(id: str, label: str, title: str, desc: str, extra: Any = None):
@@ -77,7 +77,7 @@ app_ui = ui.page_fluid(
 
 
 def server(input: Inputs, output: Outputs, session: Session):
-    @session.download()
+    @render.download()
     def download1():
         """
         This is the simplest case. The implementation simply returns the name of a file.
@@ -88,12 +88,12 @@ def server(input: Inputs, output: Outputs, session: Session):
         path = os.path.join(os.path.dirname(__file__), "mtcars.csv")
         return path
 
-    @session.download(filename="image.png")
+    @render.download(filename="image.png")
     def download2():
         """
         Another way to implement a file download is by yielding bytes; either all at
         once, like in this case, or by yielding multiple times. When using this
-        approach, you should pass a filename argument to @session.download, which
+        approach, you should pass a filename argument to @render.download, which
         determines what the browser will name the downloaded file.
         """
 
@@ -107,7 +107,7 @@ def server(input: Inputs, output: Outputs, session: Session):
             plt.savefig(buf, format="png")
             yield buf.getvalue()
 
-    @session.download(
+    @render.download(
         filename=lambda: f"新型-{date.today().isoformat()}-{np.random.randint(100,999)}.csv"
     )
     async def download3():
@@ -116,7 +116,8 @@ def server(input: Inputs, output: Outputs, session: Session):
         yield "新,1,2\n"
         yield "型,4,5\n"
 
-    @session.download(id="download4", filename="failuretest.txt")
+    @output(id="download4")
+    @render.download(filename="failuretest.txt")
     async def _():
         yield "hello"
         raise Exception("This error was caused intentionally")
--- a/shiny/api-examples/download_button/app.py
+++ b/shiny/api-examples/download_button/app.py
@@ -1,9 +1,8 @@
 import asyncio
+import random
 from datetime import date
 
-import numpy as np
-
-from shiny import App, Inputs, Outputs, Session, ui
+from shiny import App, Inputs, Outputs, Session, render, ui
 
 app_ui = ui.page_fluid(
     ui.download_button("downloadData", "Download"),
@@ -11,8 +10,8 @@ app_ui = ui.page_fluid(
 
 
 def server(input: Inputs, output: Outputs, session: Session):
-    @session.download(
-        filename=lambda: f"新型-{date.today().isoformat()}-{np.random.randint(100,999)}.csv"
+    @render.download(
+        filename=lambda: f"新型-{date.today().isoformat()}-{random.randint(100,999)}.csv"
     )
     async def downloadData():
         await asyncio.sleep(0.25)
--- a/shiny/api-examples/download_link/app.py
+++ b/shiny/api-examples/download_link/app.py
@@ -1,9 +1,8 @@
 import asyncio
+import random
 from datetime import date
 
-import numpy as np
-
-from shiny import App, Inputs, Outputs, Session, ui
+from shiny import App, Inputs, Outputs, Session, render, ui
 
 app_ui = ui.page_fluid(
     ui.download_link("downloadData", "Download"),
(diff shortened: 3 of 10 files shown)


shiny/api-examples/download/app.py
```diff
@@ -77,7 +77,7 @@ def make_example(id: str, label: str, title: str, desc: str, extra: Any = None):
 
 
 def server(input: Inputs, output: Outputs, session: Session):
```
schloerke, line 79:
```suggestion
def server(input: Inputs):
```

shiny/api-examples/download/app.py
```diff
@@ -7,7 +7,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 
-from shiny import App, Inputs, Outputs, Session, ui
+from shiny import App, Inputs, Outputs, Session, render, ui
```
schloerke, line 10:
```suggestion
from shiny import App, Inputs, render, ui
```

shiny/api-examples/download_button/app.py
```diff
@@ -1,18 +1,17 @@
 import asyncio
+import random
 from datetime import date
 
-import numpy as np
-
-from shiny import App, Inputs, Outputs, Session, ui
+from shiny import App, Inputs, Outputs, Session, render, ui
```
schloerke, line 5:
```suggestion
from shiny import App, Inputs, render, ui
```

shiny/api-examples/download_button/app.py
```diff
@@ -1,18 +1,17 @@
...
 
 app_ui = ui.page_fluid(
     ui.download_button("downloadData", "Download"),
 )
 
 
 def server(input: Inputs, output: Outputs, session: Session):
```
schloerke, line 12:
```suggestion
def server(input: Inputs):
```

shiny/api-examples/download_link/app.py
```diff
@@ -1,18 +1,17 @@
 import asyncio
+import random
 from datetime import date
 
-import numpy as np
-
-from shiny import App, Inputs, Outputs, Session, ui
+from shiny import App, Inputs, Outputs, Session, render, ui
```
schloerke, line 5:
```suggestion
from shiny import App, Inputs, render, ui
```

shiny/api-examples/download_link/app.py
```diff
@@ -1,18 +1,17 @@
...
 
 app_ui = ui.page_fluid(
     ui.download_link("downloadData", "Download"),
 )
 
 
 def server(input: Inputs, output: Outputs, session: Session):
```
schloerke, line 12:
```suggestion
def server(input: Inputs):
```

shiny/render/_render.py
```diff
@@ -492,3 +496,101 @@ async def transform(self, value: TagChild) -> Jsonifiable:
...
+
+            session = require_active_session(None)
+            return f"session/{quote(session.id)}/download/{quote(self.output_id)}?w="
+
+        # Unlike most value functions, this one's name is `url`. But we want to get the
+        # name from the user-supplied function.
+        url.__name__ = fn.__name__
+
+        # We invoke `super().__call__()` now, because it indirectly invokes
+        # `Outputs.__call__()`, which sets `self.name`, which is then used below.
```
schloerke, line 578:
```suggestion
        # `Outputs.__call__()`, which sets `self.output_id` (and `self.__name__`), which is then used below.
```

shiny/session/_session.py
```diff
@@ -748,7 +749,7 @@ def download(
         encoding: str = "utf-8",
     ) -> Callable[[DownloadHandler], None]:
         """
-        Decorator to register a function to handle a download.
+        Deprecated. Please use :func:`~shiny.render.download` instead.
```
schloerke, line 752:
```suggestion
        Deprecated. Please use :class:`~shiny.render.download` instead.
```

shiny/ui/_download_button.py
```diff
@@ -41,7 +41,7 @@ def download_button(
 
     See Also
     --------
-    ~shiny.Session.download
+    ~shiny.render.download
     ~shiny.ui.download_link
```
schloerke, line 45:
```suggestion
    * :class:`~shiny.render.download`
    * :func:`~shiny.ui.download_link`
```

shiny/ui/_download_button.py
```diff
@@ -96,7 +96,7 @@ def download_link(
...
```
schloerke, line 100:
```suggestion
    * :class:`~shiny.render.download`
    * :func:`~shiny.ui.download_link`
```

schloerke (approved):
LGTM! (given cosmetic tweaks)

wch:
Going to hold the `output` and `Session` removal for later, since those are already covered by #793.

PR #978, opened 2024-01-06
Add underscores to hide some imports

### LINKED ISSUE #972: Shiny express seems to export more than it should
Looks like `Inputs`, `Outputs`, `Session`, `annonations`, and `app` are being included by mistake?

```python
>>> import shiny.express as sx
>>> sx.[TAB]
sx.Inputs(            sx.annotations          sx.display_decorator   sx.suspend_display(
sx.Outputs(         sx.app                        sx.is_express_app(     sx.ui
sx.Session(         sx.display_body()     sx.output_args(            sx.wrap_express_app(
```

I haven't validated this, but my gut feeling is that these issues derive from the fact that `express` isn’t included in `shiny/__init__.py`...

wch:
Closes #972.

The behavior with this change:

```
>>> import shiny.express as sx
>>> sx.[TAB]
sx.annotations        sx.display_body()     sx.is_express_app(    sx.suspend_display(   sx.wrap_express_app(  
sx.app                sx.display_decorator  sx.output_args(       sx.ui                 
```

### DIFF
--- a/shiny/express/__init__.py
+++ b/shiny/express/__init__.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
-from ..session import Inputs, Outputs, Session
+# Import these with underscore names so they won't show in autocomplete from the Python
+# console.
+from ..session import Inputs as _Inputs, Outputs as _Outputs, Session as _Session
 from ..session import _utils as _session_utils
 from . import app, ui
 from ._is_express import is_express_app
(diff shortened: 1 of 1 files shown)


cpsievert:
I still see `shiny.express.annotations` and `shiny.express.app`. I'm not sure if we can get rid of the former, but maybe the latter?

wch:
I agree about `app`, so I made the change and pushed.

cpsievert (approved):
Thanks!

PR #983, opened 2024-01-09
Fixes for `shiny.express.ui.card()`

### LINKED ISSUE #948: `shiny.express.ui.card` doesn't render card items properly
For example:

```python
from shiny.express import ui

with ui.card():
    ui.card_header("Header")
    "Body"
```

Screenshot 2023-12-28 at 3 22 47 PM

Both the header and the body are being wrapped in a `.card-body`, which is leading to the extra, unwanted, padding.

### LINKED ISSUE #984: Express mode should auto opt into using the `www` folder
Currently there is no way to add a static handler in Express mode. I believe if the `www` folder exists next to the app, Express mode (and possibly Core) should opt in to supporting the static folder.

cc @karangattu

cpsievert:
Closes #948

### DIFF
--- a/shiny/ui/_card.py
+++ b/shiny/ui/_card.py
@@ -126,7 +126,7 @@ def _card_impl(
         wrapper = card_body
 
     attrs, children = consolidate_attrs(*args, class_=class_, **kwargs)
-    children = _wrap_children_in_card(*children, wrapper=wrapper)
+    children = _as_card_items(*children, wrapper=wrapper)
 
     tag = div(
         {
@@ -202,9 +202,9 @@ class WrapperCallable(Protocol):
 
 
 def _as_card_items(
-    *children: TagChild | CardItem | None,  # `TagAttrs` are not allowed here
+    *children: TagChild,  # `TagAttrs` are not allowed here
     wrapper: WrapperCallable | None,
-) -> list[CardItem]:
+) -> list[TagChild]:
     # We don't want `None`s creating empty card bodies
     children_vals = [child for child in children if child is not None]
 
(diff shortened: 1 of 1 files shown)


wch:
This code reaches into context-manager children in a way that's very specific to this one use case and also feels a bit brittle. I think this actually points to a more general issue that we're going to have to deal with.

Currently, in Shiny Core we have two stages of dealing with the tree of components:
- Building the tree with nested function calls. At this point, the tree may consist of  strings, `Tag`s, and `Tagifiable` objects (and a few other types which I'll ignore for this discussion).
- Calling `.tagify()` on the top-level node recurses through the tree and converts all nodes to strings, and `Tag`s.

With Shiny Express, there's another type of object that can be in the tree: `RecallContextManager`s. These are actually `Tagifiable`, but the problem is that some functions, like `card()`, want to check if their children are a specific class, like `CardItem`, before deciding what to do with them. This happens before the `.tagify()` step. I think we'll need a generalizable way of handling `RecallContextManager`s, because this issue will likely crop up in other places.

Maybe the steps should be something like this:

- Build the tree with strings, `Tag`s, `Tagifiable`s, and `RecallContextManager`s. As the tree is being built, functions can 
check if their children are `RecallContextManagers`, and if so, call a method`.resolve()`, which looks like this:
    ```py
    def resolve(self):
        return self.fn(*self.args, **self.kwargs)
    ```
    This would only be necessary to do in functions like `card_body()` which check if their children are objects of specific classes (that are not `Tag`).

wch:
Closing because this was superseded by #992.

PR #992, opened 2024-01-10
Cause RecallContextManagers to run when used without `with`

### LINKED ISSUE #948: `shiny.express.ui.card` doesn't render card items properly

wch:
Closes #948.


jcheng5:
I think this is subtle enough behavior to deserve a unit test, other than that LGTM

PR #997, opened 2024-01-12
Add `render.code()`

wch:
This PR adds a function `render.code()`, and, in Shiny Express, changes the output behavior for `render.text()`.

In Shiny Express:
- `render.text()` now uses `output_text()` as its output function, instead of `output_text_verbatim()`.
- The new `render.code()` uses `output_text_verbatim()` as its output function.

In Shiny Core:
- `render.text()` and `render.code()` do the same thing, although the documentation now suggests pairing `render.text` with `output_text()`, and `render.code()` with `output_text_verbatim()`.

In the future, when we add `output_code()` (perhaps with line numbers and syntax highlighting), we can change `render.code()` to use that output function instead of `output_text_verbatim()`.

This is a screenshot of usage in a Shiny Express app:

image


wch:
The changes here were incorporated in #1018.

PR #999, opened 2024-01-12
Calculate coordinate mapping after drawing figure

has2k1:
To calculate accurate coordinate mappings for matplotlib (& plotnine) plots
in display coordinates, the figure should know the location of it's subplots.

The location determined by the subplot parameters (subplot_params) and they
come from one or more of these sources, listed from lowest to highest priority:

  1. rcParams
  2. plt.subplots(..., gridspec_kw={..., **subplot_params})
  3. fig.subplot_adjust(**subplot_params)
  4. layout engine

The layout engine has the last say as it dynamically calculates subplot_params
and calls fig.subplot_adjust() to set them.

There are two things to consider:

  1. Only layout engines that are "adjust_compatible" calculate subplot_params
  2. Matplotlib runs the layout engine only when drawing the figure

This PR ensures that there is an "adjust_compatible" layout engine and the plot
is drawn/saved before the coordinate mappings are calculated.

closes has2k1/plotnine#738

### DIFF
--- a/shiny/render/_try_render_plot.py
+++ b/shiny/render/_try_render_plot.py
@@ -159,16 +159,27 @@ def try_render_matplotlib(
         )
         fig.set_dpi(ppi_out * pixelratio)
 
-        # Suppress the message `UserWarning: The figure layout has changed to tight`
-        with warnings.catch_warnings():
-            warnings.filterwarnings(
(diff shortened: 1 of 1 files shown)


wch:
Thank you!

PR #1004, opened 2024-01-13
Remove `express.ui.output_*` functions, add output args to render functions

wch:
This PR does the following:
- Removes the `output_*` functions from `shiny.express.ui`.
- Adds arguments for the `shiny.ui.output_*` functions to the `shiny.render.*` functions.
- Deprecates `output_args()`. (Probably no one is using this anyway.)

A few things yet to do:

- [ ] Come up with a solution for name collisions for some `render.*` and `output_*` function arguments, like `**kwargs`, `width`, and `height`. `kwargs` handling should be consistent across all functions that use them: `render_plot`, `render_image`, `render_table`, `render_ui`, `render_table`, `render_download`
- [ ] Update all `default_ui()` methods so that they only accept an `id` argument.
- [ ] Resolve this message: https://github.com/posit-dev/py-shiny/blob/b4ed7ff15907846ea6e88e57223e4cca2a95c9c2/shiny/render/transformer/_transformer.py#L3
- [x] Update remaining tests that use `ui_kwargs`.


wch:
Closing in favor of #1018.

PR #1009, opened 2024-01-16
Pin starlette version below 0.35

nealrichardson:
There have been issues reported where Shiny apps deployed to Posit Connect break with the latest release of starlette/fastapi. This should keep folks from running into that (for now; we can remove this after a while once Posit Connect customers have time to upgrade with the fix).


wch (approved):
Thanks!

PR #1013, opened 2024-01-16
Update `InjectAutoreloadMiddleware` to be compatible with starlette >= 0.35.0

schloerke:
Related: https://github.com/encode/starlette/pull/2381

Do not merge until motivation for https://github.com/posit-dev/py-shiny/pull/1009 is resolved


wch:
I think the root of this problem is: 
starlette claims that you can use types from starlette _or_ types from asgiref, where the latter are more rigorous (and what we have in the existing code). They show the two methods here:
https://github.com/encode/starlette/blob/39dccd9/docs/middleware.md#type-annotations

However, in https://github.com/encode/starlette/pull/2381, the explicit types they added are the starlette types, not the asgiref types, and that's why we started getting type-checking errors.

The change in this PR makes it so, on the outside, it uses the starlette types, but on the inside, it uses the asgiref types.

I think it would be better if the external interface supported both types.

PR #1016, opened 2024-01-16
`express.ui.page_opts(title = ...)` now always generates a header

### LINKED ISSUE #1015: `express.ui.page_opts(title = ...)` is sometimes not visible on the page
For example, the `Foo` here isn't visible on the page (it only gets used for the browser tab's title)

```python
from shiny.express import ui

ui.page_opts(title="Foo")

"Bar"
```

For consistency, `ui.page_opts(title="Foo")` should always generate a visible result

cpsievert:
Closes #1015

### DIFF
--- a/shiny/express/ui/_page.py
+++ b/shiny/express/ui/_page.py
@@ -19,6 +19,7 @@ def page_auto_cm() -> RecallContextManager[Tag]:
 def page_opts(
     *,
     title: str | MISSING_TYPE = MISSING,
+    window_title: str | MISSING_TYPE = MISSING,
     lang: str | MISSING_TYPE = MISSING,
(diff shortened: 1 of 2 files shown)


wch:
It would be useful to be able to set the window title separately from the on-page title. How about having two separate args, `window_title` and `title`, and if `window_title` isn't provided, it defaults to the value of `title`?

cpsievert:
Yea, we can definitely add that, but I think it'll time a bit of time to do properly, let's maybe add that as an item for the next release and take this now?

wch:
I just went ahead and implemented the `window_title` and `title` behavior.

wch:
Interesting... `title` is expected to be a string based on the type hints, but I suppose it is possible that someone would pass in a `Tag`/`Taglist` object. I'll file an issue about it.

PR #1026, opened 2024-01-17
chore: Expose `render.renderer.RendererBaseT` and do not require `| None` when defining Renderer types

schloerke:


schloerke:
(Merging early)

PR #1035, opened 2024-01-19
docs: Add Renderer docs

schloerke:
Followup to #964


shiny/render/renderer/_renderer.py
```diff
@@ -83,22 +88,46 @@
...
     """
     Output renderer class
 
-    TODO-barret-docs
+    An output renderer is a class that will take in a callable function (value
+    function), transform the returned value into a JSON-serializable object, and send
+    the result to the browser.
+
+    When the value function is received, it the renderer will be auto registered with
+    the current session's `Output` class, hookinng it into Shiny's reactive graph. By
```
cpsievert, line 113:
```suggestion
    the current session's `Output` class, hooking it into Shiny's reactive graph. By
```

shiny/render/renderer/_renderer.py
```diff
@@ -83,22 +88,46 @@
...
+    auto registering as an `Output`, it allows for App authors to skip adding `@output`
+    above the renderer. (If progromatic `id` is needed, `@output(id="foo")` can still be
```
cpsievert, line 115:
```suggestion
    above the renderer. (If programmatic `id` is needed, `@output(id="foo")` can still be
```

shiny/render/renderer/_renderer.py
```diff
@@ -203,23 +255,39 @@ def __init__(
 
     async def transform(self, value: IT) -> Jsonifiable:
         """
-        Renderer - transform docs here
-
-        TODO-barret-docs
+        Transform an non-`None` output value into a JSON-serializable object.
```
cpsievert, line 258:
```suggestion
        Transform an output value into a JSON-serializable object.
```

shiny/render/renderer/_renderer.py
```diff
@@ -203,23 +255,39 @@ def __init__(
...
+
+        If an `.render()` method is not implemented, this method **must** be
```
cpsievert, line 260:
```suggestion
        If a `.render()` method is not implemented, this method **must** be
```

cpsievert, line 260: I think it would be good to start out by explaining that, for most situations, you don't need to implement `.transform()` (and you should implement `.render()` instead?)

schloerke: It's the other way around. 

90%+ of the time you should probably implement the transform method. The other 10%, you'll need to implement the render method.

shiny/render/renderer/_renderer.py
```diff
@@ -203,23 +255,39 @@ def __init__(
...
+            " method on non-`None` values."
         )
 
     async def render(self) -> Jsonifiable:
         """
-        Renderer - render docs here
+        Renders the output value function.
+
+        This method is called when the renderer is requested to render its output. It
+        should resolve the value function `.fn` and return a (most likely) transformed
```
cpsievert, line 283: I'm a little confused why you're using `resolve` instead of `call`/`execute`?

shiny/render/renderer/_renderer.py
```diff
@@ -203,23 +255,39 @@ def __init__(
...
+        value.
 
-        TODO-barret-docs
+        The current implementation is to resolve the value function. If the value is
```
cpsievert, line 286:
```suggestion
        The default implementation is to resolve the value function. If the value is
```

shiny/render/renderer/_renderer.py
schloerke: I'm up for `call`!

```suggestion
        should call the value function `.fn` and return a (most likely) transformed
```

schloerke: Switching to `execute` 😆 
```suggestion
        should execute the value function `.fn` and return a (most likely) transformed
```

cpsievert: Cool, I think it gets used in a few other places too

PR #1037, opened 2024-01-19
Rename `suspend_display`->`hold`, `display_body`->`expressify`, `render.display`-> `render.express`

wch:
This PR supersedes #996.  It does the following:

- Rename `express.suspend_display` -> `express.ui.hold`
- Rename `express.display_body` -> `express.expressify`
- Rename `render.display` -> `render.express`

Also, for `hold`:
- `hold()` now can _only_ be used a context manager, and not as a decorator.
- You can do `with ui.hold() as x`, and then use `x` later on the page to display it.

Note that `suspend_display` is basically an alias for `hold`, and in this PR it has lost the ability to be used as a decorator, which could break code that was using the previous (experimental) API.

Example usage for `hold`:

```py
import shiny.ui

# `ui.hold() as x` can be used to save `x` for later output
with ui.hold() as hello_card:
    with ui.card():
        with ui.span():
            "This is a"
            ui.span(" card", style="color: red;")

hello_card
hello_card


# ================================================

# `ui.hold()` can be used to just suppress output
with ui.hold():
    @render.text()
    def txt():
        return f"Slider value: {input.n()}"


ui.input_slider("n", "N", 1, 100, 50)
shiny.ui.output_text("txt", inline=True)
```

Example for `expressify`:

```py
from shiny.express import expressify, ui

# @expressify converts a function to work with Express syntax
@expressify
def expressified1(s: str):
    f"Expressified function 1: {s}"
    ui.br()


expressified1("Hello")
expressified1("world")
```

Example for `render.express`:

```py

ui.input_slider("n", "N", 1, 100, 50)

# @render.express is like @render.ui, but with Express syntax
@render.express
def render_express1():
    "Slider value:"
    input.n()
```


shiny/express/ui/_hold.py
```diff
@@ -0,0 +1,73 @@
...
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_value: Optional[BaseException],
+        traceback: Optional[TracebackType],
+    ) -> bool:
+        sys.displayhook = self.prev_displayhook
+        if exc_type:
+            print(f"An exception occurred: {exc_value}")
+        return False
```
jcheng5, line 73: Is there a reason you're printing the exception here? I wouldn't expect `with ui.hold()` to get involved with exceptions.

shiny/render/_express.py
```diff
@@ -17,7 +17,7 @@
 )
 
 
-class display(Renderer[None]):
+class express(Renderer[None]):
```
jcheng5, line 20: No docstrings for this yet? (Or do they live somewhere else?)

shiny/express/ui/_hold.py
wch: Oh, no, that was some leftover code. I'll remove.

Issue #1034, opened 2024-01-19
Updating express inputs leads to an error


cpsievert:
Consider the following app, which from a user perspective is functional, but does generate an error (since the 1st time it runs a session hasn't been established):

```python
from shiny import reactive
from shiny.express import ui

ui.input_slider("n", "N", 1, 100, 50)

@reactive.effect
def _():
    ui.update_slider("n", value=10)
```

```python
Traceback (most recent call last):
  File "/Users/cpsievert/github/py-shiny/shiny/reactive/_reactives.py", line 573, in _run
    await self._fn()
  File "/Users/cpsievert/github/py-shiny/shiny/_utils.py", line 243, in fn_async
    return fn(*args, **kwargs)
  File "/Users/cpsievert/github/py-shiny/sandbox/app.py", line 8, in _
    ui.update_slider("n", value=10)
  File "/Users/cpsievert/github/py-shiny/shiny/ui/_input_update.py", line 780, in update_slider
    session = require_active_session(session)
  File "/Users/cpsievert/github/py-shiny/shiny/session/_utils.py", line 127, in require_active_session
    raise RuntimeError(
RuntimeError: update_slider() must be called from within an active Shiny session.
/Users/cpsievert/github/py-shiny/shiny/reactive/_reactives.py:555: ReactiveWarning: Error in Effect: update_slider() must be called from within an active Shiny session.
  await self._run()
```

wch:
This is a real issue, but I will note that in practice, most uses will only call the `update_*` function after something else has happened. For example, it might use `reactive.event`. The example below will not throw an error:

```py
from shiny import reactive
from shiny.express import input, ui

ui.input_slider("m", "M", 1, 100, 10)
ui.input_slider("n", "N", 1, 100, 50)


@reactive.effect
@reactive.event(input.m)
def _():
    ui.update_slider("n", value=input.m())
```

cpsievert:
That's true, but it's also not great that if you try use any `session` methods, you get an unhelpful error. For example:

```python
from shiny import reactive
from shiny.express import session

@reactive.effect
async def _():
    await session.send_custom_message("msg", {})
```

```python
Error in Effect: '_MockSession' object has no attribute 'send_custom_message'
```

You also can't avoid the error with something like `if session:` since the mock session object is truthy

jcheng5:
[...]

PR #1038, opened 2024-01-19
Set a MockSession context for the initial execution of express code

cpsievert:
This is currently a quick and dirty draft at addressing #1034.

The main idea here is that if you try to access session attributes on the 1st run of an express app, you get an informative error (to wrap that code in `if session:` ), so that something like this can work without error:

```python
from shiny import reactive
from shiny.express import session, ui

@reactive.effect
async def _():
    if session:
        x = {"message": "Hello from Python!"}
        await session.send_custom_message("send_alert", x)

ui.tags.script(
    """
    Shiny.addCustomMessageHandler("send_alert", function(x) {
        document.body.innerHTML = x.message;
    });
    """
)
```

Also, more controversially, `require_active_session()` will now throw a silent exception when there is no actual session inside a reactive context, so that something like this just works without error:

```python
from shiny import reactive, session
from shiny.express import ui

ui.input_slider("n", "N", 1, 100, 50)

@reactive.effect
def _():
    ui.update_slider("n", value=10)
```

### DIFF
--- a/shiny/express/__init__.py
+++ b/shiny/express/__init__.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
+from typing import cast
+
 # Import these with underscore names so they won't show in autocomplete from the Python
 # console.
-from ..session import Inputs as _Inputs, Outputs as _Outputs, Session as _Session
-from ..session import _utils as _session_utils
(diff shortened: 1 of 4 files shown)


wch:
Closing because we had a different fix in #1049.

PR #1043, opened 2024-01-20
Restore support for matplotlib<3.6

wch:
PR #999 depended on matplotlib's layout engine API, which was added in matplotlib 3.6.

Matplotlib is currently on 3.8, so normally this would be OK. However, the version of Pyodide we're using in Shinylive comes with matplotlib 3.5.2, so plots don't work at all on that platform -- it simply throws an error when `fig.get_layout_engine()` is called. This PR restores support for older versions of matplotlib.

@has2k1 I'm going to merge this soon to get the dev version of shinylive working again, but please take a look at the PR when you have a chance to make sure it looks good.


has2k1:
Retrospectively, this looks good.

PR #1055, opened 2024-01-24
test: Test apps locally before deploying via pytest fixtures.

karangattu:
Allow to run the deploy tests on CI but skip deploy tests when running locally.


schloerke:
Winston requested changes to move the deploys around. This clashes with my original work request, so I will revert the behavior.

.github/workflows/pytest.yaml
```diff
@@ -99,7 +99,15 @@ jobs:
       - name: Run example app tests
         timeout-minutes: 20
         run: |
-          make playwright-examples
+          make playwright-examples SUB_FILE=". -vv"
+
+      - name: Test deploys example apps on python 3.10
```
karangattu, line 104: Move this test to `deploys` GHA job

PR #1060, opened 2024-01-24
Enable error console when running locally.

wch:
This PR enables the error console when running shiny apps locally.

When apps are run locally, they are launched with `shiny run`; when they are run on a server (like Connect) they are launched with `uvicorn`, and in that case the error console will not be enabled.


jcheng5:
I wonder if we should make it based on `--reload` instead? I was thinking about suggesting that we change the way Connect and SSOS load Shiny apps, I think there are plenty of other ways we could have them inject their middleware than to load the Shiny app object and wrap it.

jcheng5:
After discussion with @wch in person, the important thing we want to ensure is that Connect never shows the error console by default, so we should write a automated test that verifies that and this PR can be merged as-is.

PR #1064, opened 2024-01-24
Add `fill` to `__all__` in `ui` and `express.ui`

wch:
`fill` is currently imported into `shiny.ui`, but are not listed in the `__all__` in that file. This PR:

- Adds it to `__all__` in `shiny.ui`
- Imports it into `shiny.express.ui` and also adds them to `__all__` in that file.

Note that this PR doesn't do the same for `css`. This is because `css` is meant for internal use.

### DIFF
--- a/shiny/ui/__init__.py
+++ b/shiny/ui/__init__.py
@@ -31,8 +31,12 @@ from htmltools import (
     tags,
 )
 
-# Expose the following modules for extended usage: ex: ui.fill.as_fill_item(x)
-from . import css, fill  # noqa: F401  # pyright: ignore[reportUnusedImport]
+from . import (
+    # The css module is for internal use, so we won't re-export it.
+    css,  # noqa: F401  # pyright: ignore[reportUnusedImport]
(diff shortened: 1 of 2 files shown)


cpsievert (approved):
Thanks!

shiny/ui/__init__.py
```diff
@@ -31,8 +31,12 @@
...
 )
 
-# Expose the following modules for extended usage: ex: ui.fill.as_fill_item(x)
-from . import css, fill  # noqa: F401  # pyright: ignore[reportUnusedImport]
+from . import (
+    # The css module is for internal use, so we won't re-export it.
+    css,  # noqa: F401  # pyright: ignore[reportUnusedImport]
+    # Expose the fill module for extended usage: ex: ui.fill.as_fill_item(x).
+    fill,
+)
```
cpsievert, line 39: Oh, these might need to be imported separately?

wch: Good catch, thanks.

Issue #1083, opened 2024-01-28
Need stronger ui.hold()-like context manager for non-UI objects


jcheng5:
We need another context manager, maybe `ui.discard()`, to prevent Shiny Express from trying to interpret non-UI objects as UI. Or, perhaps Shiny Express should not be so strict about what kinds of objects are allowed to be displayhooked (no-op instead of error).

```python
from shiny.express import ui

class A:
    def __init__(self):
        self.options = {}

    def set_option(self, name, value):
        self.options[name] = value
        
        # for method chaining purposes
        return self

a = A()

# This next line causes the error.
# Enclosing the line in `with ui.hold()` does not help.
a.set_option("host", "localhost")

with ui.card():
    "Hello"
```

jcheng5:
We could also have a more imperative "echo on/off" type of function.

```python
ui.echo(False)

# Do lots of non-UI stuff

ui.echo(True)
```

You wouldn't even have to wrap it in try/finally, instead you could say that if there is an enclosing UI context manager (like `ui.card()`), the effect of `ui.echo(False)` does not survive beyond the end of that context manager.

The nice thing about this approach is not having to indent the non-UI stuff.

There's a good chance this is a terrible idea and I'm not thinking clearly about it, it's pretty late.

wch:
The specific problem with `ui.hold()` is that it's a `RecallContextManager` that wraps a `TagList`, and the `TagList` cares about what goes into it. This is so all the things in the `with ui.hold() as x:` block can be referred to as a single object `x`. But now that I think about it, I think it could be changed to a regular list without breaking anything.

Or another alternative: the objects get put in a subclass of list, like `HoldList`. If a function consumes that object and doesn't know about `HoldList`s, then it gets treated like a list. But if function consumes it and and _does_ know about `HoldList`s (like a `RecallContextManager`) then it can use the spread operator on it.

wch:
Another thought that I had was that perhaps `RecallContextManager` could add a `filter` argument, so we could have something like:

```py
def page_auto_cm(
    *,
    filter: Callable[[object], bool] | None = None,
) -> RecallContextManager[Tag]:
    return RecallContextManager(
        ui.page_auto,
        filter=filter,
    )
```

Where the `filter` returns `True` if the displayhooked objects should be stored by the `RecallContextManager`, `False` otherwise.

And then the default `filter` function could be one that returns `True` only for `TagChild` objects.


[...]

PR #1089, opened 2024-01-30
Update express.ui.hold() to accept any type of object

wch:
This addresses one of the issues described in #1083.

### DIFF
--- a/shiny/express/ui/_hold.py
+++ b/shiny/express/ui/_hold.py
@@ -6,7 +6,6 @@ from typing import Callable, Optional, Type, TypeVar
 
 from htmltools import wrap_displayhook_handler
 
-from ... import ui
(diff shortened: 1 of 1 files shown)


jcheng5 (approved):

wch:
For future reference: The result of `ui.hold()` is a list. If a `RecallContextManager` wants to consume the result of a `ui.hold()`, it could simply use a `for` loop to iterate over the list:

```py
with ui.hold() as held:
    "a"
    "b"

with ui.card():
    for x in held:
        x
```

It could also be done in a function like this:

```py
@expressify
def spread(lst: list[object]) -> None:
    for x in lst:
        x

with ui.hold() as held:
    "a"
    " b "

with ui.card():
    spread(held)
```

PR #1094, opened 2024-02-01
Update types to allow `navset_*` functions to accept `MetadataNode` args

wch:
I noticed that many of the `navset_*` functions did not accept `MetadataNode` arguments, but they should, in case someone wants to pass in `HTMLDependency` objects.

For reference, the ability to accept `MetadataNode` args was added to many `navset_*` functions in #455, but some newer `navset_*` components did not have this typing.

### DIFF
--- a/shiny/ui/_navs.py
+++ b/shiny/ui/_navs.py
@@ -20,6 +20,7 @@ __all__ = (
     "nav",
 )
 
+import collections.abc
 import copy
 import re
 from typing import Any, Literal, Optional, Sequence, cast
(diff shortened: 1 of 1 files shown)


wch:
Now that I look at it more closely, perhaps these functions should also accept `Sequence[MetadatNode]`, as is already done in `navset_bar()` on `main`:

https://github.com/posit-dev/py-shiny/blob/aba94597c6b0c913dc92952f0c21f92fd8507515/shiny/ui/_navs.py#L1152-L1153

cpsievert (approved):

PR #1095, opened 2024-02-01
Add filtering to `RecallContextManager`

wch:
This PR implements the change mentioned in https://github.com/posit-dev/py-shiny/issues/1083#issuecomment-1913696858 and the following comment.

Some notes and things that remain to be done:

- The `UiRecallContextManager` class might not actually necessary, because it simply adds a different default `filter` function.
- The `filter_ui_objects` function is meant to return `True` for `TagChild` or `TagAttrs` or `None` objects. However, there's not really a good way to check if something is a `TagChild` or `TagAttrs`, since those aren't classes, but instead Union types which are recursive. I suppose we could recurse into the objects and check the entire tree to see if everything fits into the type, but that seems pretty crude. This issue is the most questionable part of this approach.
- [ ] Some of the UI functions accept just `TagChild` args, while others accept `TagChild | TagAttrs`, but the `filter_ui_objects` function currently only checks for the latter. There should be a different filter function that checks for the former.

### DIFF
--- a/shiny/express/_recall_context.py
+++ b/shiny/express/_recall_context.py
@@ -5,9 +5,17 @@ import sys
 from types import TracebackType
 from typing import Callable, Generic, Mapping, Optional, Type, TypeVar
 
-from htmltools import MetadataNode, Tag, TagList, wrap_displayhook_handler
+from htmltools import (
+    MetadataNode,
+    Tag,
+    TagAttrs,
+    TagChild,
+    Tagifiable,
+    TagList,
+    wrap_displayhook_handler,
+)
 
-from .._typing_extensions import ParamSpec
+from .._typing_extensions import ParamSpec, TypeGuard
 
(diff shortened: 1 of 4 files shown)


jcheng5:
@wch and I met on this but it was cut short. My notes:

1. For this release, the behavior we agreed to was that an error would still be thrown, but it would identify the file and line number (and ideally, code) that caused the problem. (It would be good if the code was printed at the console but it is important that the code is NOT rendered in the browser, for security/privacy)
2. In order to accomplish ☝️ we need to throw at the time the problematic object is encountered, not when the enclosing context manager exits. So `filter` or something like it is still needed.
3. Rather than filter--which can only allow, ignore, or throw--I'd prefer a method override. This would still permit allow, ignore, and throw, but also "proceed but with modified input", "proceed but I'll modify the return value", "execute more than once", etc.

Issue #1079, opened 2024-01-26
reactive.poll doesn't poll when defined in Express-adjacent modules


jcheng5:
While trying to port this app: https://github.com/posit-dev/py-shiny/blob/9ad2ab12ae0054ac745efed78da5fceb1e09fc43/shiny/api-examples/poll/app.py

I tried defining a `@reactive.poll` in a separate file (data.py), and then using it from the Express app.

<details>
<summary>Full source code</summary>

### app.py

```python
from shiny import reactive
from shiny.express import input, render, ui

from data import stock_quotes, SYMBOLS

ui.markdown(
    """
    # `shiny.reactive.poll` demo

    This example app shows how to stream results from a database (in this
    case, an in-memory sqlite3) with the help of `shiny.reactive.poll`.
    """
)

ui.input_selectize("symbols", "Filter by symbol", [""] + SYMBOLS, multiple=True)


def filtered_quotes():
    df = stock_quotes()
    if input.symbols():
        df = df[df["symbol"].isin(input.symbols())]
    return df

@render.express
def table():
    ui.HTML(
        filtered_quotes().to_html(
            index=False, classes="table font-monospace w-auto"
        )
    )
```

### data.py
```python
import asyncio
import random
import sqlite3
from datetime import datetime
from typing import Any, Awaitable

import pandas as pd

from shiny import reactive

SYMBOLS = ["AAA", "BBB", "CCC", "DDD", "EEE", "FFF"]


def timestamp() -> str:
    return datetime.now().strftime("%x %X")


def rand_price() -> float:
    return round(random.random() * 250, 2)


# === Initialize the database =========================================


def init_db(con: sqlite3.Connection) -> None:
    cur = con.cursor()
    try:
        cur.executescript(
            """
            CREATE TABLE stock_quotes (timestamp text, symbol text, price real);
            CREATE INDEX idx_timestamp ON stock_quotes (timestamp);
            """
        )
        cur.executemany(
            "INSERT INTO stock_quotes (timestamp, symbol, price) VALUES (?, ?, ?)",
            [(timestamp(), symbol, rand_price()) for symbol in SYMBOLS],
        )
        con.commit()
    finally:
        cur.close()


conn = sqlite3.connect(":memory:")
init_db(conn)


# === Randomly update the database with an asyncio.task ==============


def update_db(con: sqlite3.Connection) -> None:
    """Update a single stock price entry at random"""

    cur = con.cursor()
    try:
[...]

PR #1101, opened 2024-02-02
Add `shiny.express.app_globals()`

wch:
This is a possible solution for #1079.

Usage:

```py
## app.py ##
from shiny.express import app_globals, render

with app_globals():
    from shared import x

@render.text
def txt():
    return f"The value of shared.x() is {x()}"


## shared.py ##
from shiny import reactive

@reactive.poll(check_fn, 1)
def x():
    ...

def check_fn():
    ...
```


wch:
Closing in favor of #1172.

PR #1108, opened 2024-02-05
Fix input_task_button for modules

### LINKED ISSUE #1107: `ui.input_task_button` doesn't work inside module
It seems to me that the freshly introduced `input_task_button` doesn't work with shiny modules.

The following code:
```python
from shiny import App, module, reactive, render, ui


@module.ui
def button_ui():
    return ui.p(
        # ui.input_task_button("btn", label="Click me!"),
        ui.input_action_button("btn", label="Click me!"),
        ui.output_text("text_counter"),
    )


@module.server
def button_server(input, output, session):
    counter = reactive.Value(0)
    @render.text
    def text_counter():
        return f"Button clicked {counter()} times"
    
    @reactive.effect
    @reactive.event(input.btn)
    def increment_counter():
        counter.set(counter() + 1)


app_ui = ui.page_fluid(
    button_ui("button")
)


    button_server("button")


app = App(app_ui, server)
```

Work with `ui.input_action_button` it works, but with `ui.input_task_button` I get the following message in the terminal:

```
INFO:     connection open
_send_error_response: The string 'button-btn' is not a valid id; only letters, numbers, and underscore are permitted
INFO:     connection closed
```

Am I write, or the `input_task_button` should be used in a different way?

---

`shiny == 0.7.0`, `python == 3.10.13`

jcheng5:
Fixes #1107

The task button installs an input handler, which it uses to implement its auto-reset feature. One of the arguments to the input handler is the name/id of the input, a string. When a task button is in a module, that id string contains the module/namespace name as well.

Module-aware code in Shiny assumes regular strings are un-namespaced id's (akin to a relative path), and `ResolvedId` objects as namespaced id's (akin to an absolute path). In this case, we had a string that contained a namespaced id. That's then passed to `update_task_button`, which complains that it contains characters that shouldn't be part of an id.

This PR fixes the problem by changing the `input_handler` invoking code, changing the string id to a ResolvedId as early as possible.

### DIFF
--- a/shiny/input_handler.py
+++ b/shiny/input_handler.py
@@ -10,9 +10,10 @@ from typing import TYPE_CHECKING, Any, Callable, Dict
 if TYPE_CHECKING:
     from .session import Session
 
+from .module import ResolvedId
 from .types import ActionButtonValue
 
-InputHandlerType = Callable[[Any, str, "Session"], Any]
+InputHandlerType = Callable[[Any, ResolvedId, "Session"], Any]
 
 
 class _InputHandlers(Dict[str, InputHandlerType]):
@@ -31,7 +32,9 @@ class _InputHandlers(Dict[str, InputHandlerType]):
     def remove(self, type: str) -> None:
         del self[type]
 
-    def _process_value(self, type: str, value: Any, name: str, session: Session) -> Any:
+    def _process_value(
+        self, type: str, value: Any, name: ResolvedId, session: Session
+    ) -> Any:
         handler = self.get(type)
         if handler is None:
             raise ValueError("No input handler registered for type: " + type)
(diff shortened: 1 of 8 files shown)


shiny/session/_session.py
```diff
@@ -334,6 +335,7 @@ def verify_state(expected_state: ConnectionState) -> None:
                 ...
             except Exception as e:
                 try:
+                    traceback.print_exception(*sys.exc_info())
```
jcheng5, line 338: Without this fix, errors in the InputHandler stack kill the session without giving a traceback.

shiny/input_handler.py
```diff
@@ -94,27 +97,31 @@ def _(value, name, session):
...
+def _(
+    value: str | list[str], name: ResolvedId, session: Session
+) -> date | tuple[date, date]:
     if isinstance(value, str):
         return datetime.strptime(value, "%Y-%m-%d").date()
-    return tuple(  # pyright: ignore[reportReturnType]
-        datetime.strptime(v, "%Y-%m-%d").date() for v in value
+    return (
+        datetime.strptime(value[0], "%Y-%m-%d").date(),
+        datetime.strptime(value[1], "%Y-%m-%d").date(),
```
jcheng5, line 107: This slight rewrite means we don't need to use pyright pragmas. Same below.

shiny/session/_session.py
wch: If this `sys.exc_info` is needed only for Python 3.9 and below, can you add a comment saying that, and saying what it can be changed to in the future, when we drop 3.9 support?
https://docs.python.org/3/library/traceback.html#traceback.print_exception

wch (approved):

Issue #704, opened 2023-08-24
Passing additional options to input.select


iainwallacebms:
Hi,

Is it possible to pass additional options to input.select? The use case is to allow users to create additional items by passing something similar to R where we do _options = list(create = TRUE)_

I don't seen an example of the options argument from the [R documentation](https://shiny.posit.co/r/reference/shiny/latest/selectinput) in the [python documentation](https://shiny.posit.co/py/api/ui.input_selectize.html) but perhaps I am missing something obvious

         options
         A list of options. See the documentation of selectize.js(https://selectize.dev/docs/usage) for possible options (character 
        option values inside [base::I()](https://rdrr.io/r/base/AsIs.html) will be treated as literal JavaScript code; see 
        [renderDataTable()](https://shiny.posit.co/r/reference/shiny/latest/renderDataTable.html) for details).

thanks a million,

Iain

schloerke:
Duplicate of https://github.com/posit-dev/py-shiny/issues/158 . Closing here

PR #1121, opened 2024-02-12
Selected rows method

### LINKED ISSUE #1120: Add class-based rows selected method
We want to have a method for getting the selected rows from a data frame. This will help with discoverability over the current implementation. 

## Current 
ui.output_data_frame("my_df")
input.my_df_selected_rows()

## Future 
my_df.input_selected_rows()

## Possible future methods
my_df.update_selected_rows
my_df.input_data_selected()
my_df.update_cell()
my_df.update_row()
my_df.input_cell()

gshotwell:
Closes #1120


shiny/session/_session.py
```diff
@@ -608,22 +608,26 @@ def _send_remove_ui(self, selector: str, multiple: bool) -> None:
     @overload
     def _send_progress(
         self, type: Literal["binding"], message: BindingProgressMessage
-    ) -> None: ...
```
gshotwell, line 611: These were to resolve a flake8 error, #704

shiny/api-examples/data_frame/app-core.py
```diff
@@ -44,9 +44,9 @@ def summary_data():
 
     @reactive.calc
     def filtered_df():
-        # input.summary_data_selected_rows() is a tuple, so we must convert it to list,
+        # summary_data.selected_rows() is a tuple, so we must convert it to list,
         # as that's what Pandas requires for indexing.
-        selected_idx = list(req(input.summary_data_selected_rows()))
```
gshotwell, line 49: One issue here is that this does introduce some inconsistency with the plot brushing options, maybe we should hold off changing the examples until we wrap those functions in methods.

shiny/render/_dataframe.py
```diff
@@ -238,7 +248,7 @@ class data_frame(Renderer[DataFrameResult]):
     Row selection
     -------------
     When using the row selection feature, you can access the selected rows by using the
```
schloerke, line 250: (Couldn't apply single suggestion. Split into two parts)

```suggestion
    When using the row selection feature, you can access the selected rows by using the
    `<data_frame_renderer>.input_selected_rows()` method, where `<data_frame_renderer>` is the render function name that corresponds with the `id=` used in :func:`~shiny.ui.outout_data_frame`. Internally, this method retrieves the selected row value from session's `input.<id>_selected_rows()` value. The value returned will be `None` if no rows
```

- [ ] Please reformat text block within IDE

schloerke (changes requested):
- [x] Get `make pyright` to pass

shiny/session/_session.py
schloerke: IDK about the GitHub issue link... but the code change is good!

shiny/api-examples/data_frame/app-core.py
gshotwell: Just that for plot brushing we still have the `input.<id>_brushed_points` rather than using the object. I'm probably overthinking this.

schloerke: Yes, we haven't implemented the plot brushing yet. 

I'm happy that we're exploring this approach and the changes are good.

shiny/api-examples/data_frame/app-core.py
```diff
@@ -44,9 +44,11 @@ def summary_data():
 
     @reactive.calc
     def filtered_df():
-        # input.summary_data_selected_rows() is a tuple, so we must convert it to list,
+        req(summary_data.input_selected_rows())
+
+        # summary_data.selected_rows() is a tuple, so we must convert it to list,
         # as that's what Pandas requires for indexing.
-        selected_idx = list(req(input.summary_data_selected_rows()))
+        selected_idx = list(req(summary_data.input_selected_rows()))
```
schloerke, line 51: `req()` is done above. Removing here

```suggestion
        selected_idx = list(summary_data.input_selected_rows())
```

shiny/api-examples/data_frame/app-express.py
```diff
@@ -66,9 +66,10 @@ def country_detail_percap():
 
 @reactive.calc
 def filtered_df():
-    # input.summary_data_selected_rows() is a tuple, so we must convert it to list,
+    # summary_data.input_selected_rows() is a tuple, so we must convert it to list,
```
schloerke, line 69:
```suggestion
    req(summary_data.input_selected_rows())

    # summary_data.input_selected_rows() is a tuple, so we must convert it to list,
```

shiny/api-examples/data_frame/app-express.py
```diff
@@ -66,9 +66,10 @@ def country_detail_percap():
...
     # as that's what Pandas requires for indexing.
-    selected_idx = list(req(input.summary_data_selected_rows()))
+
+    selected_idx = list(req(summary_data.input_selected_rows()))
```
schloerke, line 72:
```suggestion
    selected_idx = list(summary_data.input_selected_rows())
```

examples/dataframe/app.py
```diff
@@ -93,12 +93,12 @@ def handle_edit():
     @render.text
     def detail():
         if (
```
schloerke, line 95: I like how you had it below in the test app. Copying it here.

```suggestion
        selected_rows = grid.input_selected_rows() or ()
        if len(selected_rows) > 0:
            # "split", "records", "index", "columns", "values", "table"
            return df.iloc[list(selected_rows)]
```

examples/dataframe/app.py
```diff
@@ -93,12 +93,12 @@ def handle_edit():
...
-            input.grid_selected_rows() is not None
-            and len(input.grid_selected_rows()) > 0
+            grid.input_selected_rows() is not None
```
schloerke, line 96:
```suggestion
```

examples/dataframe/app.py
```diff
@@ -93,12 +93,12 @@ def handle_edit():
...
+            and len(grid.input_selected_rows()) > 0
         ):
             # "split", "records", "index", "columns", "values", "table"
```
schloerke, line 100:
```suggestion
```

examples/dataframe/app.py
```diff
@@ -93,12 +93,12 @@ def handle_edit():
...
 
-            return df().iloc[list(input.grid_selected_rows())]
+            return df().iloc[list(grid.input_selected_rows())]
```
schloerke, line 101: Please, remove this line as the top comment as the suggested code. 

(Could not suggest a full change in a single comment. ...weird

schloerke (approved):

shiny/render/_dataframe.py
```diff
@@ -270,6 +279,15 @@ async def transform(self, value: DataFrameResult) -> Jsonifiable:
             )
         return value.to_payload()
 
+    def input_selected_rows(self) -> Optional[tuple[int]]:
```
wch, line 282: I think it would be better if this always returned a tuple. Especially when the selection mode is multiple, having zero rows selected is a different set of selected rows (it's a selection of zero rows), but it is not a fundamentally different kind (like `None`).

Returning an empty tuple would be similar to `brushed_points`, which, if no points are selected, returns an empty data frame instead of `None`.

shiny/session/_session.py
wch: Just FYI, after #1125, it's not necessary to make these changes. That PR updated the flake8 configuration as recommended by Black, which allows this formatting.

https://github.com/posit-dev/py-shiny/pull/1125/files#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52

shiny/render/_dataframe.py
wch: Oh, actually, after testing it out, I see that:
- If the mode is single or multiple and no rows are selected, this returns an empty tuple.
- If the mode is none, then this returns `None`.

It would help if the docs accurately reflected this.

Also, a small nit: 

```suggestion
    def input_selected_rows(self) -> tuple[int] | None:
```

Although this is technically equivalent to using `Optional`, we've tried to use `Optional` to convey that an argument is optional, which is different from this case.

wch:
A couple module-related questions:

- Does this work in a module?
- Does it work if there's a module which returns the `renderer` object? For example, suppose the module is something like this, and then the server function that instantiates the module wants to get the selected rows:
    ```py
    @module.server
    def mod_server(input: Inputs, output: Outputs, session: Session): 
        @render.data_frame
        def grid():
            ...

        return grid.selected_rows
    ```

PR #1126, opened 2024-02-14
Update `@render.data_frame` renderer to work properly with fill/fillable

### LINKED ISSUE #1118: Modify data grid to work well with fill/fillable
Currently dataframes do not look particularly good when they're included in a filling layout. Some things we want to solve:

- Data grids should properly fill their parent
- When included in  a card it would be nice if the pagination and footers were pushed to the bottom of the card.

cpsievert:
Closes #1118


js/dataframe/index.tsx
```diff
@@ -476,6 +484,8 @@ export class ShinyDataFrameOutput extends HTMLElement {
 
     const myDiv = document.createElement("div");
     myDiv.classList.add("html-fill-container", "html-fill-item");
+    myDiv.style.marginLeft = "auto";
+    myDiv.style.marginRight = "auto";
```
jcheng5, line 488: I don't love the centering. Do we do that anywhere else?

jcheng5:
Overall this looks great! Only regressing case I could find is `height="auto"` in a non-fill context. There's a ton of extra space now that wasn't there before.

```python
import seaborn as sns

df = sns.load_dataset("iris").iloc[:5,]

@render.data_frame
def data_frame():
    return render.DataGrid(
        df,
        height="auto",
    )

with ui.p():
    "Hello"
```

jcheng5:
I also noticed that even on main, `output_data_frame` always makes the outermost element fill/fillable, which I don't think is correct if `height="auto"` is passed. Like if you take the example from my previous comment, and wrap all the UI in a `ui.card(height="800px")`.

cpsievert:
> I don't love the centering. Do we do that anywhere else?

Agreed (and I don't think so) in general, but it's maybe worth doing inside a card? [87daed7](https://github.com/posit-dev/py-shiny/pull/1126/commits/87daed7054ede31aa306a4afa270dc8d21c30dde) restricts it to just a `.card-body` context, but I can probably be talked into not doing that.

>  There's a ton of extra space now that wasn't there before.

Oops, that's cause of the `flex-basis` I added last minute without thinking it through fully. [87daed7](https://github.com/posit-dev/py-shiny/pull/1126/commits/87daed7054ede31aa306a4afa270dc8d21c30dde) reverts that.

> I also noticed that even on main, output_data_frame always makes the outermost element fill/fillable

I don't think this is a problem, at least for the situation you're thinking of. It could be a problem though if the children of that element are for some reason not compatible with a `display:flex; flex-direction:column` layout.

Just to illustrate, here's the example you have in mind, with the result (which seems right):

```r
import seaborn as sns

df = sns.load_dataset("iris").iloc[:5,]

with ui.card(height="400px"):
    @render.data_frame
    def data_frame():
        return render.DataGrid(df, height="auto")

with ui.p():
    "Hello"
```

Screenshot 2024-02-23 at 3 06 29 PM

js/dataframe/index.tsx
cpsievert: See https://github.com/posit-dev/py-shiny/pull/1126#issuecomment-1961994819

jcheng5:
I actually meant this:

```python
import seaborn as sns

df = sns.load_dataset("iris").iloc[:5,]

with ui.card(height="400px"):
    @render.data_frame
    def data_frame():
        return render.DataGrid(df, height="auto")

    with ui.p():
        "Hello"
```

image

If you drop the fill classes from the `ui.output_data_frame`'s outermost element, you get this instead (which I think is better):

image

jcheng5:
We can merge without that fix though, if you don't think there's anything we can do about it right now.

cpsievert:
Ah, good point. [789147c](https://github.com/posit-dev/py-shiny/pull/1126/commits/789147c13d815bbe8a31fb589ba1850ee982d45c) addresses that.

cpsievert:
@gregswinehart "loves it" :)

jcheng5:
OK, I'm happy to be overruled!

PR #1139, opened 2024-02-16
Don't error on malformed date input

### LINKED ISSUE #1138: Date selector component crashes shiny process on invalid input
This issue is replicable on the component gallery in the docs: 

https://shiny.posit.co/py/components/inputs/date-range-selector/

Rather than selecting a date through the calendar widget, if you manually clear the date from the text input (with the intention of typing in a valid date), the Python process will attempt to parse the empty string as a date and emit an error which breaks the app: 

Screenshot 2024-02-16 at 9 40 11 AM

wch:
This fixes #1138.

This returns `None` for each malformed date component. If it's a date range, then it returns a length-2 tuple where either or both components could be `None`. I thought this behavior seemed reasonable, but I'm open to other ideas. Another possibility is to simply not change and invalidate `input.x()` until there's a valid value, but I think that fix would be a bit more involved.

Update: I've thought about this a bit more, and I think the behavior in this PR is what we should go with -- the UI component can have an empty value, and I think it makes sense to reflect that in the server-side input value (as opposed to not changing the input value until there's a valid value in the UI).


jcheng5:
LGTM

PR #1142, opened 2024-02-20
chore: Remove deprecated datetime method calls

schloerke:
> The method "utcfromtimestamp" in class "datetime" is deprecated
>    Use timezone-aware objects to represent datetimes in UTC; e.g. by calling .fromtimestamp(datetime.UTC) (reportDeprecated)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Other changes
 
 * Replaced use of `sys.stderr.write()` with `print(file=sys.stderr)`, because on some platforms `sys.stderr` can be `None`. (#1131)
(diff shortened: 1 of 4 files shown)


wch:
I think the changelog should mention that this is a potentially breaking change, since the returned object is different, and the string formatting is different.


```py
now = datetime.utcnow().timestamp()

y = datetime.utcfromtimestamp(now)
y
#> datetime.datetime(2024, 2, 21, 5, 37, 58, 761000)
print(y)
#> 2024-02-21 05:37:58.761000

x = datetime.fromtimestamp(now, timezone.utc)
x
#> datetime.datetime(2024, 2, 21, 5, 37, 58, 761000, tzinfo=datetime.timezone.utc)
print(x)
# 2024-02-21 05:37:58.761000+00:00
```

PR #1158, opened 2024-02-27
test: Consolidate playwright nav component test

joesho112358:
Updated tests/playwright/shiny/components/nav/test_nav.py to loop and soft assert

### DIFF
--- a/tests/playwright/shiny/components/nav/test_nav.py
+++ b/tests/playwright/shiny/components/nav/test_nav.py
@@ -55,14 +55,14 @@ def test_nav(page: Page, local_app: ShinyAppProc) -> None:
         }
     }
 
-    def tester(tab_name):
-        tab = nav_data[tab_name]['control'](page, tab_name)
-        tab.expect_nav_values(["a", "b", "c"])
-        tab.expect_value("a")
-        tab.expect_content(nav_data[tab_name]['verify'] + ": tab a content")
-        tab.set("b")
-        tab.expect_value("b")
-        tab.expect_content(nav_data[tab_name]['verify'] + ": tab b content")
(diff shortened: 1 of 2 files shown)


tests/playwright/shiny/components/nav/test_nav.py
```diff
@@ -20,75 +20,60 @@ def test_nav(page: Page, local_app: ShinyAppProc) -> None:
...
-    navset_underline.set("b")
-    navset_underline.expect_value("b")
-    navset_underline.expect_content("navset_underline(): tab b content")
+    nav_list = [
+        "navset_tab", "navset_pill", "navset_underline", "navset_card_tab",
+        "navset_card_pill", "navset_card_underline", "navset_pill_list", "page_navbar"
+    ]
+    failures = []
+    # soft assert
+    for nav in nav_list:
```
schloerke, line 73: Use data already defined

```suggestion
    failures = []
    # soft assert
    for nav in list(nav_data.keys()):
```

schloerke:
@joesho112358 Thank you for your PR! Not every day someone is in the weeds of testing.

I appreciate the idea of soft asserting, but by only presenting the error at the end, playwright's test stack debugging is very minimal and difficult to read.  So given we need all the tests to pass, I'm ok with the hard assertion knowing if the first control fails, the others will not be tested. (We could parameterize the test, but I feel it would be unnecessary server startup time added to the testing suite.)

`py-shiny` also needs to have full type support. Since the `verify` field and name were very similar, I combined the info and put it and the control into a `dataclass`. This allows for easier typing without double record keeping the controls too much. (I tried using a common base class, but the common base class' init method was not consistent with the subclass.)

PR #1164, opened 2024-02-28
Update examples with lower-case `calc`, `effect`, and `value`

### LINKED ISSUE #823: Update examples to use lowercase `reactive.{Calc,Effect,Value}`
From #822 
> This PR changes those names to lowercase, and adds aliases so that the old capitalized names still work.
> 
> This PR does not change any examples because if users on a previous version of shiny were to use the example code, it wouldn't work for them, and it would be unclear to them why. In a future release, we should update the examples to use the new lowercase names.

wch:
Closes #823. This PR updates examples to use lower-case `calc`, `effect`, and `value`.

There was some discussion in #822 whether we should use lower-case `value` for the examples. I went ahead and did it here, in a separate commit, but it can be easily reversed.


jcheng5:
LGTM

PR #1165, opened 2024-02-29
fix(playwright): Add playwright workaround within `Accordion`'s `.expect_open()` and `.expect_panels()`

joesho112358:
The goal of this is to remove the early return from the accordion test so that more functionality is tested. Looks like an infinite loop may have been being introduced in the code that is commented. I am not sure what the intention of that block was, and can remove if not needed instead of commenting it.

The specific infinite loop in playwright:
https://github.com/microsoft/playwright-python/blob/main/playwright/_impl/_sync_base.py#L110


tests/playwright/controls.py
```diff
@@ -13,7 +13,7 @@
 from playwright.sync_api import FilePayload, FloatRect, Locator, Page, Position
 from playwright.sync_api import expect as playwright_expect
 
-# Import `shiny`'s typing extentions.
+# Import `shiny`'s typing extensions.
```
schloerke, line 16: 🤦 Thank you

tests/playwright/controls.py
```diff
@@ -1090,31 +1090,29 @@ def expect_locator_values_in_list(
...
+            # has_locator = has_locator.locator(
+            #     f"xpath=self::*[{_xpath_match_str(key, item)}]"
+            # )
+            #
+            # # Given the container, make sure it contains this locator
+            # loc_container = loc_container.locator(
+            #     # Return self
+            #     "xpath=.",
+            #     has=has_locator,
+            # )
```
schloerke, line 1106: The goal of `expect_locator_values_in_list` is to make sure that the attr values found (e.g. `data-value`) on matching elements are the **only** values found. Order matters. 

Ideally, playwright would have an `expect(loc).to_have_attribute(key, array_of_values)` support, but they do not. 😞 

I'd like to keep the expectation (below in the try/except) as a single expectation, rather than a loop of expectations (which would be MUCH easier to reason about).  (If a failure occurs, we do loop over the different possibilities in an effort to debug the situation. Time does move forward while the expectation loop churns, allowing Shiny being able to update the accordion in between tests is not ideal.)

----------------

So unfortunately, I'd like to keep this locator creation code as it was. Thank you!

schloerke (changes requested):

schloerke:
@joesho112358 Any idea on how to stop the inf loop? (While keeping the locator creation as it is)

tests/playwright/controls.py
joesho112358: Hi, I am not sure I fully understand the reason to not loop through with the built in `to_have_attribute(key, value)`. Could you elaborate on that?

schloerke: Time does move forward in the DOM when performing multiple expectations in a loop. 

It is possible for Shiny to update the items in the list in question while in the middle of the for loop of expectations. So in the first have it is listA and in the back half it is listB. This makes debugging this rare situation really hard. 

Also detecting items that are _not_ in the list becomes difficult. I wouldn't know how to check for it as I wouldn't be able to guarantee that the K items found are the original set of K we're looking for. 

--------------

To get around this difficulty, I use a loop to construct a very complicated locator. If this locator exists, then we can guarantee that the items in the list are the only items and their order is preserved. 


Pseudo code for an `expect(els_loc).to_have_attribute(attrs)`:
- for each el, make sure the corresponding attr matches.
- find all of the matching els; Verify the count is the expected count

If we can do this in a single expectation, then the expectation is playwrite-like in that it can be repeated until the expectation is satisfied.

If we can create another single expectation that achieves the same logic and doesn't block forever in the accordion test, I'm all for it!

joesho112358: I see, thanks for the explanation. Will think about it!

joesho112358:
@schloerke Hi, ready for another look. I had the thought that if we want everything all at once, another option is to construct 1 xpath that can handle it. Thoughts? Will still need to check some styling stuff I think though if you think this is an acceptable path.

joesho112358:
Closing this PR as I have not made progress and do not see a clear path forward

joesho112358:
@schloerke hi again, this was bugging me a bit over the past few months. sorry if i repeat what was above here, but i have determined it is not the selector that is the issue. for some reason `to_have_count` goes into an infinite loop with this selector... so i worked around it because the `count` does not have that issue. it's a bit dirty, so if we need to close this again, then no worries!

shiny/playwright/controller/_expect.py
```diff
@@ -218,9 +221,13 @@ def expect_locator_values_in_list(
     # and all elements all unique, then it should have a count of `len(arr)`
     loc_inputs = loc_container.locator(loc_item)
     try:
-        playwright_expect(loc_inputs).to_have_count(len(arr), timeout=timeout)
+        if kwargs.get("alt_verify"):
+            sleep(1)  # to make up for not using to_have_count
+            assert loc_inputs.count() == len(arr)
```
schloerke, line 226: Would this work if we tried to check for any text values? Thinking that `.to_have_text()` works on an array of elements. Maybe it is jus the `.to_have_count()` _count_ part that is bad. 

```suggestion
            anything = re.compile(r".*")
            any_text_values = [anything for _ in range(len(arr))]
            playwright_expect(loc_inputs).to_have_text(any_text_values, timeout = timeout)
```

shiny/playwright/controller/_expect.py
```diff
@@ -146,6 +148,7 @@ def expect_locator_values_in_list(
     is_checked: bool | MISSING_TYPE = MISSING,
     timeout: Timeout = None,
     key: str = "value",
+    **kwargs,
```
schloerke, line 151:
```suggestion
    alt_verify: bool = False,
```

shiny/playwright/controller/_expect.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from time import sleep
+
```
schloerke, line 4:
```suggestion
```

shiny/playwright/controller/_accordion.py
```diff
@@ -281,6 +276,7 @@ def expect_panels(
             arr=value,
             key="data-value",
             timeout=timeout,
+            **kwargs,
```
schloerke, line 279:
```suggestion
            alt_verify=True,
```

shiny/playwright/controller/_accordion.py
```diff
@@ -255,13 +252,11 @@ def expect_open(
...
```
schloerke, line 255:
```suggestion
            alt_verify=True,
```

shiny/playwright/controller/_accordion.py
```diff
@@ -238,10 +238,7 @@ def expect_width(self, value: StyleValue, *, timeout: Timeout = None) -> None:
         _expect_style_to_have_value(self.loc_container, "width", value, timeout=timeout)
 
     def expect_open(
-        self,
-        value: list[PatternOrStr],
-        *,
-        timeout: Timeout = None,
+        self, value: list[PatternOrStr], *, timeout: Timeout = None, **kwargs
```
schloerke, line 241: Do not expose the parameter publicly

```suggestion
        self,
        value: list[PatternOrStr],
        *,
        timeout: Timeout = None,
```

tests/playwright/shiny/components/accordion/test_accordion.py
```diff
@@ -92,7 +89,8 @@ def test_accordion(page: Page, local_app: ShinyAppProc) -> None:
             "Section E",
             "Section F",
             "Section G",
-        ]
+        ],
+        alt_verify=True,
```
schloerke, line 93:
```suggestion
```

tests/playwright/shiny/components/accordion/test_accordion.py
```diff
@@ -102,7 +100,8 @@ def test_accordion(page: Page, local_app: ShinyAppProc) -> None:
...
```
schloerke, line 104:
```suggestion
```

shiny/playwright/controller/_expect.py
joesho112358: unfortunately not. `.to_have_text()` is yielding the same results as the `.to_have_count()`

schloerke: Damn

shiny/playwright/controller/_expect.py
```diff
@@ -171,6 +172,11 @@ def expect_locator_values_in_list(
         The timeout for the expectation. Defaults to `None`.
     key
         The key. Defaults to `"value"`.
+    alt_verify
+        Determines if multiple expectations should be performed.
+        Defaults to `False`, a single (and very complicated) locator is asserting.
```
schloerke, line 177:
```suggestion
        Defaults to `False`, a single (and very complicated) locator is asserted.
```

shiny/playwright/controller/_expect.py
```diff
@@ -171,6 +172,11 @@ def expect_locator_values_in_list(
...
+        `True` will perform multiple assersions, which have the possibility of being invalid.
```
schloerke, line 178:
```suggestion
        `True` will perform multiple assertions, which have the possibility of being invalid.
```

shiny/playwright/controller/_expect.py
```diff
@@ -195,6 +201,28 @@ def expect_locator_values_in_list(
...
+        for item, i in zip(arr, range(len(arr))):
+            # Expecting item `{i}` to be `{item}`
+            playwright_expect(
+                loc_container_orig.locator(loc_item).nth(i)
+            ).to_have_attribute(key, item, timeout=timeout)
+        return
+
+    if alt_verify:
+        # Accordion has issues where the single locator assertion waits forever within playwright.
+        # Perform multiple assertions until playwright fixes bug.
```
schloerke, line 222:
```suggestion
        # Perform multiple assertions until playwright fixes bug.
```

shiny/playwright/controller/_expect.py
```diff
@@ -220,21 +248,8 @@ def expect_locator_values_in_list(
...
-
-        for item, i in zip(arr, range(len(arr))):
-            # Expecting item `{i}` to be `{item}`
-            playwright_expect(
-                loc_container_orig.locator(loc_item).nth(i)
-            ).to_have_attribute(key, item, timeout=timeout)
-
+        # Debug expectations
+        perform_multiple_assertions()
+
```
schloerke, line 253:
```suggestion

```

schloerke:
@joesho112358 We've given it our fairest shake to resolve the issues and we can not. 

py-shiny does use multiple assertions in other pieces of code. While not ideal, it isn't _too_ far of a stretch to perform many assertions here as well. I'd rather have _ok_ coverage than missing coverage.

Thank you for resurrecting this!

schloerke (approved):

joesho112358:
@schloerke sorry the resurrection was so unrefined! I probably should have put it as a draft

PR #1167, opened 2024-03-01
Make autoreload work on codespaces

jcheng5:
This PR makes Shiny detect that it's running on Codespaces, where the browser can't just point to `http://127.0.0.1:{autoreload-port}` and expect to succeed. Instead, a proxy URL is used. See [Forwarding ports in your codespace](https://docs.github.com/en/codespaces/developing-in-a-codespace/forwarding-ports-in-your-codespace).

Note that this alone is not enough to make autoreload work quite right on codespaces. The Shiny app will try to connect to the autoreload websocket server and fail. This is because Codespace's forwarded ports are private by default, and need you to access each proxy hostname via HTTP document load first so it can set some cookies and whatnot. A forthcoming PR for pyshiny-vscode will try to address this.

**Update:** Here's the corresponding pyshiny-vscode PR https://github.com/posit-dev/pyshiny-vscode/pull/27

### DIFF
--- a/shiny/_autoreload.py
+++ b/shiny/_autoreload.py
@@ -200,6 +200,11 @@ async def _coro_main(
     ) -> Optional[tuple[http.HTTPStatus, websockets.datastructures.HeadersLike, bytes]]:
         # If there's no Upgrade header, it's not a WebSocket request.
         if request_headers.get("Upgrade") is None:
+            # For some unknown reason, this fixes a tendency on GitHub Codespaces to
+            # correctly proxy through this request, but give a 404 when the redirect is
+            # followed and app_url is requested. With the sleep, both requests tend to
+            # succeed reliably.
+            await asyncio.sleep(1)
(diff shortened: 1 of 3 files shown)


shiny/_autoreload.py
```diff
@@ -200,6 +200,11 @@ async def process_request(
     ) -> Optional[tuple[http.HTTPStatus, websockets.datastructures.HeadersLike, bytes]]:
         # If there's no Upgrade header, it's not a WebSocket request.
         if request_headers.get("Upgrade") is None:
+            # For some unknown reason, this fixes a tendency on GitHub Codespaces to
+            # correctly proxy through this request, but give a 404 when the redirect is
+            # followed and app_url is requested. With the sleep, both requests tend to
+            # succeed reliably.
+            await asyncio.sleep(1)
```
wch, line 207: I just want to check: it doesn't take this code path when developing locally, right?

PR #1168, opened 2024-03-01
Add python 3.12 to test matrix

cpsievert:
3.12 was officially release a few months ago, so we should be testing against it https://www.python.org/downloads/


wch:
It looks like there are test failures here due to flake8 wanting different formatting in f-strings, and I believe black doesn't do reformat inside of f-strings. I don't know why flake8 only cares about that in Python 3.12 though.

PR #1172, opened 2024-03-01
Automatically load `globals.py` for Express apps

### LINKED ISSUE #1079: reactive.poll doesn't poll when defined in Express-adjacent modules

wch:
Closes #1079.

For Shiny Express apps, this PR makes it so that if there is a `globals.py` file in the same directory as the app file, it will be loaded with the session context set to `None`.

Note that even if the user does not have an `import globals` statement in their app file, this will still run the code in `globals.py`. If they want to actually access the variables defined in that file, they will have to an import, as in `import globals`. The part of this that may be surprising is that the code in globals.py could execute even though it's not being referred to anywhere in the app code.

@jcheng5 It feels weird to me that `globals.py` would be executed simply because it's in the same directory as the app file, and that behavior could be surprising for users. Like, Imagine that `globals.py` sets up a database connection and/or writes something to a log file, and the user removes or comments out `import globals` from their app. If it still keeps establishing the database connection and writing to a log file, that would be really surprising and would be difficult to diagnose and debug if you didn't know that how Shiny Express always loads that file.

Another way we could get the same effect is to walk the AST and look for `import globals` or `from globals import ...` and if it finds that, rewrite the AST to:

```
with session_context(None):
    import globals
```

Or, instead of rewriting the AST, during evaluation time we could simply set the session context to `None` before evaluating the import line.

*****

With this change, the example app from #1079 works, after renaming `data` to `globals`:

<details>
<summary>Full source code</summary>

### app.py

```python
from shiny import reactive

from globals import stock_quotes, SYMBOLS

ui.markdown(
    """
    # `shiny.reactive.poll` demo

    This example app shows how to stream results from a database (in this
    case, an in-memory sqlite3) with the help of `shiny.reactive.poll`.
    """
)

ui.input_selectize("symbols", "Filter by symbol", [""] + SYMBOLS, multiple=True)


def filtered_quotes():
    df = stock_quotes()
    if input.symbols():
        df = df[df["symbol"].isin(input.symbols())]
    return df

@render.express
def table():
    ui.HTML(
        filtered_quotes().to_html(
            index=False, classes="table font-monospace w-auto"
        )
    )
```

### globals.py
```python
import asyncio
import random
import sqlite3
from datetime import datetime
from typing import Any, Awaitable

import pandas as pd

from shiny import reactive

SYMBOLS = ["AAA", "BBB", "CCC", "DDD", "EEE", "FFF"]


def timestamp() -> str:
    return datetime.now().strftime("%x %X")


def rand_price() -> float:
    return round(random.random() * 250, 2)


# === Initialize the database =========================================


def init_db(con: sqlite3.Connection) -> None:
    cur = con.cursor()
    try:
        cur.executescript(
            """
            CREATE TABLE stock_quotes (timestamp text, symbol text, price real);
            CREATE INDEX idx_timestamp ON stock_quotes (timestamp);
            """
        )
        cur.executemany(
            "INSERT INTO stock_quotes (timestamp, symbol, price) VALUES (?, ?, ?)",
            [(timestamp(), symbol, rand_price()) for symbol in SYMBOLS],
        )
        con.commit()
    finally:
        cur.close()


conn = sqlite3.connect(":memory:")
init_db(conn)


# === Randomly update the database with an asyncio.task ==============


def update_db(con: sqlite3.Connection) -> None:
    """Update a single stock price entry at random"""

    cur = con.cursor()
    try:
        sym = SYMBOLS[random.randint(0, len(SYMBOLS) - 1)]
        print(f"Updating {sym}")
        cur.execute(
            "UPDATE stock_quotes SET timestamp = ?, price = ? WHERE symbol = ?",
            (timestamp(), rand_price(), sym),
        )
        con.commit()
    finally:
        cur.close()


async def update_db_task(con: sqlite3.Connection) -> Awaitable[None]:
    """Task that alternates between sleeping and updating prices"""
    while True:
        await asyncio.sleep(random.random() * 1.5)
        update_db(con)


asyncio.create_task(update_db_task(conn))


# === Create the reactive.poll object ===============================


def tbl_last_modified() -> Any:
    print("polling")
    df = pd.read_sql_query("SELECT MAX(timestamp) AS timestamp FROM stock_quotes", conn)
    return df["timestamp"].to_list()


@reactive.poll(tbl_last_modified, 0.5)
def stock_quotes() -> pd.DataFrame:
    return pd.read_sql_query("SELECT timestamp, symbol, price FROM stock_quotes", conn)
```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * For Shiny Express apps, added `express.app_opts()`, which allows setting application-level options, like `static_assets` and `debug`. (#1170)
 
+* Closed #1079: For Shiny Express apps, automatically run a `globals.py` file in the same directory as the app file, if it exists. The code in `globals.py` will be run with the session context set to `None`. (#1172)
+
 ### Other changes
 
 * `@render.data_frame` now properly fills its container by default. (#1126)
--- a/shiny/_utils.py
+++ b/shiny/_utils.py
@@ -12,7 +12,9 @@ import secrets
 import socketserver
 import sys
 import tempfile
+import warnings
 from pathlib import Path
+from types import ModuleType
 from typing import Any, Awaitable, Callable, Generator, Optional, TypeVar, cast
 
 from ._typing_extensions import ParamSpec, TypeGuard
@@ -541,6 +543,13 @@ def package_dir(package: str) -> str:
         return os.path.dirname(pkg_file)
 
 
+class ModuleImportWarning(ImportWarning):
+    pass
+
+
+warnings.simplefilter("always", ModuleImportWarning)
+
+
 def import_module_from_path(module_name: str, path: Path):
     import importlib.util
 
@@ -552,6 +561,25 @@ def import_module_from_path(module_name: str, path: Path):
         raise ImportError(f"Could not import module {module_name} from path: {path}")
 
     module = importlib.util.module_from_spec(spec)
+
+    prev_module: ModuleType | None = None
+
+    if module_name in sys.modules:
+        prev_module = sys.modules[module_name]
+        warnings.warn(
+            f"A module named {module_name} is already loaded, but is being loaded again.",
+            ModuleImportWarning,
+            stacklevel=1,
+        )
+
     sys.modules[module_name] = module
-    spec.loader.exec_module(module)
+
+    try:
+        spec.loader.exec_module(module)
+    except Exception:
+        if prev_module is None:
+            del sys.modules[module_name]
+        else:
+            sys.modules[module_name] = prev_module
+        raise
     return module
--- a/shiny/_utils.py
+++ b/shiny/_utils.py
@@ -544,6 +544,9 @@ def package_dir(package: str) -> str:
 def import_module_from_path(module_name: str, path: Path):
     import importlib.util
 
+    if not path.is_absolute():
+        raise ValueError("Path must be absolute")
+
     spec = importlib.util.spec_from_file_location(module_name, path)
     if spec is None or spec.loader is None:
         raise ImportError(f"Could not import module {module_name} from path: {path}")
--- a/shiny/_utils.py
+++ b/shiny/_utils.py
@@ -10,7 +10,9 @@ import os
 import random
 import secrets
 import socketserver
+import sys
 import tempfile
+from pathlib import Path
 from typing import Any, Awaitable, Callable, Generator, Optional, TypeVar, cast
 
 from ._typing_extensions import ParamSpec, TypeGuard
@@ -537,3 +539,16 @@ def package_dir(package: str) -> str:
         if pkg_file is None:
             raise RuntimeError(f"Could not find package dir for '{package}'")
         return os.path.dirname(pkg_file)
+
+
+def import_module_from_path(module_name: str, path: Path):
+    import importlib.util
+
+    spec = importlib.util.spec_from_file_location(module_name, path)
+    if spec is None or spec.loader is None:
+        raise ImportError(f"Could not import module {module_name} from path: {path}")
+
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[module_name] = module
+    spec.loader.exec_module(module)
+    return module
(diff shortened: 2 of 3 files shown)


jcheng5:
In both cases, "globals" has special meaning--you'd have to be told that in the documentation. I far prefer "Shiny Express looks for globals.py and loads it before the app starts, outside of any session." instead of "Shiny Express detects imports of globals.py directly from the app.py file (not from any modules imported by app.py) and treats it differently so that objects inside don't belong to the current session."

jcheng5:
By the way, pytest has conftest.py that's imported before tests run. It's almost exactly analogous to this situation. I'm not saying I love it, but I far far prefer it to the rewriting, which even after it's explained is still surprising IMHO.

wch:
Hm, OK, those things make sense, and I can get on board with this approach.

PR #1175, opened 2024-03-03
test: Do not install shinylive

schloerke:
Without changes from #1013 , I believe the type checking should be failing.  (I'd be happy to be wrong!)


schloerke:
After running `make clean` locally, I am not able to reprex the pyright errors associated with starlette 0.37.1. 

@wch We might be ok closing this PR and #1013 ?

schloerke:
Typings work now. I see what you did in #1013 . Thank you.

Closing this PR as it is not needed.

PR #1192, opened 2024-03-05
Fix `express.ui.update_date` and`express.ui.update_text` links

wch:
For the Express API docs, the `update_date` and `update_text` items point to the non-Express functions:

image

This PR fixes them, and it also simplifies the config file by removing `dynamic: true`, which I believe is no longer needed.


gadenbuie (approved):
Looks good to me, thanks!

PR #1197, opened 2024-03-06
test: Speed up tests by using single page. Do not attempt deploys job unless conditions are valid

schloerke:
* Also mark resolve_id test as flaky to make sure it's not a random error.
* Always save trace/video when debugging playwright.


.github/workflows/pytest.yaml
```diff
@@ -118,9 +118,35 @@ jobs:
...
+        uses: ./.github/py-shiny/setup
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Test that deployable example apps work
+        timeout-minutes: 5 # ~10s locally
+        env:
+          DEPLOY_APPS: "false"
+        run: |
+          make playwright-deploys SUB_FILE=". -vv"
```
karangattu, line 143: Add the upload artifact step here

.github/workflows/pytest.yaml
```diff
@@ -118,9 +118,35 @@ jobs:
...
+
+  playwright-deploys:
+    needs: [playwright-deploys-precheck]
+    if: github.event_name != 'release' && (github.event_name == 'push' || (github.event_name == 'pull_request' && startsWith(github.head_ref, 'deploy')))
```
karangattu, line 147: Remove duplicate if logic in deploys step below

PR #1203, opened 2024-03-07
Force UTF-8 for Shiny Express on Windows

jcheng5:


wch:
Should check if there are other places we call `open(file)` to read in Python code.

wch (approved):

PR #1198, opened 2024-03-06
feat: Editable data frame

schloerke:
Related: #1135 
Related: #1119 

* Fixes followup to #1120 `mydf.input_selected_rows()`
* Fixes #1116 ; `mydf.update_selected_rows()`
* Adds feature of getting selected data from dataframe renderer. ; `mydf.data_selected_rows()`
* Adds feature of getting data from dataframe renderer. ; `mydf.data()`
* Adds feature of getting edited patches from dataframe renderer. ; `mydf.cell_patches()`

PR #1208, opened 2024-03-08
Use TypeScript strict mode

wch:
This enables TypeScript's strict mode, and fixes some issues that it turned up.

I did not fix issues with the dataframe component, because there are many changes in #1198 that could conflict. I set the tsconfig.json file to exclude the `dataframe/` directory so that this won't cause problems with CI. After this is merged to `main`, then `main` can be merged into the branch for #1198, and that exclusion can be removed there.


wch:
I just pushed some more changes to tsconfig.json which were informed by this article: https://www.totaltypescript.com/tsconfig-cheat-sheet

PR #1211, opened 2024-03-11
chore: Use `uv` and `ruff` for local package installation and linting/formatting

schloerke:
Related:
* https://github.com/astral-sh/uv
* https://docs.astral.sh/ruff/
* https://docs.astral.sh/ruff/rules/


.pre-commit-config.yaml
```diff
@@ -1,6 +1,7 @@
 repos:
-  - repo: https://github.com/psf/black
-    rev: "24.2.0"
+  - repo: https://github.com/astral-sh/ruff-pre-commit
```
wch, line 2: Check that this does not block the user from committing (it's OK if it fixes things that are fixable, though).

schloerke: It will block, but black would also block for more extreme situations like file can't parse.

To me, it feels like there is more benefit to having it than not. 

For the rare case that you'd like to commit with no pre-commit hooks, you can use `--no-verify` in the commit call.

From https://stackoverflow.com/a/65397546/591574
> Git config
> ```
> [alias]
>         fcommit = commit --no-verify
> ```
> Run command that skips the hooks
> ```
> git fcommit -m "MSG"
> ```

PR #1220, opened 2024-03-13
Modules for Shiny Express

jcheng5:
Example:

### app.py

```python
from counter import counter

from shiny.express import ui

counter("one")
ui.hr()
counter("two")
```

### counter.py

```python
from shiny import reactive
from shiny.express import module, render, ui


@module
def counter(input, output, session, starting_value=0):
    count = reactive.value(starting_value)

    ui.input_action_button("btn", "Increment")

    with ui.div():

        @render.express
        def current_count():
            count()

    @reactive.effect
    @reactive.event(input.btn)
    def increment():
        count.set(count() + 1)
```

### DIFF
--- a/shiny/express/__init__.py
+++ b/shiny/express/__init__.py
@@ -11,6 +11,7 @@ from ..session import (
 from .. import render
 from . import ui
 from ._is_express import is_express_app
+from ._module import module
 from ._output import (  # noqa: F401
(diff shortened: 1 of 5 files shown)


wch (approved):

wch:
Note that there is a type-checking error, but it seems to be new to pyright 1.1.359 and is unrelated to the changes in this PR.

PR #1233, opened 2024-03-20
For Express, emit message if need to upgrade rsconnect-python

wch:
Addresses #1232.

When a user runs `shiny run app.py` on a Shiny Express app, this checks the version of the rsconnect-python package.

If (A) rsconnect-python is installed and (B), the version is less than 1.22.0, then this will print a message to stderr like this, and then it will run the app as usual.

```
Warning: rsconnect-python 1.21.0 is installed, but it does not support deploying Shiny Express applications. Please upgrade to at least version 1.22.0. If you are using pip, you can run `pip install --upgrade rsconnect-python
```

### DIFF
--- a/shiny/_main.py
+++ b/shiny/_main.py
@@ -10,7 +10,7 @@ import re
 import sys
 import types
 from pathlib import Path
-from typing import Any, Literal, Optional
+from typing import Any, Optional
 
 import click
 import uvicorn
--- a/shiny/_main.py
+++ b/shiny/_main.py
@@ -663,7 +663,6 @@ def _verify_rsconnect_version() -> None:
 
     try:
         package_version = version(PACKAGE_NAME)
-        print(package_version)
         if _compare_package_versions(package_version, MIN_VERSION) < 0:
             print(
                 f"Warning: rsconnect-python {package_version} is installed, but it does not support deploying Shiny Express applications. "
--- a/shiny/_main.py
+++ b/shiny/_main.py
@@ -632,40 +632,22 @@ class ReloadArgs(TypedDict):
 
 
 # Check that the version of rsconnect supports Shiny Express; can be removed in the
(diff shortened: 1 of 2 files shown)


shiny/_main.py
```diff
@@ -626,3 +629,47 @@ class ReloadArgs(TypedDict):
...
+            if part1 > part2:
+                return 1
+            elif part1 < part2:
+                return -1
+
+        return 0
+
+    try:
+        package_version = version(PACKAGE_NAME)
+        print(package_version)
```
gshotwell, line 666: Errant print?

shiny/_main.py
```diff
@@ -626,3 +629,47 @@ class ReloadArgs(TypedDict):
...
+
+# Check that the version of rsconnect supports Shiny Express; can be removed in the
+# future once this version of rsconnect is widely used. (Added 2024-03)
+def _verify_rsconnect_version() -> None:
+    PACKAGE_NAME = "rsconnect-python"
+    MIN_VERSION = "1.22.0"
+
+    from importlib.metadata import PackageNotFoundError, version
+
+    def _safe_int(x: str) -> int:
```
gshotwell, line 642: I think we already have an indirect dependency on `packaging`, do we want to make it direct  and use their version parsing? 

```
    from packaging.version import parse
    from importlib.metadata import PackageNotFoundError, version
    import sys

    try:
        installed_version = parse(version(PACKAGE_NAME))
        required_version = parse(MIN_VERSION)
        if installed_version < required_version:
      ...
```

shiny/_main.py
wch: Oh oops, I had deleted it on my local copy of the file, but forgot to commit. 🤦

shiny/_main.py
wch: Good call -- I see that htmltools uses `packaging` to check versions, so I've updated the code as you suggested.

Issue #1237, opened 2024-03-21
Updates when python min version is 3.9


schloerke:

wch:
Just curious: why does this Python 3.9?

schloerke:
When inheriting from a class that has a ParamSpec, in Python >= 3.9 the paramspec can be defined as `[]`. In python 3.8, it does not understand it and raises. You can pass in a `P` paramspec, but for the case of `AsycnValueFn`, the ParamSpec should be empty and I couldn't find a way to define it as empty.

schloerke:
Closing as it will not be utilized

PR #1238, opened 2024-03-21
chore: Remove dead code until python 3.8 support is dropped

schloerke:
Related #1237


schloerke:
Python 3.13 is scheduled for release on 2024-10-01 (https://peps.python.org/pep-0719/).  

Would rather keep the commented code and then support the generic base class properly when we drop support for python 3.8

schloerke:
Closing in favor of #1701

PR #1242, opened 2024-03-23
chore: Install packages using UV; Be explicit within Makefiles 

schloerke:
Subset of `uv` installation from #1211

Require users to use `.venv` folder and not `venv`.

By using `uv` to install, GHA install time is reduced from ~90s to ~25s.

Q: 
- [ ] Should we auto remove the legacy `venv` folder? Present a message? Do nothing?
- [ ] Should we instead make uv leverage `venv` folder (and not `.venv` by default)?


schloerke:
As much as I like this PR, it requires that the user has a `.venv` folder. While this is automatically handled, it is too opinionated for a public usage repo (where as it would be fine for the website).

Closing as https://github.com/posit-dev/py-shiny/pull/1565 contains a good middle ground solution of using `uv` on CI, but not requiring users to have uv or use uv's `.venv`.

PR #1248, opened 2024-03-26
api(render.data_frame): Separate `mode=` into `selection_mode=` and `editable=`

schloerke:
```python
class DataGrid():
    def __init__(
        ....
        editable: bool = False,
        selection_mode: Literal["none", "row", "rows"] = "none",
        ....
class DataGrid():
    def __init__(
        ....
        editable: bool = False,
        selection_mode: Literal["none", "row", "rows"] = "none",
        ....
```

... with room to grow with `"col"`, `"cols"`, and `"rect"`.


shiny/render/_data_frame_utils/_datagridtable.py
```diff
@@ -0,0 +1,331 @@
...
+
+def as_editable(
+    editable: bool,
+    *,
+    name: str,
+) -> bool:
+    editable = bool(editable)
+    if editable:
+        print(
+            f"`{name}(editable=true)` is an expirmental feature. "
```
wch, line 53:
```suggestion
            f"`{name}(editable=true)` is an experimental feature. "
```

shiny/render/_data_frame_utils/_selection.py
```diff
@@ -0,0 +1,450 @@
...
+# (same for cols)
+# select single col; all rows
+x = {"cols": col_idx}
+x = {"cols": col_idx, "rows": None}
+x = {"cols": [col_idx]}
+x = {"cols": [col_idx], "rows": None}
+# select multiple cols; all rows
+x = {"cols": [col_idx, col_idx + 3, col_idx + 5]}
+x = {"cols": [col_idx, col_idx + 3, col_idx + 5], "rows": None}
+
```
wch, line 65: Seems like there's a lot of debugging/testing code in this file which can be removed.

shiny/render/_data_frame_utils/_datagridtable.py
```diff
@@ -0,0 +1,331 @@
...
+    def __init__(
+        self,
+        data: pd.DataFrame | PandasCompatible,
+        *,
+        width: str | float | None = "fit-content",
+        height: str | float | None = "500px",
+        summary: bool | str = True,
+        filters: bool = False,
+        editable: bool = False,
+        selection_mode: SelectionMode = "none",
```
wch, line 244: When we enable column and region selection mode, how will people specify that they want to enable some combination of these, for example, multi-row, multi-column, and region selection?

shiny/render/_data_frame_utils/_selection.py
```diff
@@ -0,0 +1,450 @@
...
+#     # Region
+#     if is_location_range(x["rows"]) and is_location_range(x["cols"]):
+#         return {"type": "region", "rows": x["rows"], "cols": x["cols"]}
+#     else:
+#         raise ValueError(
+#             "SelectionLocation['rows'] and ['cols'] must be a tuple[int, int]"
+#         )
+
+
+def to_list_or_none(x: int | Sequence[int] | None) -> Sequence[int] | None:
```
wch, line 285:
```suggestion
def to_list_or_none(x: int | Sequence[int] | None) -> list[int] | None:
```

Also, I'm not sure if this will be helpful, but you could add overrides which say that `None` will go to `None`, and `int | Sequence[int]` will go to `list[int]`.

shiny/render/_data_frame_utils/_selection.py
```diff
@@ -0,0 +1,450 @@
...
+) -> SelectionMode:
+    if row_selection_mode == "deprecated":
+        # Disable selection_mode if `editable=True``
+        if editable:
+            warnings.warn(
+                '`editable` can not be `True` while `selection_mode` != `"none"`. '
+                'Setting `selection_mode="none"`',
+                stacklevel=2,
+            )
+            return "none"
```
wch, line 434: I'm not sure I fully understand the logic of how `row_selection_mode` is used here... Can you clarify with a comment?

But regarding the warning: this is temporary, right? At some point we will want to allow row selection and editing.

shiny/render/_dataframe.py
```diff
@@ -483,7 +193,8 @@ class data_frame(Renderer[DataFrameResult]):
     Returns
     -------
     :
-        * If the row selection mode is `None`, the calculation will throw a silent error (`req(False)`)
+        * If the row selection mode is `None`, the calculation will throw a silent error
+          (`req(False)`)
```
wch, line 197: It would be helpful to have a link to `req` docs here.

shiny/render/_dataframe.py
```diff
@@ -743,51 +476,50 @@ async def _send_message_to_browser(self, handler: str, obj: dict[str, Any]):
...
-                "You can't update row selections when row_selection_mode is 'none'"
-            )
-
-        if mode == "single_row" and len(idx) > 1:
-            raise ValueError(
-                "Attempted to set multiple row selection values when row_selection_mode is 'single'"
+            selection_mode = self.selection_mode()
+        if selection_mode == "none":
+            warnings.warn(
+                "Cell selection can not be updated when `.selection_mode()` is 'none'. "
```
wch, line 491:
```suggestion
                "Cell selection cannot be updated when `.selection_mode()` is 'none'. "
```

shiny/render/_data_frame_utils/_selection.py
```diff
@@ -0,0 +1,450 @@
...
+
+
+# For receiving selection info from JS:
+BrowserCellSelection = Union[
+    BrowserCellSelectionRegion,
+    BrowserCellSelectionRow,
+    BrowserCellSelectionCol,
+    BrowserCellSelectionAll,
+    BrowserCellSelectionNone,
+]
```
wch, line 187: I mentioned this elsewhere, but how will we support multiple types of selections in the same data table?

shiny/render/_data_frame_utils/_selection.py
schloerke: The overload type did not work when it was being used. Skipping.

Suggestion: √

shiny/render/_data_frame_utils/_selection.py
schloerke: Commented code has been reduced

shiny/render/_data_frame_utils/_datagridtable.py
schloerke: `selection_mode` -> `selection_modes`

shiny/render/_data_frame_utils/_selection.py
```diff
@@ -177,274 +162,258 @@ class BrowserCellSelectionRegion(TypedDict):
...
+        )
+        selection_modes_tuple = ("none",)
+
+    # Trim values
+    if "none" in selection_modes_tuple:
+        return ("none",)
+    if "row" in selection_modes_tuple and "rows" in selection_modes_tuple:
+        selection_modes_tuple = tuple(
+            mode for mode in selection_modes_tuple if mode != "row"
+        )
```
wch, line 413: I think this will result in some slightly odd behavior :

- If the input is `(,)`, then this returns `(,)`
- If the input is `("rows", "none")`, then this returns `("none",)`
- If the input is `("rows", "row", "rows")`, then this returns `("rows", "rows")`

Also a small nit: it's inconsistent that the none code path has an immediate return, but the row+rows code path sets a value that's returned later.

Something like this will be more robust, and it will disallow unknown values. It also will mean that the explanation of resolution rules can be simpler:

```py
selection_modes_set = set(selection_modes)

if not selection_modes_set.issubset({"none", "row", "rows", "col", "cols"}):
    raise ValueError( ... )

if len(selection_modes_set) == 0:
    return ("none",)

if "none" in selection_modes_set and len(selection_modes_set) > 1:
    raise ValueError( ... )

if "row" in selection_modes_set and "rows" in selection_modes_set:
    raise ValueError( ... )

if "col" in selection_modes_set and "cols" in selection_modes_set:
    raise ValueError( ... )

return tuple(s)
```

Although, maybe the internal representation for no selection _should_ just be an empty tuple?

wch: One more thought: I think that the "correct" data structure for selection modes actually looks like this:

```py
@dataclass
class SelectionModes:
    row: Literal["none", "single", "multiple"]
    col: Literal["none", "single", "multiple"]
    rectangle: Literal["none", "multiple"]  # This one might need better names
```

Then there's no wiggle room in the type system (no empty tuples, no duplicate entries, etc) and it makes some of the code slightly cleaner -- instead of searching in a tuple for `"rows"`, you can index directly to it, as in `s.row == "multiple"`.

That said, since this is small and internal-facing, the exact representation isn't super important, and we can always change it in the future. But to me this feels like the right way to represent this information.

wch (commented):
Just a few comments. Note that I feel strongly about the argument name, but the other stuff isn't as important.

shiny/render/_data_frame_utils/_datagridtable.py
```diff
@@ -91,14 +97,19 @@ class DataGrid(AbstractTabularData):
         If `True`, allows the user to edit the cells in the grid. When a cell is edited,
         the new value is sent to the server for processing. The server can then return
         a new value for the cell, which will be displayed in the grid.
-    selection_mode
-        Single string to set the mode of the table.
+    selection_modes
+        Single string or a Sequence of string values to define possible ways to select
```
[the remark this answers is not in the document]
schloerke:
```suggestion
        Single string or a `list`/`tuple` of string values to define possible ways to select
```

shiny/render/_data_frame_utils/_selection.py
schloerke: Using new class structure in python and TS

Currently using `rect: Literal["cell", "region"]` within the class

schloerke:
Going to merge. Any changes can be made in a followup PR 😄

PR #1253, opened 2024-03-27
feat: Create `Session.set_message_handler(name, handler)`

schloerke:
> Sets a method that can be called by the client via `Shiny.shinyapp.makeRequest()`. `Shiny.shinyapp.makeRequest()` makes a request to the server and waits for a response. By using `makeRequest()` (JS) and `set_message_handler()` (python), you can have a much richer communication interaction than just using Input values and re-rendering outputs.
> 
> For example, `@render.data_frame` can have many cells edited. While it is possible to set many input values, if `makeRequest()` did not exist, the data frame would be updated on the first cell update. This would cause the data frame
> to be re-rendered, cancelling any pending cell updates. `makeRequest()` allows for individual cell updates to be sent to the server, processed, and handled by the existing data frame output.
> 

------------

Removes `Session._output_binding_request_handler()` and all methods used to support it. 

Now, there is no coincidences and users opt-in directly to the method: 

Ex: 
```
        ...
        request_key = session.set_message_handler(
            f"data_frame_patches_{self.output_id}",
            handler_fn,
        )
```

The `request_key` is then set to the client so that it know where it can make requests. The `request_key` is namespaced by the session, so authors do not need to handle it themselves.


shiny/session/_session.py
```diff
@@ -406,9 +411,18 @@ async def _dispatch(self, message: ClientMessageOther) -> None:
 
         try:
             # TODO: handle `blobs`
-            value = await func(*message["args"])
+
+            # Isolate so that the function can be run without an explicit session.
+            # Let the handler function manage its own session context. (due to modules)
```
wch, line 416: Change comment to explain that the function should be run on demand, and without throwing an error due to reading a reactive value in a non-reactive function.

shiny/session/_session.py
```diff
@@ -390,7 +395,7 @@ def _is_hidden(self, name: str) -> bool:
 
     async def _dispatch(self, message: ClientMessageOther) -> None:
         try:
-            func = self._message_handlers[message["method"]]
+            async_func = self._message_handlers[message["method"]]
```
wch, line 398: Store both the session and the function here, so that we can use `with_session_context()` below.

wch (approved):

PR #1296, opened 2024-04-09
Fix panel_conditional examples

Pierre-Bartet:
The `panel_conditional` examples are broken as can be seen in the [doc](https://shiny.posit.co/py/api/core/ui.panel_conditional.html) (the elements are always shown whatever the condition's value):

image

This is simply caused by the same id "slider" being used both for the `input_select` and the `input_slider`. Using different ids fixes the issue:

image


cpsievert:
Thanks!

PR #1297, opened 2024-04-09
Add `dev_mode` param to `App()`

wch:
This PR does a few things:

- It adds a new parameter so you can call `App(dev_mode=True)`.
- When the `App` object is created, if `dev_mode` is `None` (the default), then it looks at the env var `SHINY_DEV_MODE`. The setting for `dev_mode` is stored in the `App` object.
- Before the change in this PR, it only used that env var to determine whether to dev mode, and it looked at the env var each time that `shiny_deps()` was called. This is potentially problematic when running multiple concurrent apps, as with shinylive apps embedded in a web page. With the change, the `App` object stores the value of `dev_mode`, so the setting for one `App` will not affect any others.

The reason for this PR is for Shinylive -- I want to enable the dev mode error console when there's are both editor and viewer components, but not when there's just the viewer component.


schloerke (approved):

wch:
Closing because this was fixed a different way in https://github.com/posit-dev/shinylive/commit/5290904a5a7283e0ef0883bb028782723c9b872b

PR #1300, opened 2024-04-10
feat(data frame): Render Tag-like cells

schloerke:
New feature: Data Frame cells now support `htmltools.TagNode` like objects: `Tag`, `TagList`, `Tagifiable`, `HTML`, `ReprHtml`.  

These types of values will be processed by `session._process_ui(val)` and sent to the browser to be rendered with `window.shiny.renderContentAsync(tsVal)` and de-rendered via `window.Shiny.unbindAll(curTdRef)` and `curTdRef.replaceChildren("")`

If any pandas column type is marked as `unknown`, the cells are always searched for any tag like objects. If one is found, the column type is marked as `"html"`. 

If a edited cell returns a tag-like object, it will be processed just like an HTML column. Currently, cell patches are not type checked, however, `._type_hints` is now being calculated and contains the most recent type hints sent to the browser.

Other updates
* HTML-based cells are not editable
	* If a cell to the left hits `tab` to enter an HTML-based cell, the next possible cell is entered. If a boundary of the table is hit, we back out and the current cell stays selected.

Questions:
* Should html based cells be sortable?
	* A: Disabling
* Should html based cells be filterable?
	* A: Disabling


DataVizLucas:
Is this feature also planned for the render.text output?

wch:
@DataVizLucas If you want to dynamically render HTML in the page, you can use `@render.ui`.

PR #1302, opened 2024-04-11
use pinned version instead of develop version for rsconnect

karangattu:
The deploy shiny apps to connect were failing since we were using a develop version `1.23.1.dev93+g7d5713b`
Changing it to the PyPI version - `1.23.0` seems to fix it.


schloerke (approved):

wch:
There's a good chance this has something to do with the changes I've recently made to rsconnect-python. I looked at the logs from Github Actions (for example https://github.com/posit-dev/py-shiny/actions/runs/8640466203/job/23688460059) but it wasn't clear to me what exactly was going wrong with the deployment. Do you have any details on why the deployment failed?

karangattu:
So we swallow the exception when a deploy fails to connect or shinyapps.io to prevent the secrets or tokens from being exposed in the stdout.
When I was debugging this on my local machine with the develop version of `rsconnect`, I got a 
```sh
  File "/Users/.pyenv/versions/3.10.12/lib/python3.10/site-packages/rsconnect/api.py", line 1056, in validate_app_mode
    existing_app_mode = AppModes.get_by_ordinal(app.app_mode, True)
AttributeError: 'dict' object has no attribute 'app_mode'
Internal error: 'dict' object has no attribute 'app_mode'
```
error

wch:
@karangattu Thanks for that information, I'll look into it.

wch:
I've created a PR with a fix in https://github.com/rstudio/rsconnect-python/pull/579. Once that's merged, can we revert this PR which pinned the version of rsconnect? Having it not pinned was helpful in detecting the bug.

schloerke:
Sure, np

PR #1306, opened 2024-04-11
bug(data frame): Unselect single row when clicked

### LINKED ISSUE #743: Deselect or clear selected rows in Datagrid
Hi, 
Loving shiny for python and the new interactive tables!

In your example:
https://shinylive.io/py/examples/#data-frame-grid

I noticed that you can only deselect rows when selection is "multiple", not when "single". Would it be possible to update so we can deselect rows even when using single?

On a related note, is there a way to programatically update selected rows e.g. if I want to sync two tables so that the equivalent rows are selected when I select a row in one of the tables?

Thanks for your help!

schloerke:
Fixes #743


schloerke:
After talking @wch, there is an out by using the CMD key. 

The current `"row"` implementation is similar to how folders are selected or even rows are selected in google sheets. 

Therefore, we should **not** add toggle on click and leave it as it.

---------------------

Closing

PR #1331, opened 2024-04-23
Add `Session` abstract base class

wch:
This PR does the following:

- Renames the existing `Session` class to `AppSession`.
- Adds a new abstract base class named `Session`, from which `AppSession`, `SessionProxy`, and `ExpressMockSesssion` are subclasses of.
- Adds a new method, `Session.is_real_session()`, which returns `True` for `AppSession` and `SessionProxy`s made from an `AppSession`, and returns `False` for `ExpressMockSession` and `SessionProxy`s made from one of those.

The new `Session` class is an abstract base class, but it may be more useful to think of it as an interface which the other classes must implement.

Most existing code which takes a `Session` object can remain unchanged. In most cases, when code wants a `Session` object, it can work with an `AppSession`, `SessionProxy`, or `ExpressMockSession`.

- An `AppSession` represents a real, instantiated Shiny session for an app
- An `ExpressMockSession` represents the fake session object that is present when rendering the UI of a Shiny Express app. Many of the methods on this class should not be called in practice, and so they will raise an exception if called.
- A `SessionProxy` represents a Shiny module session, where the parent a `Session`. The parent may be an `AppSession` or `ExpressMockSession`, or another `SessionProxy`.

All of these must implement the various methods defined in `Session`.

One of the useful things about this is that `Session` class makes it very clear which attributes and methods are part of the interface we want to expose in `AppSession` and `SessionProxy`. The `AppSession` contains many other attributes and methods, but they are implementation details. With the state of the code before this PR, it was not clear which methods were meant to be part of the external interface and which were meant to be internal.

This is still a work in progress, and there are still some open questions about which methods should be part of the interface and which should not.

*****

Notes:
- The name `.is_real_session()` isn't great, so I'm open to other ideas.
- There are a number of properties which seem like they shouldn't be part of the interface, but they are. These include `_outbound_message_queues`, `_downloads`, and `_is_hidden`.
- There are a number of methods which have a leading underscore, which makes it seem like they should be internal -- but in practice I found that they needed to be added to the new base class because they are somewhat external-facing. For example, `_send_message`, `_send_message_sync`, `_process_ui`. So perhaps these should be renamed to remove the leading underscore.
- Docstrings from the `AppSession` class should be moved to `Session`.


shiny/express/_mock_session.py
```diff
@@ -39,12 +58,81 @@ def on_ended(
     ) -> Callable[[], None]:
         return lambda: None
 
-    def make_scope(self, id: Id) -> Session:
+    def make_scope(self, id: Id) -> SessionProxy:
```
jcheng5, line 61: Shouldn't this still be `-> Session`? I think of the fact that it's actually a `SessionProxy` as an implementation detail.

shiny/express/_mock_session.py
```diff
@@ -39,12 +58,81 @@ def on_ended(
...
+    ):
+        self._not_implemented("download")
 
-    def __getattr__(self, name: str):
-        raise AttributeError(
+    def _not_implemented(self, name: str) -> Never:
+        raise NotImplementedError(
             textwrap.dedent(
                 f"""
             The session attribute `{name}` is not yet available for use. Since this code
```
jcheng5, line 138: This message needs to change to use the `is_real_session` check instead, right?

shiny/express/_mock_session.py
```diff
@@ -39,12 +58,81 @@ def on_ended(
...
+        self._not_implemented("_unhandled_error")
+
+    def download(
+        self,
+        id: Optional[str] = None,
+        filename: Optional[str | Callable[[], str]] = None,
+        media_type: None | str | Callable[[], str] = None,
+        encoding: str = "utf-8",
+    ):
+        self._not_implemented("download")
```
jcheng5, line 132: I think most of these could be no-ops (not `_unhandled_error`)

shiny/express/_mock_session.py
```diff
@@ -24,11 +30,24 @@ class ExpressMockSession:
...
+        self.app = None  # pyright: ignore
+        self.id = None  # pyright: ignore
+
+        self._outbound_message_queues = None  # pyright: ignore
+        self._downloads = None  # pyright: ignore
 
         # Application-level (not session-level) options that may be set via app_opts().
         self.app_opts: AppOpts = {}
 
+    def is_real_session(self) -> Literal[False]:
```
jcheng5, line 48: Regarding the name of this method, eventually we'll probably have a MockSession that's used for testing, including testing server code (so would "implement" a lot more than ExpressMockSession, I'd imagine). If that happens, `is_real_session` becomes a bit muddy.

`is_server_session`? `is_user_session`?

Naming things is hard...

jcheng5:
I think this is a good start. Doesn't need to be part of this PR but it strikes me that part of what makes the AppSession class feel bad is that it does a whole bunch of things that are not that related to each other, and it's all jammed together because we want them to be presented to the user in one `session` object. It would be nice to extract out the major pieces of functionality into separate classes, which the AppSession then instantiates and delegates.

wch:
I agree that at some point it would make sense to break it up into separate classes.

PR #1340, opened 2024-04-26
Make sure ionrangeslider also gets sass options in build script

cpsievert:
Before this change, it was possible for `script/htmlDependencies.R` to generate a CSS file for ionrangeslider that included a large embedded source map (when `options(shiny.devmode = TRUE)`). This will prevent that from happening again


gadenbuie (approved):
Thanks!

PR #1393, opened 2024-05-17
fix: Add wait till pulse animation has started

karangattu:
Tests were failing since the pulse busy indicator had not started before we started assertions.
Now we verify computations have stopped before we do assertions.

### DIFF
--- a/tests/playwright/shiny/components/busy_indicators/test_busy_indicators.py
+++ b/tests/playwright/shiny/components/busy_indicators/test_busy_indicators.py
@@ -50,7 +50,8 @@ def test_busy_indicators(page: Page, local_app: ShinyAppProc) -> None:
     output_txt.expect_value("0")
     spinner_type.set("pulse")
     render_button.click()
-    # since we are not using locators wait for up to 2 secs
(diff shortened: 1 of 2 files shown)


tests/playwright/shiny/components/busy_indicators/test_busy_indicators.py
```diff
@@ -46,11 +46,21 @@ def test_busy_indicators(page: Page, local_app: ShinyAppProc) -> None:
         assert get_spinner_computed_property(page, element_id, "width") == height
 
     # Verify pulse indicator behavior
-    expect_not_to_have_class(page.locator("html"), "shiny-busy", timeout=8000)
+    output_txt = OutputTextVerbatim(page, "counter")
+    output_txt.expect_value("0")
     spinner_type.set("pulse")
     render_button.click()
+    # since we are not using locators wait for up to 2 secs
```
schloerke, line 53:
```suggestion
    # `::after` is not an implemented selector in playwright
    # Since we are not using locators wait for up to 2 secs
```

schloerke (approved):

PR #1400, opened 2024-05-21
In ExpressStubSession, set 'app' and 'id'

wch:
This fixes a problem with using modules in Shiny Express. Prior to the change, this app would raise an error; after the change, it works:

```py
from shiny import reactive
from shiny.express import module, render, ui


@module
def counter(input, output, session, starting_value: int=0):
    count = reactive.value(starting_value)

    ui.input_action_button("btn", "Increment")

    with ui.div():

        @render.express
        def current_count():
            count()

    @reactive.effect
    @reactive.event(input.btn)
    def increment():
        count.set(count() + 1)


counter("one")
ui.hr()
counter("two")
```

### DIFF
--- a/docs/_quartodoc-express.yml
+++ b/docs/_quartodoc-express.yml
@@ -137,6 +137,10 @@ quartodoc:
         - express.ui.modal_remove
         - express.ui.modal_button
         - express.ui.Progress          # uses class.rst
+    - title: Modules
+      desc: ""
+      contents:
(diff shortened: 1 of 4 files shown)


jcheng5:
Can you also put a code example on the `shiny.express.module` function? That would've been enough for CI to catch this problem, right?

wch:
Yup, good idea.

PR #1416, opened 2024-05-23
bug(CI): Only update the shiny submodule when making the dev docs

schloerke:
https://github.com/posit-dev/shinylive/blob/144a5de66cc198044c918df9d786e92804d9ff25/Makefile#L94C4-L94C20

### DIFF
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -67,7 +67,7 @@ jobs:
         run: |
           cd shinylive-repo
           make submodules
-          make submodules-pull
+          make submodules-pull-shiny
 
       - name: Build shinylive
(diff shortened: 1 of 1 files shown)


.github/workflows/build-docs.yaml
```diff
@@ -67,7 +67,7 @@ jobs:
         run: |
```
schloerke, line 67: Update the step name

.github/workflows/build-docs.yaml
```diff
@@ -67,7 +67,7 @@ jobs:
...
           cd shinylive-repo
           make submodules
-          make submodules-pull
```
schloerke, line 70: Also update htmltools

schloerke:
Closing in favor of #1417

PR #1453, opened 2024-06-05
Add `shiny.ui.Chat`

cpsievert:
Adds a `shiny.ui.Chat` class, designed to support any LLM provider (i.e., response assistant) of your choosing (e.g., OpenAI, Anthropic, Ollama, LangChain, Google, etc). To get started, consider this bare-bones `Chat` example that uses no provider at all. It just displays a starting message, and then for each input message, just adds `"You said: "` in it's response:

<details>

<summary>App code</summary>

```python
from shiny.express import ui

ui.page_opts(title="Hello Shiny Chat")

# Create a chat instance, with an initial message
chat = ui.Chat(
    id="chat",
    messages=[
        {"content": "Hello! How can I help you today?", "role": "assistant"},
    ],
)

# Display the chat
chat.ui()

# Define a callback to run when the user submits a message
@chat.on_user_submit
async def _():
    user_msg = chat.get_user_input()
    await chat.append_message(f"You said: {user_msg}")
```

</details>

Screenshot 2024-06-28 at 4 28 47 PM

An "actual" LLM-powered chat bot might look something more like this (this example requires an OpenAI API key to run). Note also that, by default, responses are interpreted as markdown strings, and code blocks are rendered with code highlighting + copy/paste:

<details>
<summary>App code</summary>

```python
# A basic Shiny Chat example powered by OpenAI via LangChain.
# To run it, you'll need OpenAI API key.
# To get one, follow the instructions at https://platform.openai.com/docs/quickstart
# To use other providers/models via LangChain, see https://python.langchain.com/v0.1/docs/modules/model_io/chat/quick_start/
import os

from langchain_openai import ChatOpenAI

from shiny.express import ui

# Provide your API key here (or set the environment variable)
llm = ChatOpenAI(api_key=os.environ.get("OPENAI_API_KEY"))

# Set some Shiny page options
ui.page_opts(
    title="Hello LangChain Chat Models",
    fillable=True,
    fillable_mobile=True,
)

# Create and display an empty chat UI
chat = ui.Chat(id="chat")
chat.ui()


# Define a callback to run when the user submits a message
@chat.on_user_submit
async def _():
    # Get messages currently in the chat
    messages = chat.get_messages()
    # Create a response message stream
    response = llm.astream(messages)
    # Append the response stream into the chat
```

</details>

Screenshot 2024-06-28 at 4 22 05 PM

A whole collection of other examples are available under the `examples/chat` directory. See the `basic` and `enterprise` sub-directories for getting started with various providers.

### Follow up tasks

- [x] Tests


examples/chat/dynamic-ui/app.py
```diff
@@ -0,0 +1,24 @@
+from openai import OpenAI
+
+from shiny.express import input, render, ui
+
+ui.input_select("model", "Model", choices=["gpt-4o", "gpt-3.5-turbo"])
+
+client = OpenAI()
+
+
+@render.ui
```
cpsievert, line 10: Come up with a better example for this and use render.express

examples/chat/openai/app.py
```diff
@@ -0,0 +1,29 @@
...
+
+# on user submit, generate and append a response
+@chat.on_user_submit
+async def _():
+    response = await client.chat.completions.create(
+        model="gpt-4o",
+        messages=chat.messages(),
+        stream=True,
+    )
+    await chat.append_message_stream(response)
```
cpsievert, line 29: Look into a LangChain version of this.

examples/chat/RAG/recipes/utils.py
```diff
@@ -0,0 +1,106 @@
+import aiohttp
```
wch, line 1: This package doesn't seem to be installed after doing `pip install -e ".[dev]"`, so I think it needs to be added to `setup.cfg` explicitly.

examples/chat/basic/openai/app.py
```diff
@@ -0,0 +1,36 @@
...
+    messages=[
+        {"content": "Hello! How can I help you today?", "role": "assistant"},
+    ],
+    # assistant_response_transformer=lambda x: HTML(f"<h1>{x}</h1"),
+)
+
+# Display the chat
+chat.ui()
+
+# Create the LLM client (assumes OPENAI_API_KEY is set in the environment)
```
wch, line 24: It would help to have API key setup instructions at the top of each app -- without that information up front, there's a lot of friction in trying to just run each app.

And in addition, it would also be good if there was a README in the examples/chat directory that explained how to set these environment variables and how to get the keys from the providers.

iainmwallace: It would be great if the set up instructions could also show how to use the azure end points

cpsievert: @iainmwallace I assume you're referring to [these Azure docs](https://learn.microsoft.com/en-us/azure/ai-services/openai/quickstart?tabs=command-line%2Cpython-new&pivots=programming-language-python#create-a-new-python-application)? If you start from `./examples/chat/basic/openai/app.py`, it's hopefully pretty straightforward to change `AsyncOpenAI()` to `AzureOpenAI()`?

wch: One more idea: use the `python-dotenv` package, similar to how it's done in https://github.com/jcheng5/multimodal/blob/main/query.py.

This would allow users to create a `.env` file with the key, and then the app would automatically use that file if present; if they don't want to create a `.env` file, then setting the environment variable will still work.

For this it would be good to also add `.env` to a `.gitignore` in the examples/chat/ directory.

shiny/ui/_chat.py
```diff
@@ -0,0 +1,782 @@
...
+            sanitize = self.on_error == "sanitize"
+            raise NotifyException(str(e), sanitize=sanitize)
+
+    def get_messages(
+        self,
+        *,
+        token_limits: tuple[int, int] | None = (4096, 1000),
+        apply_user_transform: bool = True,
+        apply_assistant_transform: bool = False,
+    ) -> tuple[ChatMessage, ...]:
```
wch, line 286: The return type of this function is `tuple[ChatMessage, ...]`. Looking at the type annotations for anthropic and openai and ollama, it looks to me like the message formats between them are similar, but they diverge in how they represent content other than basic text (like images, function calls, and other metadata).

I think we'll need to figure out how to handle these other kinds of content. One way we could do it is to simply just store the llm provider's native format without modification (I think that's what currently happens in this PR?). If we do that, then there are two challenges:

- Any Python code that operates on content like images will need to know how to handle the various formats for different LLMs. For example, you could imagine that the app would save an image that's generated by the LLM, or for an image that's uploaded by the user, the app would resize it before sending it to the LLM. A function like this would need to know how to handle the different kinds of representations for each LLM provider.
- The client-side UI code will need to know how to render the different kinds of data sent from the server.

On the other hand, if we were to (internally) normalize the data in Python to a single type of representation, then when `get_messages()` is called, we would need to have a way to transform the data to the right format for the particular LLM provider. That might be another parameter to this method, and/or an option when instantiating the class. Some thoughts here:
- It simplifies the client-side code so that it would only need to handle a single format.
- Coming up with an internal representation that is rich enough to encompass all the LLM providers will take some care.
 

I don't think we necessarily have to have every detail of this figured out right away, but we should design it with the goal of not painting ourselves into a corner.

cpsievert: > store the llm provider's native format without modification (I think that's what currently happens in this PR?)

`normalize_message()`/`normalize_message_chunk()` is currently applied before storage, meaning we essentially go from the provider's format to `ChatMessage` for storage. In addition to the benefits you mention, this way we can guarantee a consistent type across user and assistant messages. Also, perhaps more importantly, if we were to instead return the provider format verbatim, you wouldn't get able to pass `.get_messages()` directly into to the `.create()` method.

If and when we do go to add multi-modal input/output, `ChatMessage`'s `content` key will have to change to support a list of content types. I don't think we necessarily have to make a breaking change for this, but we might need to introduce a new class/option that you opt-into which assumes `list[ContentType]` instead of `str`. Also, when it comes to defining `ContentType`, I'm thinking we'll probably roughly follow [OpenAI's convention](https://community.openai.com/t/how-to-get-an-image-described/607957/3)

cpsievert: Added a follow up task item about this (the return type of `.get_messages()`)

shiny/ui/_chat.py
```diff
@@ -0,0 +1,900 @@
...
+
+
+def _express_is_active() -> bool:
+    from ..express._run import get_top_level_recall_context_manager
+
+    try:
+        get_top_level_recall_context_manager()
+        return True
+    except RuntimeError:
+        return False
```
cpsievert, line 888: @wch this apparently will return `True` in some Shiny Core situations. Any idea how we do this more accurately?

wch: This function is used from `chat.ui()`. The case where it could be problematic is when `chat.ui()`is used in a function marked with `@expressify` or `@shiny.express.module` AND that function/module is used in a Shiny Core app. In that case, this `_express_is_active()` function will return `False`, and `chat.ui()` will see that value and raise an error, but that would be incorrect behavior.

It _should_ be possible to use `chat.ui()` from a function/module that's decorated with `@expressify` or `@shiny.express.module`, even if that function/module is used in a Shiny Core app.

wch: After some discussion, @cpsievert and I decided just to not use this check.

wch (approved):

PR #1454, opened 2024-06-06
feat(data frame): Enable cell selection and editable cells

schloerke:
TODO:
- [x] Remove restriction
- [x] Improve keyboard navigation on selected cell
- [x] Require selected cell to be double clicked to enter
- ~Allow selected cell to have typed letter enter edit mode~
- [x] Pressing Enter with a selected row begins editing the first visible column in that row


js/data-frame/index.tsx
```diff
@@ -123,6 +119,27 @@ const ShinyDataGrid: FC<ShinyDataGridProps<unknown>> = ({
 
   const editCellsIsAllowed = payloadOptions["editable"] === true;
 
+  // const [cellBeingEdited, setCellBeingEdited] = useState<{
+  //   rowIndex: number;
+  //   columnIndex: number;
+  // } | null>(null);
+
+  const [isEditingCell, setIsEditingCell] = useState(false);
```
jcheng5, line 127: This should be a useMemo

PR #1467, opened 2024-06-18
Use `np.nan`

wch:
In numpy 2.0, they removed `np.NaN`. This PR switches to `np.nan` instead, which is available at least as far back as numpy 1.15, according to the docs.

https://numpy.org/doc/1.15/reference/constants.html#numpy.nan


schloerke:
Closing this PR in favor of #1468 (already merged) and #1469

PR #1468, opened 2024-06-18
chore: Use `np.nan` instead of `np.NaN`

schloerke:
Related: https://github.com/numpy/numpy/pull/24376 ; (Removes `np.NaN` for v2.0.0 release)


jcheng5:
Coming in with little context here but do we actually care if it’s numpy >= 2?

PR #1484, opened 2024-06-25
Add return type for `_task()`

wch:
This fixes an issue that is flagged by pyright 1.1.369.


wch:
I think the error we were seeing with pyright 1.1.369 was due to a bug: https://github.com/microsoft/pyright/issues/8229

PR #1485, opened 2024-06-27
fix(tests): dynamically determine the path to the shiny app

karangattu:
Use approach that allows for dynamic path resolution based on runtime conditions

### DIFF
--- a/shiny/pytest/_fixture.py
+++ b/shiny/pytest/_fixture.py
@@ -30,8 +30,8 @@ def create_app_fixture(
     app
         The path to the Shiny app file.
 
-        If `app` is a `Path` or `PurePath` instance and `.is_file()` returns `True`, then this value will be used directly.
-        Note, file paths will be checked from where `pytest` was called, not from the calling function.
+        If `app` is a `Path` or `PurePath` instance and `Path(app).is_file()` returns `True`, then this value will be used directly.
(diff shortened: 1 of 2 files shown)


shiny/pytest/_fixture.py
```diff
@@ -29,13 +29,22 @@ def create_app_fixture(
     ----------
     app
         The path to the Shiny app file.
+
+        If `app` is a `Path` or `PurePath` instance and `.is_file()` returns `True`, then this value will be used directly.
+        Note, file paths will be checked from where `pytest` was called, not from the calling function.
```
schloerke, line 34:
```suggestion
        Note, `app`'s file path will be checked from where corresponding `pytest` test is collected, not necessarily where `create_app_fixture()` is called.
```

shiny/pytest/_fixture.py
```diff
@@ -29,13 +29,22 @@ def create_app_fixture(
     ----------
     app
         The path to the Shiny app file.
+
+        If `app` is a `Path` or `PurePath` instance and `.is_file()` returns `True`, then this value will be used directly.
```
schloerke, line 33:
```suggestion
        If `app` is a `Path` or `PurePath` instance and `Path(app).is_file()` returns `True`, then this value will be used directly.
```

PR #1486, opened 2024-06-28
fix: support static files on pyodide / py.cafe under a prefix

maartenbreddels:
When a shiny app is running under a prefix (root_path in ASGI terms), the static files are not served correctly under pyodide. This is because the ASGI path includes the root_path, and the root_path should be removed.
Starlette 0.33 and 0.34 however, did not set root_path correctly, and in those cases, we have to rely on route_path.

Related discussions:
 https://github.com/encode/starlette/pull/2400
 https://github.com/encode/starlette/discussions/2361

A related fix we had in Solara:
 https://github.com/widgetti/solara/pull/413/

But this fix also did not seem to work for our situation at https://py.cafe

I logged the output of the relevant entries in the scope dict, together with the starlette version:

```
starlette 0.32.0 {'path': '/strftime-min.js', 'root_path': '/_app/lib/strftime-0.9.2'}
starlette 0.33.0 {'path': '/_app/lib/strftime-0.9.2/strftime-min.js', 'root_path': '', 'route_root_path': '/lib/strftime-0.9.2', 'route_path': '/strftime-min.js'}
starlette 0.34.0 {'path': '/_app/lib/strftime-0.9.2/strftime-min.js', 'root_path': '', 'route_root_path': '/lib/strftime-0.9.2', 'route_path': '/strftime-min.js'}
starlette 0.35.0 {'path': '/_app/lib/strftime-0.9.2/strftime-min.js', 'root_path': '/_app/lib/strftime-0.9.2'}
starlette 0.36.0 {'path': '/_app/lib/strftime-0.9.2/strftime-min.js', 'root_path': '/_app/lib/strftime-0.9.2'}
starlette 0.37.2 {'path': '/_app/lib/strftime-0.9.2/strftime-min.js', 'root_path': '/_app/lib/strftime-0.9.2'}
```

This was using a similar situation as on py.cafe:
```python
....
routes = [
    Mount('/static', app=app_static),
    Mount('/_app', app=app_shiny)
]

app = Starlette(routes=routes)
```

Which led me to the following fix, making shiny work under pyodide in combination with a prefix with the above mentioned versions of starlette.

### DIFF
--- a/shiny/http_staticfiles.py
+++ b/shiny/http_staticfiles.py
@@ -11,6 +11,8 @@ not true when running in native Python, we want to be as safe as possible.
 
 from __future__ import annotations
 
+import re
+
 __all__ = (
     "StaticFiles",
     "FileResponse",
@@ -53,7 +55,9 @@ else:
         async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
             if scope["type"] != "http":
                 raise AssertionError("StaticFiles can't handle non-http request")
-            path = scope["path"]
+            # following starlette >=0.33, tested to be compatible with 0.32-0.37.2
+            root_path = scope.get("route_root_path", scope.get("root_path", ""))
+            path = scope.get("route_path", re.sub(r"^" + root_path, "", scope["path"]))
             path_segments = path.split("/")
             final_path, trailing_slash = _traverse_url_path(self.dir, path_segments)
             if final_path is None:


wch:
Thanks for the PR!

I had previously tried updating the version of starlette included with shinylive but that caused shiny to break, and this change appears to fix that issue.

PR #1505, opened 2024-07-09
docs(ui.autocomplete): Update autocomplete docs

schloerke:


shiny/ui/_input_text.py
```diff
@@ -37,11 +37,11 @@ def input_text(
     placeholder
         A hint as to what can be entered into the control.
     autocomplete
-        Whether to enable browser autocompletion of the text input (default is ``None``).
-        If `None`, then it will use the browser's default behavior. Other possible values
+        Whether to enable browser autocompletion of the text input.
+        If `None`, then it will use the browser's default behavior. Possible values
```
karangattu, line 41:
```suggestion
        If `None`, then it will use the browser's default behavior. Some values
```

karangattu (approved):
LGTM

PR #1509, opened 2024-07-09
`Chat.messages()` now keeps system messages

cpsievert:
Before this PR, `chat.messages()` would discard system messages if they didn't fit within the `token_limits`. Now, system messages are always kept. And, if the system messages don't fit sensibly within the `token_limits`, you now get an informative error to increase the limit


shiny/ui/_chat.py
```diff
@@ -738,34 +740,57 @@ def _store_message(
...
+        max_tokens -= system_token_count
+
         messages2: list[StoredMessage] = []
         for i, m in enumerate(reversed(messages)):
-            if sum(token_counts[-i - 1 :]) > max_tokens:
-                break
-            messages2.append(m)
+            if m["role"] == "system":
+                messages2.append(m)
+            elif sum(token_counts[-i - 1 :]) <= max_tokens:
```
wch, line 782: I think that `token_counts[-i - 1 :]` will result in incorrect indexing in some edge cases. The `messages` data structure could have a system message that is not the first message. In the code above, where it does the first pass iterating over `messages`, it appends to `token_counts` only for non-system messages.

However, in this loop where it iterates over `messages`, it indexes into `token_counts` using `i`, where the `i` also is incremented with system messages -- but the `token_counts` list doesn't include counts for system messages.

To fix this, I think it would be simpler to replace `max_tokens` with `remaining_non_system_tokens`, and subtract from it in each iteration of the loop.

shiny/ui/_chat.py
```diff
@@ -738,34 +740,57 @@ def _store_message(
...
+        system_messages: list[str] = []
+        other_messages: list[str] = []
+        for m in messages:
+            count = m["token_count"]
+            # Count can be None if the tokenizer is None
+            if count is None:
+                return messages
+            content = m["content_server"]
+            if m["role"] == "system":
+                system_messages.append(content)
```
wch, line 760: It looks like `system_messages` and `other_messages` are only used for their `len()`, below, so these can be replaced with simple counters instead of lists.

And if you do that, `content` can be removed entirely.

wch (changes requested):

shiny/ui/_chat.py
cpsievert: Thanks! Addressed in [215109d](https://github.com/posit-dev/py-shiny/pull/1509/commits/215109d8f4743f646f1a1e65e5a75cc7f049758a)

wch (approved):

PR #1515, opened 2024-07-10
Follow up to #1508: move dotenv complexity to `app_utils.py` module

cpsievert:
This makes the `app.py` file for the `Chat` examples a bit more readable, and also doesn't crash the app if `{dotenv}` isn't installed.

### DIFF
--- a/examples/chat/enterprise/aws-bedrock-anthropic/app.py
+++ b/examples/chat/enterprise/aws-bedrock-anthropic/app.py
@@ -4,18 +4,17 @@
 # To get started, follow the instructions at https://aws.amazon.com/bedrock/claude/
 # as well as https://github.com/anthropics/anthropic-sdk-python#aws-bedrock
 # ------------------------------------------------------------------------------------
-from pathlib import Path
-
 from anthropic import AnthropicBedrock
-from dotenv import load_dotenv
 
 from shiny.express import ui
+
+# In Shiny Core, do `from app_utils import load_dotenv`
+from .app_utils import load_dotenv
 
 # Either explicitly set the AWS environment variables before launching the app, or set
 # them in a file named `.env`. The `python-dotenv` package will load `.env` as
 # environment variables which can be read by `os.getenv()`.
-_ = load_dotenv(Path(__file__).parent / ".env")
-
+load_dotenv()
 llm = AnthropicBedrock(
     # aws_secret_key=os.getenv("AWS_SECRET_KEY"),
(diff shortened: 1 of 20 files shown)


examples/chat/enterprise/aws-bedrock-anthropic/app_utils.py
```diff
@@ -0,0 +1,27 @@
...
+
+def load_dotenv(dotenv_path: os.PathLike[str] = env_file, **kwargs) -> None:
+    """
+    A convenience wrapper around `dotenv.load_dotenv` that warns if `dotenv` is not installed.
+    It also returns `None` to make it easier to ignore the return value.
+    """
+    try:
+        import dotenv
+
+        _ = dotenv.load_dotenv(dotenv_path=dotenv_path, **kwargs)
```
wch, line 16:
```suggestion
        dotenv.load_dotenv(dotenv_path=dotenv_path, **kwargs)
```

examples/chat/enterprise/aws-bedrock-anthropic/app_utils.py
```diff
@@ -0,0 +1,27 @@
...
+        return None
```
wch, line 17: Small nit: I suggest removing both `return` statements, or at least changing this to just `return` for consistency with the one below.

examples/chat/enterprise/aws-bedrock-anthropic/app.py
```diff
@@ -4,18 +4,17 @@
...
 # ------------------------------------------------------------------------------------
-from pathlib import Path
-
 from anthropic import AnthropicBedrock
-from dotenv import load_dotenv
 
 from shiny.express import ui
 
+# In Shiny Core, do `from app_utils import load_dotenv`
+from .app_utils import load_dotenv
```
wch, line 12: I think this should work as an absolute import, right? At least after the changes in #1464.

```suggestion
from app_utils import load_dotenv
```

wch (changes requested):

examples/chat/enterprise/aws-bedrock-anthropic/app.py
cpsievert: Yea, I was going back and forth on this. It does work as an absolute import, but pyright/pylance complains about it. Thinking about it a bit more, I think I'd prefer this?

```python
from app_utils import load_dotenv  # pyright: ignore[reportMissingImports]
```

PR #1518, opened 2024-07-11
tests(deploys): separate deploy test workflow from pytest workflow

karangattu:
We want to only run the **Deploy pre-checks** on every commit in a PR.
We will separate the deploy tests into its own workflow file from `pytest.yaml`


schloerke (approved):

schloerke:
The failures are unrelated to this PR. 

Merging to `main`

PR #1522, opened 2024-07-11
Add `format` parameter to `Chat`'s `.messages()` method

cpsievert:


shiny/ui/_chat_provider_types.py
```diff
@@ -0,0 +1,123 @@
...
+        return as_anthropic_message(message)
+    if format == "google":
+        return as_google_message(message)
+    if format == "langchain":
+        return as_langchain_message(message)
+    if format == "openai":
+        return as_openai_message(message)
+    if format == "ollama":
+        return as_ollama_message(message)
+    raise ValueError(f"Unknown format: {format}")
```
cpsievert, line 65: Take a strategies pattern here so that one could (in theory) add more formats

wch (approved):
There seems to be a linting error, but otherwise looks good.

shiny/ui/_chat_provider_types.py
cpsievert: After chatting with Winston, we're gonna kick this down the road

PR #1530, opened 2024-07-15
Ensure first message is user message for anthropic

wch:
For Anthropic, the first message in a chat must have `"role": "user"`; otherwise the API will return an error.

This PR ensures that when the messages are trimmed, the first message is a `user` message.

One possibly confusing thing: Suppose you initialize the chat with something like this:

```py
chat = ui.Chat(
    "chat",
    messages=[
        {"role": "assistant", "content": "Arrgh! I am a chatbot that speaks like a pirate. How may I help ye?"},
    ],
)
```

Then when the user enters a message and the messages get sent to the LLM, this first message will get trimmed off, and so the LLM will not know that it's supposed to respond like a pirate. The problem is that the app author won't have any indication that this first message is getting trimmed, so if they expect anthropic to behave like openai (where the first message can be `assistant`), they will likely be confused as to why the first message would be ignored.

### DIFF
--- a/shiny/ui/_chat.py
+++ b/shiny/ui/_chat.py
@@ -864,6 +864,7 @@ class Chat:
             )
 
         messages2: list[StoredMessage] = []
+        n_other_messages2: int = 0
         for m in reversed(messages):
             if m["role"] == "system":
                 messages2.append(m)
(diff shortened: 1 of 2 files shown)


shiny/ui/_chat.py
```diff
@@ -872,6 +873,11 @@ def _trim_messages(
             if remaining_non_system_tokens >= 0:
                 messages2.append(m)
 
+        if format == "anthropic":
+            # For anthropic, the first message must be a user message.
+            while messages2[-1]["role"] != "user":
+                messages2.pop()
```
cpsievert, line 879: It's technically possible that `messages2` is of length 0 here, so we'll need to be more careful about that. 

```python
from shiny import reactive
from shiny.express import ui

chat = ui.Chat(id="chat")
chat.ui()

@reactive.effect
def _():
    print(chat.messages(format="anthropic"))
```

PR #1532, opened 2024-07-15
Add `chat.messages(format='internal')`

wch:
This changes PR changes `chat.messages()` so that:

- A value for `format` is required. Previously the default value was `MISSING`.
- To get the internal/native format, the user calls it with `chat.messages(format="internal")` (instead of `format=MISSING`, which was the previously the default).

Previously, the default `MISSING` would result in it returning the messages in the internal `ChatMessage` format.  However, this format is actually the least likely format that a user would want -- it is used internally in Shiny, but is not directly usable by any LLMs out there. I think that we don't want to have a default format which is not useful.

Note that this builds on PR #1530.

Also, I'm open to ideas other than `"internal"` for the name.

### DIFF
--- a/shiny/ui/_chat.py
+++ b/shiny/ui/_chat.py
@@ -466,7 +466,7 @@ class Chat:
 
         messages = self._messages()
         if token_limits is not None:
-            messages = self._trim_messages(messages, token_limits)
+            messages = self._trim_messages(messages, token_limits, format)
 
         res: list[ChatMessage | ProviderMessage] = []
         for i, m in enumerate(messages):
@@ -827,7 +827,8 @@ class Chat:
     @staticmethod
     def _trim_messages(
         messages: tuple[StoredMessage, ...],
-        token_limits: tuple[int, int] = (4096, 1000),
+        token_limits: tuple[int, int],
+        format: MISSING_TYPE | ProviderMessageFormat,
     ) -> tuple[StoredMessage, ...]:
 
         n_total, n_reserve = token_limits
@@ -871,6 +872,11 @@ class Chat:
             remaining_non_system_tokens -= count
             if remaining_non_system_tokens >= 0:
                 messages2.append(m)
+
+        if format == "anthropic":
+            # For anthropic, the first message must be a user message.
+            while messages2[-1]["role"] != "user":
+                messages2.pop()
 
         messages2.reverse()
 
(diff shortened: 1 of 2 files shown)


shiny/ui/_chat.py
```diff
@@ -406,7 +406,7 @@ def messages(
     def messages(
         self,
         *,
-        format: MISSING_TYPE | ProviderMessageFormat = MISSING,
+        format: ProviderMessageFormat | Literal["internal"],
```
cpsievert, line 409: Was it intentional to require a value here? I'd prefer if it wasn't necessary since the `"internal"` format works for most formats

```suggestion
        format: ProviderMessageFormat | Literal["internal"] = "internal",
```

shiny/ui/_chat.py
```diff
@@ -872,6 +873,11 @@ def _trim_messages(
             if remaining_non_system_tokens >= 0:
                 messages2.append(m)
 
+        if format == "anthropic":
+            # For anthropic, the first message must be a user message.
+            while messages2[-1]["role"] != "user":
+                messages2.pop()
```
cpsievert, line 879: It's technically possible that `messages2` is of length 0 here, so we'll need to be more considerate of that. Also, for similar reasons, it's also possible that there's no user message.

```python
from shiny import reactive
from shiny.express import ui

chat = ui.Chat(id="chat")
chat.ui()

@reactive.effect
def _():
    print(chat.messages(format="anthropic"))
```

shiny/ui/_chat.py
wch: Oh wait, I think I misunderstood what the default format was -- I thought it was the actual internal format that the `Chat` class uses, but that's not right. The internal format is `StoredMessage`, not `ChatMessage`. That's why I thought it wouldn't make sense to return that format by default.

So to summarize, prior to this PR, the default format was `ChatMessage`, which is sort of a lowest-common-denominator format that has `role` and `content`. 

So it might make sense to just drop this PR entirely. That said, it could still be useful to be able to access the `StoredMessage`s in the chat object. That's something I actually wanted to do when I was trying to get a handle on the sizes of messages being sent to the LLM (but then I got sidetracked before I went in depth on it).

cpsievert: Ah, I see. I'd be interested to know more about why you want access to `StoredMessage`.

I'll close this PR and create another one for the `format="anthropic"` thing

PR #1540, opened 2024-07-16
`shiny.experimental`: Remove deprecated functions

gadenbuie:
Removes the deprecated functions from `shiny.experimental` that have largely made their way into the main `shiny` package.


schloerke (approved):
🥳

Issue #1498, opened 2024-07-05
Data_frame with two tables in edit mode, hangs browser. Potential Bug?


maxmoro:
Here below is a short example that hangs my browser.
I've two panels, with a large dataset (20cols, 5000 rows). 
- Switching between the panels ->  no problem
- Editing cells in one panel -> no problem
- Editing cells in one panel, then switching to the other panel -> Browser Hangs, CPU usage to the roof.

I'm attaching the code to create the fake dataset and the code that creates the problem.

I tried running in  Shiny Live, sometimes it hangs and sometimes it takes a long time to switch the tab. I can't share the Shiny Live link as it is too long. (and can't use the `faker` library on Shiny Live).

I'm using **shiny==0.10.2**

Please try by 
1) create the large fake dataset, then 
2) run the app.py

# Large Fake dataset creation
The dataset is **fake**. It is created using the [Faker](https://faker.readthedocs.io/en/master/) library.

```
from faker import Faker
import pandas as pd
#penguins = load_penguins()
def fakeDS():
    fake = Faker()
    df1 = pd.DataFrame([{
            'Name': fake.name(),
            'Email': fake.email(),
            'Address': fake.address(),
            'Job': fake.job(),
            'Company': fake.company(),
            'Phone Number': fake.phone_number(),
            'City': fake.city(),
            'Country': fake.country(),
            'Text': fake.text(),
            'Date': fake.date(),
            'Latitude': fake.latitude(),
            'Longitude': fake.longitude(),
            'CreditCardNumber': fake.credit_card_number(),
            'CreditCardProvider': fake.credit_card_provider(),
            'CreditCardExpire': fake.credit_card_expire(),
            'Currency': fake.currency_code(),
            'Price': fake.pydecimal(left_digits=3, right_digits=2, positive=True),
            'Password': fake.password(),
            'URL': fake.url(),
        }
        for _ in range(5000)
    ])
    df1.to_json('fakeds.json')
    df2 = pd.read_json('fakeds.json')
```

# App.py that shows the issue
```
from pathlib import Path
from palmerpenguins import load_penguins
from shiny import App, render, ui, reactive, Outputs
import pandas as pd
infile = Path(__file__).parent / 'fakeds.json'
print("reading")
df1=pd.read_json(infile)
df2=df1.copy()
print("done reading")

app_ui = ui.page_fluid(
    ui.page_navbar(  
        ui.nav_panel(
            "A",
            ui.h2("Fake data 1"),
[...]

PR #1550, opened 2024-07-18
bug(data frame): Fix scrolling in fillable cards

### LINKED ISSUE #1549: Horizontal scroll no longer works with @render.data_frame
In the tips dashboard, there's a data frame that can easily be wider than the card it's in. Until 6611277da634503e82adcdee3aa1dd67d7bf0a87, there was a horizontal scrollbar; since that commit, there isn't.

The tip dashboard is at `shiny/templates/app-templates/dashboard-tips/app-express.py`

jcheng5:
Fixes #1549.

Without this fix, the data frame in the tips dashboard example is sized to be as tall as the (non-virtualized) table. This causes the summary line ("Viewing rows _x_ through _y_ of _z_") to be covered, as well as the horizontal scroll bar on the data frame.

Working:
image

Broken:
image

The change seems to have happened when Tanstack was updated. I set a breakpoint on these lines:
https://github.com/posit-dev/py-shiny/blob/798d02c252d02eaaf7bb876522e51eceeb4f83b0/js/data-frame/index.tsx#L653-L657

For both working and non-working code, both scrollHeight and clientHeight are undefined the first time through; this causes `.scrolling` to be applied. The second time through, for the working, scrollHeight and clientHeight are different and I can see cells in the table (while the debugger is paused); but for the non-working, scrollHeight and clientHeight are the same and there are no cells in the table. **This causes `.scrolling` to be removed.** And once removed, it won't be applied in the future, because `scrollHeight` and `clientHeight` will always be the same value--because it's not scrolling.

The fix is to use a layout effect (runs after DOM nodes are inserted/mutated but before the browser has had a chance to paint) to temporarily apply scrolling, measure, and then remove scrolling if appropriate. I'm a little worried that temporarily applying scrolling is too violent, but I don't know what else to do, and it appears to be flicker-free. In the common case where scrolling is correctly turned on, I think it will be very quick, as no CSS classes will actually be applied or removed.


schloerke:
This fix still gives expected behavior for https://github.com/posit-dev/py-shiny/issues/1498. 👍

schloerke (approved):

PR #1553, opened 2024-07-18
ci: Adjust pytest verbosity output levels

schloerke:
Return the summary output back to normal size.  Add [`verbosity_test_cases=2`](https://docs.pytest.org/en/stable/reference/reference.html#confval-verbosity_test_cases) and [`verbosity_assertions=2`](https://docs.pytest.org/en/stable/reference/reference.html#confval-verbosity_assertions) to increase status reporting and error debugging.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,3 +3,5 @@ asyncio_mode=strict
 testpaths=tests/pytest/
 ; Note: Browsers are set within `./Makefile`
 addopts = --strict-markers --durations=6 --durations-min=5.0 --numprocesses auto --tracing=retain-on-failure --video=retain-on-failure
+verbosity_test_cases=2
+verbosity_assertions=2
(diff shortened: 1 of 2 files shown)


gadenbuie (approved):
LGTM! (although it's hard to tell what it looks like because we didn't have any failures)

pytest.ini
```diff
@@ -3,3 +3,5 @@ asyncio_mode=strict
 testpaths=tests/pytest/
 ; Note: Browsers are set within `./Makefile`
 addopts = --strict-markers --durations=6 --durations-min=5.0 --numprocesses auto --tracing=retain-on-failure --video=retain-on-failure
+verbosity_test_cases=2
+verbosity_assertions=2
```
schloerke, line 7: Short summary is currently too large

```suggestion
```

PR #1557, opened 2024-07-18
bug(data frame filter): When the filter is reset, be sure to use a "" value for the input

schloerke:
Noticed when calling `.update_filter(None)`, string column filters were not being reset. The filtering was actually reset, however the value presented in the input box was not reset.


schloerke:
Merge queue successfully ran on rerun. Squash merging

PR #1569, opened 2024-07-25
bug(input_text_area): Auto resized text areas resize on visibility change

schloerke:
From https://discord.com/channels/1109483223987277844/1265694689378697227/1266038169053171712

When the input text area would be displayed after an accordion expanded or a new tab was displayed, the text area would have height of 0 as the original scroll height was 0 (as it was hidden). 

Now, if the scroll height is 0 an intersection observer is used until the text area is slightly on screen, then we update the height. To signal that we should perform this default behavior, a value of `rows=0` is used. This is not a possible value, so `.getAttribute("rows")` must be used to check for the `0` value. (Otherwise, `target.rows` will return 2, the default value.)


js/text-area/textarea-autoresize.ts
```diff
@@ -19,9 +19,41 @@ function onDelegatedEvent(
...
+
+    // If `.getAttribute("rows")` is set to 0 (as `target.rows` is always > 0 and
+    // invalid values use the default value of `2`), then we can wait for the textarea
+    // to become visible and call `update_height` again as the scroll height should not
+    // be 0.
+    //
+    // Note: `rows` is only set to `0` from the python code when the textarea is
+    // autoresized and no `rows` value is given.
+    const targetRows = target.getAttribute("rows");
+    if (targetRows == "0") {
```
cpsievert, line 37: I can't seem to follow why it's beneficial to use `rows=0` as an additional signal to resize (when visible) instead of always updating the height?

Also, wouldn't this mean that if a user specifies `rows=n`, then you don't get the same resize-on-visible behavior?

shiny/ui/_input_text.py
```diff
@@ -163,7 +163,10 @@ def input_text_area(
     if autoresize:
         classes.append("textarea-autoresize")
         if rows is None:
-            rows = 1
+            # Use a placeholder of `_auto_` (an invalid value) to signal that the textarea
+            # should autoresized on load
+            # rows = "_auto_"
```
cpsievert, line 168: Comment should be updated

shiny/ui/_input_text.py
```diff
@@ -163,7 +163,10 @@ def input_text_area(
...
+            rows = 0
```
cpsievert, line 169: If we do this, I think we'll want to make the same change to `chat.ts` 

(Although, ideally, I'd like to avoid using `rows` as a signal for resize-on-visible)

js/text-area/textarea-autoresize.ts
schloerke: You would not get it.  But it would resize.

I'm thinking we need to add an extra attribute to avoid the confusion.

schloerke: Yes. Removed.

It should always be put into the intersection observer if scrollHeight is 0. Removing the `rows=0` init value change.

shiny/ui/_input_text.py
schloerke: It has been removed. It will still auto resize once visible. So I don't know where chat.ts will need to be updated

js/text-area/textarea-autoresize.ts
```diff
@@ -19,9 +19,33 @@ function onDelegatedEvent(
...
+      if (entry.intersectionRatio > 0) {
+        // Stop observing the target
+        observer.unobserve(entry.target);
+
+        // Update the height of the textarea
+        update_height(entry.target as HTMLTextAreaElement);
+      }
+    });
+  });
+  observer.observe(target);
```
cpsievert, line 48: IIRC, it's surprisingly expensive to create an observer -- is there a way we can create just one that observes all the relevant targets?

schloerke: I had figured, but I didn't know they were that bad.

Related: https://www.bennadel.com/blog/3954-intersectionobserver-api-performance-many-vs-shared-in-angular-11-0-5.htm

I converted the multiple intersection observers to a single observer iff it is needed.

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Require shinyswatch >= 0.7.0 and updated examples accordingly. (#1558)
 
+* Fixed bug where an auto resized input text area was not initially auto resized if the element was hidden from view (e.g. in a closed accordion or a hidden tab). (#1560)
```
cpsievert, line 20:
```suggestion
* `input_text_area(autoresize=True)` now resizes properly even when it's not visible when initially rendered (e.g. in a closed accordion or a hidden tab). (#1560)
```

cpsievert (approved):
Thanks!

PR #1570, opened 2024-07-25
feat: support DataFrames via narwhals

machow:
Currently, we have code in `_tbl_data.py`, which uses generic functions to handle DataFrame logic for both pandas and polars. While this works well, it requires that we basically maintain a tiny DataFrame translation layer. This PR starts work towards using narwhals to handle the pandas and polars logic.

The big advantage of narwhals is that it translates the polars API over different DataFrame types (e.g. pandas). Wiring it up means we use our polars logic, but can apply it to pandas, etc..

* @schloerke said he'll take this and run with it (feel free to blast anything; I'm happy to try a narwhals PR if anything is needed)

(Big thanks to @MarcoGorelli for pairing w/ @schloerke, @jcheng5, and I to work on the key pieces :)

### DIFF
--- a/shiny/render/_data_frame_utils/_tbl_data.py
+++ b/shiny/render/_data_frame_utils/_tbl_data.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from functools import singledispatch
 from typing import Any, List, Tuple, cast
 
+import narwhals.stable.v1 as nw
 from htmltools import TagNode
 
 from ..._typing_extensions import TypeIs
@@ -184,7 +185,7 @@ def _(col: PlSeries) -> FrameDtype:
 
     from ._html import col_contains_shiny_html
 
-    if col.dtype.is_(pl.String):
+    if col.dtype == pl.String():
         if col_contains_shiny_html(col):
             type_ = "html"
         else:
@@ -193,6 +194,30 @@ def _(col: PlSeries) -> FrameDtype:
         type_ = "numeric"
 
     elif col.dtype.is_(pl.Categorical()):
+        categories = col.cat.get_categories().to_list()
+        return {"type": "categorical", "categories": categories}
+    else:
+        type_ = "unknown"
+        if col_contains_shiny_html(col):
+            type_ = "html"
+
+    return {"type": type_}
+
+
+@serialize_dtype.register
+def _(col: nw.Series) -> FrameDtype:
+
+    from ._html import col_contains_shiny_html
+
+    if col.dtype == nw.String():
+        if col_contains_shiny_html(col):
+            type_ = "html"
+        else:
+            type_ = "string"
+    elif col.dtype.is_numeric():
+        type_ = "numeric"
+
+    elif col.dtype == nw.Categorical():
         categories = col.cat.get_categories().to_list()
         return {"type": "categorical", "categories": categories}
     else:
@@ -218,6 +243,8 @@ def _(data: PdDataFrame) -> FrameJson:
     return serialize_frame_pd(data)
 
 
+# TODO: test this
+@serialize_frame.register(nw.DataFrame)
 @serialize_frame.register
 def _(data: PlDataFrame) -> FrameJson:
     import json
@@ -308,6 +335,7 @@ def _(
         return data.iloc[indx_rows, indx_cols]
 
 
+@subset_frame.register(nw.DataFrame)
 @subset_frame.register
 def _(
     data: PlDataFrame,
@@ -321,7 +349,7 @@ def _(
         else slice(None)
     )
     indx_rows = rows if rows is not None else slice(None)
-    return data[indx_rows, indx_cols]
+    return data[indx_cols][indx_rows]
 
 
 # get_frame_cell -----------------------------------------------------------------------
@@ -341,9 +369,10 @@ def _(data: PdDataFrame, row: int, col: int) -> Any:
     )
 
 
+@get_frame_cell.register(nw.DataFrame)
 @get_frame_cell.register
 def _(data: PlDataFrame, row: int, col: int) -> Any:
-    return data[row, col]
+    return data.item(row, col)
 
 
 # shape --------------------------------------------------------------------------------
@@ -359,6 +388,7 @@ def _(data: PdDataFrame) -> Tuple[int, ...]:
     return data.shape
 
 
+@frame_shape.register(nw.DataFrame)
 @frame_shape.register
 def _(data: PlDataFrame) -> Tuple[int, ...]:
     return data.shape
@@ -377,6 +407,7 @@ def _(data: PdDataFrame) -> PdDataFrame:
     return data.copy()
 
 
+@copy_frame.register(nw.DataFrame)
 @copy_frame.register
 def _(data: PlDataFrame) -> PlDataFrame:
     return data.clone()
@@ -393,6 +424,7 @@ def _(data: PdDataFrame) -> List[str]:
     return data.columns.to_list()
 
 
+@frame_column_names.register(nw.DataFrame)
 @frame_column_names.register
 def _(data: PlDataFrame) -> List[str]:
     return data.columns
--- a/tests/pytest/test_render_data_frame.py
+++ b/tests/pytest/test_render_data_frame.py
@@ -1,3 +1,4 @@
+import narwhals as nw
 import pandas as pd
 import pytest
 
--- a/tests/pytest/test_render_data_frame_tbl_data.py
+++ b/tests/pytest/test_render_data_frame_tbl_data.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass
 from datetime import datetime
 from typing import Any, Union
 
+import narwhals.stable.v1 as nw
 import pandas as pd
 import polars as pl
 import polars.testing as pl_testing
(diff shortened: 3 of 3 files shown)


shiny/render/_data_frame_utils/_tbl_data.py
```diff
@@ -321,7 +349,7 @@ def _(
         else slice(None)
     )
     indx_rows = rows if rows is not None else slice(None)
-    return data[indx_rows, indx_cols]
+    return data[indx_cols][indx_rows]
```
MarcoGorelli, line 352: we just pushed out a new Narwhals version which supports passing a tuple of sequences to `__getitem__`, so if you upgrade your Narwhals version then `data[indx_rows, indx_cols]` should work

shiny/render/_data_frame_utils/_pandas.py
```diff
@@ -119,8 +122,10 @@ def serialize_pd_dtype(
         }
     elif t in {"datetime64", "datetime"}:
         t = "datetime"
-    elif t in {"timedelta", "timedelta64"}:
-        t = "timedelta"
+
+    # # Disable timedelta as narwhals does not support it
+    # elif t in {"timedelta", "timedelta64"}:
+    #     t = "timedelta"
```
MarcoGorelli, line 128: quick comment while scanning through this, but the equivalent is [Duration](https://narwhals-dev.github.io/narwhals/api-reference/dtypes/#narwhals.dtypes.Duration)

schloerke: Thank you!!

shiny/render/_data_frame.py
```diff
@@ -327,17 +333,47 @@ def self_cell_patches() -> list[CellPatch]:
         self.cell_patches = self_cell_patches
 
         @reactive.calc
-        def self_data() -> DataFrameLikeT:
+        def self__value_data() -> IntoDataFrameT:
```
jcheng5, line 336: This can be a regular method

shiny/render/_data_frame.py
```diff
@@ -327,17 +333,47 @@ def self_cell_patches() -> list[CellPatch]:
...
+        self._value_data = self__value_data
+
+        @reactive.calc
+        def self__nw_data() -> DataFrame[IntoDataFrameT]:
+            return as_data_frame(self._value_data())
+
+        self._nw_data = self__nw_data
+
+        @reactive.calc
+        def self__data_is_nw() -> bool:
```
jcheng5, line 353: This can be a regular method too

shiny/render/_data_frame.py
```diff
@@ -327,17 +333,47 @@ def self_cell_patches() -> list[CellPatch]:
...
+            return isinstance(self._value_data(), DataFrame)
+
+        self._data_is_nw = self__data_is_nw
+
+        # @overload
+        # def
+        @reactive.calc
+        def self__nw_data_to_original_type_fn():
```
jcheng5, line 361: Remove this outer layer entirely, make the inside a helper method

shiny/render/_data_frame.py
```diff
@@ -327,17 +333,47 @@ def self_cell_patches() -> list[CellPatch]:
...
+                    return cast(IntoDataFrameT, nw_data)
+                return data_frame_to_native(nw_data)
+
+            return _
+
+        self._nw_data_to_original_type_fn = self__nw_data_to_original_type_fn
+
+        @reactive.calc
+        def self_data() -> IntoDataFrameT:
+            return data_frame_to_native(self._nw_data())
```
jcheng5, line 375: Use the helper method that is immediately above

shiny/render/_data_frame.py
```diff
@@ -55,22 +56,19 @@
...
-    "DataTable[DataFrameLikeT]",
-]
-DataFrameValue = Union[None, DataGrid[DataFrameLikeT], DataTable[DataFrameLikeT]]
-
 
 @add_example()
-class data_frame(Renderer[DataFrameResult[DataFrameLikeT]]):
+class data_frame(
+    Renderer[
+        Union[None, IntoDataFrameT, DataGrid[IntoDataFrameT], DataTable[IntoDataFrameT]]
```
jcheng5, line 63:
```suggestion
        Union[IntoDataFrameT, DataGrid[IntoDataFrameT], DataTable[IntoDataFrameT]]
```

shiny/render/_data_frame.py
```diff
@@ -181,7 +185,7 @@ class data_frame(Renderer[DataFrameResult[DataFrameLikeT]]):
     Reactive value of the data frame's edits provided by the user.
     """
 
-    data: reactive.Calc_[DataFrameLikeT]
+    data: reactive.Calc_[IntoDataFrameT]
```
jcheng5, line 188: The documentation for these isn't quite rendering correctly; they're showing up as attributes, and clicking them doesn't go anywhere.

shiny/render/_data_frame_utils/_tbl_data.py
```diff
@@ -43,356 +38,299 @@
...
 
 # subset_frame -------------------------------------------------------------------------
-def subset_frame__typed(
-    data: DataFrameLikeT, *, rows: RowsList = None, cols: ColsList = None
-) -> DataFrameLikeT:
-    return cast(DataFrameLikeT, subset_frame(data, rows=rows, cols=cols))
-
-
-@singledispatch
 def subset_frame(
```
jcheng5, line 267:
```
# subset_frame -------------------------------------------------------------------------
def subset_frame(
    data: DataFrameT,
    *,
    rows: RowsList = None,
    cols: ColsList = None,
) -> DataFrameT:
    """Return a subsetted DataFrame, based on row positions and column names.

    Note that when None is passed, all rows or columns get included.
    """

    # Note that this type signature assumes column names are strings things.
    # This is always true in Polars, but not in Pandas (e.g. a column name could be an
    # int, or even a tuple of ints)

    if rows is None and cols is None:
        return data
    row_idx = slice() if rows is None else list(rows)
    col_names = data.columns if cols is None else [data.columns[col] if isinstance(col, int) else col for col in cols]
    if len(col_names) == 0:
        return data[[], []]
    return data[row_idx, col_names]
```

shiny/render/_data_frame_utils/_tbl_data.py
```diff
@@ -43,356 +38,299 @@
...
+        else:
+            # Be sure rows is a list, not a tuple. Tuple
+            return data[list(rows), col_names]
 
 
-@get_frame_cell.register
-def _(data: PlDataFrame, row: int, col: int) -> Any:
-    return data[row, col]
+# # get_frame_cell -----------------------------------------------------------------------
+def get_frame_cell(data: DataFrame[Any], row: int, col: int) -> Any:
```
jcheng5, line 309: Remove the very simple functions and just inline. We're not afraid to tightly bound to Narwhals' (stable) API

shiny/render/_data_frame_utils/_tbl_data.py
schloerke: While I like the shorter code more, I ran into more issues that it was worth. 

It was difficult to support `slice()` or `slice(None)` as narwhals didn't appreciate it.

Keeping code as is

pyproject.toml
```diff
@@ -46,6 +47,8 @@ dependencies = [
     "prompt-toolkit;platform_system!='Emscripten'",
     "python-multipart>=0.0.7;platform_system!='Emscripten'",
     "setuptools;python_version>='3.12'",
+    "narwhals>=1.9.0,<1.10.0",
```
MarcoGorelli, line 50: Our intention with the backwards compatibility policy is that it shouldn't be necessary to set an upper bound, as you're using `narwhals.stable.v1` https://narwhals-dev.github.io/narwhals/backcompat/ 

In tomorrow's release we'd like to add `dtypes` and `typing` to `stable.v1` - with that, would you confident enough in not setting an upper-bound?
1.10.0 would likely come out next week, so this'd  be really tight 😄 

If another project were to adopt Narwhals and set - say - `1.10` as the minimum version, that'd already be enough to enter "dependency hell" 🔥

MarcoGorelli (commented):
Amazing effort here, and thanks the reports you've issued to Narwhals along the way! 🤗 

I've just got a question about the upper bound

pyproject.toml
schloerke: Absolutely!

I got bit pretty hard on the 1.9 narwhals release and slightly on 1.8. But both were dtype related. 

If they move to stable, I'm very happy to remove the upper bound on the narwhals version 😊

schloerke: Thank you again for the blazing fast turn around on all of the found issues. Amazing!! 💙

MarcoGorelli: thanks! cool, will do so tomorrow morning (I have a policy to only make releases in the morning, so if there's any urgent issues I can be awake to address them 😄 it's now nearly 10 pm where I live)

we'll add py-shiny to our CI to make sure we don't break anything https://github.com/narwhals-dev/narwhals/issues/1127

schloerke: Resolved in https://github.com/posit-dev/py-shiny/pull/1717

PR #1593, opened 2024-07-30
Delay sending of chat UI messages until reactive graph is flushed

### LINKED ISSUE #1592: Starting `Chat()` messages don't appear when ui is dynamically rendered
A minimal example (notice how `"A starting message"` doesn't appear on page load)

```python

chat = ui.Chat(id="chat", messages=["A starting message"])

@render.ui
def chat_output():
    return chat.ui()
```

cpsievert:
Closes #1592


shiny/ui/_chat.py
```diff
@@ -641,7 +641,9 @@ async def _send_append_message(
 
         # print(msg)
 
-        await self._send_custom_message(msg_type, msg)
+        # When streaming (i.e., chunk is thruthy), we can send messages immediately
```
cpsievert, line 644:
```suggestion
        # When streaming (i.e., chunk is truthy), we can send messages immediately
```

PR #1607, opened 2024-08-06
Chat tweaks

jcheng5:
Chat input is rendered like plain text today. With this change, it becomes Markdown, same as the output except raw HTML tags are escaped. I think this is pretty nice, especially for code, but if we want to be more like ChatGPT we could treat it as plaintext but preserve breaks. (Edit: @wch suggests we make this configurable)

I also made a subtle but important change to the NotifyException, it now chains the underlying exception (`raise x from y`). We should do a sweep sometime to make sure we do this everywhere; it's so hard to debug without being able to trace back to the original exception.


jcheng5:
The failing tests are real, at least some of them. Need to update the tests probably.

shiny/ui/_chat.py
```diff
@@ -585,7 +585,9 @@ async def _append_message_stream(self, message: AsyncIterable[Any]):
         id = _utils.private_random_id()
 
         empty = ChatMessage(content="", role="assistant")
-        await self._append_message(empty, chunk="start", stream_id=id)
+        async for msg in message:
+            await self._append_message(msg, chunk="start", stream_id=id)
+            break
```
cpsievert, line 590: This changes mean that the stream is now iterated over (and sent) *twice*. Is that intentional?

jcheng5: The first for loop waits for the first message and then breaks. The next for loop sees all the rest of the messages.

```python
>>> "hello".__iter__()
<str_iterator object at 0x104882920>
>>> it = "hello".__iter__()
>>> for c in it: print(c)
...
h
e
l
l
o
>>> for c in it: print(c)
...
>>>
```

Although now looking at this again, this is probably much more confusing looking than it needs to be, rather than just having a single loop that keeps a flag of whether it's sent any values yet or not.

cpsievert: Ah, sorry, misread the intention. 

> a single loop that keeps a flag of whether it's sent any values yet or not.

That or perhaps use `next()`?

js/chat/chat.ts
```diff
@@ -371,10 +395,12 @@ class ChatContainer extends LightElement {
       lastMessage.removeAttribute("is_streaming");
       this.#finalizeMessage();
       return;
+    } else {
+      lastMessage.setAttribute("is_streaming", "");
+      if (!message.chunk_type) {
+        lastMessage.setAttribute("content", message.content);
+      }
```
cpsievert, line 402: I'm a little confused why this `if` is here. Because it is, the starting message will get dropped. I think for some providers that's OK, but IIRC some models like Gemini start with non-trivial content on the 1st iteration.

shiny/ui/_chat.py
```diff
@@ -584,16 +584,20 @@ async def _handle_error():
...
-        await self._append_message(empty, chunk="start", stream_id=id)
-
+        is_start = True
         try:
             async for msg in message:
-                await self._append_message(msg, chunk=True, stream_id=id)
+                await self._append_message(
+                    msg, chunk="start" if is_start else True, stream_id=id
+                )
+                is_start = False
```
cpsievert, line 593: Turns out this change will change the order of messages in something like this:

```python
from shiny import reactive
from shiny.express import ui

ui.page_opts(title="Hello Chat")

chat = ui.Chat(id="chat")
chat.ui()

@reactive.effect
async def _():
    await chat.append_message_stream(("FIRST ", "FIRST ", "FIRST"))

@reactive.effect
async def _():
    await chat.append_message("SECOND SECOND SECOND")
```

This seems important enough that I'd like to try to keep synchronously starting the stream, if possible. Seems we could keep that and achieve a better/simpler loading message experience -- I have a look into it.

jcheng5: Yeah, sounds good.

jcheng5: I'm not super sympathetic to concurrent operations on the chat object like this, but I also don't love that (with my code) the front end doesn't know that message streaming is even happening until the first chunk is received.

wch:
Currently on this branch, when rendering a code block from user input, long lines get cut off to the left and there is no way to scroll. For example:

image


I don't know if this is the best solution, but I found that if I turn off `align-self: flex-end`, then the code displays more how I would expect:

image

cpsievert:
Thanks @wch, fixed by [af655fb](https://github.com/posit-dev/py-shiny/pull/1607/commits/af655fb483af61e42e071b2b0d83237d228a4d7c)

cpsievert:
Reviewed in person with Joe.

PR #1625, opened 2024-08-21
setup.cfg -> pyproject.toml

jcheng5:
This builds properly, but it doesn't quite work on local builds; requests for `shiny.js` result in 404. I think it may have to do with the `+` character that setuptools_scm puts into the version string, e.g. `1.0.1.dev21+g4d7a2e5e.d20240821`.


jcheng5:
@wch @schloerke can either of you think of any downstream problems from shiny having version numbers that look like `1.0.1.dev27`? In shinylive or something? Because this branch adds `setuptools_scm` and the way I've configured it, it will use version numbers like that for untagged builds.

schloerke:
Projects: 
* **py-shinywidgets**: ✅ ; Does not use the shiny version
* **py-shinyswatch**: ✅ ; Uses installed shiny package and only [uses bootstrap version](https://github.com/posit-dev/py-shinyswatch/blob/04820b92295fc54accb174bbff3556271eaa25e4/scripts/update_themes.py#L9).
* **shinylive assets**: ✅ ; Does not use shiny version
* **py-shinylive**: ✅ ; Currently [only uses assets version and it's own package version](https://github.com/posit-dev/py-shinylive/blob/6df837fe035969b84ab35ffa450961764b4a46e9/shinylive/_version/__init__.py#L1-L5)
* **shinylive quarto extension**: ✅ ; It only compares the shinylive assets version. 
	* There could be possible issues in the future if we add `.dev` versions to `py-shinylive`. But in [the current form](https://github.com/quarto-ext/shinylive/blob/7f0b6b76dd21d740facdb0260ba4fc92c07609dd/_extensions/quarto-ext/shinylive/shinylive.lua#L241) it should work as long as the major and minor versions are not confused with `.dev`.


I may have missed something and would like @wch to add his stamp of approval. But I have confidence so far!


--------------------------------


If we do need it on the packaging side, we can leverage `packaging.version import Version` for [version comparison](https://stackoverflow.com/a/28485211/591574).  

To remind myself... Unlike R where versions go from `1.1.0` to `1.1.0.9000`, python's `.dev` signals a dev version of the number that it's attached to. So after a `1.1.0` release, py-shiny should immediately update the version to `1.2.0.dev`. As `Version("1.2.0.dev0") < Version("1.2.0")` is `True`.

jcheng5:
TODO: Check if additional (unwanted) files are included in the sdist package bundle due to using setuptools_scm

pyproject.toml
```diff
@@ -0,0 +1,150 @@
...
+Repository = "https://github.com/posit-dev/py-shiny"
+"Bug Tracker" = "https://github.com/posit-dev/py-shiny/issues"
+
+[build-system]
+requires = ["setuptools>=60", "wheel", "setuptools_scm>=8.0"]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools]
+packages = {find = {include = ["shiny", "shiny.*"]}}
+package-data = {shiny = ["py.typed"]}
```
machow, line 128: I think setuptools scm might include package data by default (if it's committed in git), so there isn't a need for package-data and include-package-data, but I could be totally wrong

pyproject.toml
```diff
@@ -0,0 +1,150 @@
...
+    "griffe<1.0.0",
+]
+
+[project.urls]
+Homepage = "https://github.com/posit-dev/py-shiny"
+Documentation = "https://shiny.posit.co/py/"
+Repository = "https://github.com/posit-dev/py-shiny"
+"Bug Tracker" = "https://github.com/posit-dev/py-shiny/issues"
+
+[build-system]
```
machow, line 122: As a developer, it feels helpful to have build-system and tools.setuptools at the top, since it hints at what sections to expect in pyproject.toml / how things get made

machow: A helpful example might be ibis. They put their poetry config up top and build system at bottom. But because poetry is at the top you know what the build system will be set to.

pyproject.toml
jcheng5: I removed these directives and confirmed by inspection of the sdist that the same data is included as the pre-pyproject.toml version of Shiny. The only exception is that `/shiny.egg-info/not-zip-safe` is no longer included (has to do with setup.py vs. `build` module I think), I haven't yet investigated what that means.

pyproject.toml
jcheng5: Thanks!

pyproject.toml
jcheng5: According to https://setuptools.pypa.io/en/latest/deprecated/zip_safe.html the flag is obsolete.

PR #1641, opened 2024-08-26
Allow `@chat.transform_assistant_response` function to return `None`

wch:
This allows the `@chat.transform_assistant_response` function to return `None`, in which case nothing is sent to the client (and it will not update the output).

It also fixes a small issue where, when streaming, the last message would not be displayed. (This issue only manifests when the transform function returns `None`, because previously the last piece of text would get sent twice, once with `chunk_type=None` and once with `chunk_type="message_end"`.


js/chat/chat.ts
```diff
@@ -369,6 +372,7 @@ class ChatContainer extends LightElement {
 
     if (message.chunk_type === "message_end") {
       lastMessage.removeAttribute("is_streaming");
+      lastMessage.setAttribute("content", message.content);
```
cpsievert, line 375: Could you add context for this change? I'm fairly sure an [empty message gets send on `message_end`](https://github.com/posit-dev/py-shiny/blob/77a679/shiny/ui/_chat.py#L595), so I'm a bit surprised that this would change would do anything?

cpsievert, line 375: After chatting with @wch, we discovered/remembered this is needed for @wch's hand-rolled throttling implementation, but it's also useful if want to transform the content when the message is `done`, for example:

```python
from shiny.express import ui

chat = ui.Chat(id="chat")
chat.ui()


@chat.transform_assistant_response
def transform(content: str, chunk: str, done: bool) -> str | None:
    if done:
        return content + "...DONE!"
    else:
        return content


@chat.on_user_submit
async def _():
    await chat.append_message_stream(("Basic ", "response"))
```

cpsievert (approved):

PR #1648, opened 2024-08-27
Fix KeyError when serving static files

jcheng5:
Not sure what is triggering this but it's happening intermittently for both @cpsievert and myself.

Not sure why Content-Type would ever be missing, much less intermittently.

Edit: This happens when the server returns `304 Not Modified`... makes sense.


cpsievert (approved):

cpsievert:
Merging now since this is an important fix and makes the failures on main no worse than they are already

PR #1656, opened 2024-08-28
Add more information about tokenizers package

wch:
This PR adds more information about why the tokenizer package should be installed for the Chat component.

### DIFF
--- a/shiny/ui/_chat_tokenizer.py
+++ b/shiny/ui/_chat_tokenizer.py
@@ -54,7 +54,9 @@ def get_default_tokenizer() -> TokenizersTokenizer | None:
     except ImportError:
         warnings.warn(
             "`Chat` is unable obtain a default tokenizer without the `tokenizers` "
-            "package installed. Please `pip install tokenizers` or set "
+            "package installed. The tokenizer to use for calculating token counts, "
+            "which is required to impose `token_limits` in `.messages()`.\n"
+            "Please run `pip install tokenizers` or set "
             "`Chat(tokenizer=None)` to disable tokenization.",
             stacklevel=2,
         )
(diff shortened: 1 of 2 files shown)


shiny/ui/_chat_tokenizer.py
```diff
@@ -54,7 +54,9 @@ def get_default_tokenizer() -> TokenizersTokenizer | None:
     except ImportError:
         warnings.warn(
             "`Chat` is unable obtain a default tokenizer without the `tokenizers` "
-            "package installed. Please `pip install tokenizers` or set "
+            "package installed. The tokenizer to use for calculating token counts, "
+            "which is required to impose `token_limits` in `.messages()`.\n"
```
nealrichardson, line 58: Could the message hint at why I would want to impose token limits? Assume I don't know anything about LLMs or what makes them faster/better/cheaper.

shiny/ui/_chat_tokenizer.py
```diff
@@ -54,7 +54,9 @@ def get_default_tokenizer() -> TokenizersTokenizer | None:
...
+            "Please run `pip install tokenizers` or set "
             "`Chat(tokenizer=None)` to disable tokenization.",
```
nealrichardson, line 60: So when I see this message, my app is still working fine. So I'm guessing that it is falling back to None automatically? Would it be more accurate to say "set None to stop seeing this warning"?

wch:
Some notes from a discussion with @cpsievert:

- The reason for the default behavior of trimming messages is so that users don't suddenly run into errors when the conversation exceeds a particular length. The downside of that default is that we 
- I don't think that any other Python or JS LLM/chat libraries do trimming by default.
- I think it makes sense to set the default to `None`, and (A) make sure there the error message is clear when the user hits token limits, and (B) make it clear in the docs and examples that for any real chat app, they should install `tokenizers` and set `token_limits`.

cpsievert:
#1657 supersedes this

PR #1658, opened 2024-08-29
CI(deploy): Add more installation configs to surface failure cause

schloerke:
Configs:
* (New) Deploy to connect.posit.co; pypi shiny and rsconnect
	* Determine if deployment can happen
* (New) Deploy to dogfood; pypi shiny and rsconnect
	* Determine if deployment can happen to dogfood server
* (New) Deploy to dogfood; pypi shiny and github rsconnect
	* Determine if deployment can happen with latest rsconnect
* (New) Deploy to dogfood; github shiny@v1.0.0 and github rsconnect
	* Determine if deployment can happen with a github shiny package
* (Previously existed) Deploy to dogfood; github shiny and rsconnect
	* Determine if deployment can happen with the latest github version of shiny package

When errors occur on deployment, we will now be able to quickly determine if the dogfood server, rsconnect package, or shiny is the reason.

PR #1660, opened 2024-08-30
CI(debug): Print full stdout/stderr when command fails

schloerke:
* Better stderr/stdout output for #1658 
* Fix bug where `make playwright-deploys` was testing all playwright tests, not just `./playwright/deploys` tests


schloerke:
Goals of PR (in PR description) achieved. Merging as CI failures are expected (deploys) or unrelated (run tests)

PR #1678, opened 2024-09-11
test: Add skip for airmass example on python 3.9

schloerke:
Error when trying to run `./examples/airmass/app.py`:

```
    import astropy.units as u
  File "/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/astropy/units/__init__.py", line 23, in <module>
    from .quantity import *
  File "/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/astropy/units/quantity.py", line 36, in <module>
    from .quantity_helper import can_have_arbitrary_unit, check_output, converters_and_unit
  File "/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/astropy/units/quantity_helper/__init__.py", line 16, in <module>
    from . import erfa, function_helpers, helpers, scipy_special
  File "/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/astropy/units/quantity_helper/function_helpers.py", line 79, in <module>
    np.prod, np.product, np.cumprod, np.cumproduct,
  File "/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/numpy/__init__.py", line 410, in __getattr__
    raise AttributeError("module {!r} has no attribute "
AttributeError: module 'numpy' has no attribute 'product'
```

So, disabling test on python 3.9 as it is a smoke test that is covered by other python versions.

astropy is very aggressive on their min python version. Therefore, it is possible to get a not-optimal combination of packages being installed:
* `numpy` == 2.0.2
* `astropy` == 5.3.2

PR #1708, opened 2024-09-27
chore: revert check where airmass skips on python 3.9

schloerke:
Related #1678


schloerke:
Failure unrelated to PR goal. Merging

PR #1717, opened 2024-10-04
build: set Narwhals 1.9.1 as minimum, remove upper bound, use narwhals.stable.v1 for dtypes and typing

MarcoGorelli:
Hey - as discussed in https://github.com/posit-dev/py-shiny/pull/1570, a blocker to not upper-bounding Narwhals was that `dtypes` and `typing`  weren't available in `narwhals.stable.v1`

We have now:
- made a release which includes them
- added a downstream test in the Narwhals CI which checks that the `py-shiny` test suite + type checking keeps working on every commit

so I'd like to suggest removing the upper-bound

Thanks again for the collaboration here, it's been very rewarding to see this happen! 🤗

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ dependencies = [
     "prompt-toolkit;platform_system!='Emscripten'",
     "python-multipart>=0.0.7;platform_system!='Emscripten'",
     "setuptools;python_version>='3.12'",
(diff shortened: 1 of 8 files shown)


schloerke:
Very exciting that the dtypes made it into stable 🥳 

Thank you!

PR #1725, opened 2024-10-09
docs: move description list render logic into quartodoc

machow:
Recently, quartodoc cribbed the structure shiny used to render parameters in its documentation 😁. This PR attempts to take swap out the existing logic in the shiny docs renderer with quartodoc.

Note that I've set the CI to install quartodoc from main. Once everything looks good here and in [this Great Tables PR](https://github.com/posit-dev/great-tables/pull/427), I can do a quartodoc release!

Here are some critical changes:

* Introduced the ParamRow dataclass, to represent a row in a parameter table (prev just used a tuple).
* Moved sanitization into ParamRow, rather than performing in each render method
* **New!**: Added html classes to headers. E.g. "## Examples {.doc-section, .doc-section-examples}

### DIFF
--- a/docs/_renderer.py
+++ b/docs/_renderer.py
@@ -24,6 +24,7 @@ from plum import dispatch
 from quartodoc import MdRenderer
 from quartodoc.pandoc.blocks import DefinitionList
 from quartodoc.renderers.base import convert_rst_link_to_md, sanitize
+from quartodoc.renderers.md_renderer import ParamRow
 
 # from quartodoc.ast import preview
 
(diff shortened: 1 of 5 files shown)


shiny/input_handler.py
```diff
@@ -52,18 +52,6 @@ def _process_value(
 for more information on how to create custom input bindings. (The article is about
 Shiny for R, but the JavaScript and general principles are the same.)
 
-Methods
```
machow, line 55: I think I deleted this while pairing w/ Barret, but can put back in. I think it caused griffe to error when parsing the docstring (but could be wrong).

docs/_renderer.py
```diff
@@ -279,7 +258,7 @@ def read_file(file: str | Path, root_dir: str | Path | None = None) -> FileConte
 
 
 def check_if_missing_expected_example(el, converted):
-    if re.search(r"(^|\n)#{2,6} Examples\n", converted):
+    if re.search(r"(^|\n)#{2,6} Examples", converted):
```
machow, line 261: quartodoc adds class names to the sections now, so we had to remove the newline from the regex.

schloerke:
Closing in favor of #1727

PR #1754, opened 2024-11-02
Change chat custom normalizer error message to point to correct function

wch:
Previously, the error message said this:

```
Consider registering a custom normalizer via shiny.ui._chat_types.registry.register()
```

But the function `shiny.ui._chat_types.registry.register()` doesn't exist. I changed it to what I think is the correct function, `shiny.ui._chat_normalize.message_normalizer_registry.register()`, but please confirm that it's right.

### DIFF
--- a/shiny/ui/_chat_normalize.py
+++ b/shiny/ui/_chat_normalize.py
@@ -277,7 +277,7 @@ def normalize_message(message: Any) -> ChatMessage:
             return strategy.normalize(message)
     raise ValueError(
         f"Could not find a normalizer for message of type {type(message)}: {message}. "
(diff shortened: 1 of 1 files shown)


cpsievert (approved):

cpsievert:
Thanks, just curious, did you have to implement a new strategy, or is #1755 enough to accomplish what you need?

cpsievert:
Good to know. As a heads up, we're soon going to promote [chatlas](https://github.com/cpsievert/chatlas) as the way to manage the conversation state (and generate responses) instead of `chat.message(format="...")`. That said, we don't currently have a good way to take advantage of Anthropic's approach to prompt caching, so probably worth sticking `chat.message(format="anthropic")` in shiny assistant, at least for now.

PR #1755, opened 2024-11-02
Add support for Anthropic prompt caching

wch:
This change adds support for Anthopic's beta prompt caching feature.

### DIFF
--- a/shiny/ui/_chat_normalize.py
+++ b/shiny/ui/_chat_normalize.py
@@ -185,14 +185,28 @@ class AnthropicNormalizer(BaseMessageNormalizer):
             # The actual MessageStreamEvent is a generic, so isinstance() can't
             # be used to check the type. Instead, we manually construct the relevant
             # union of relevant classes...
-            return (
+            if (
                 isinstance(chunk, RawContentBlockDeltaEvent)
                 or isinstance(chunk, RawContentBlockStartEvent)
                 or isinstance(chunk, RawContentBlockStopEvent)
                 or isinstance(chunk, RawMessageDeltaEvent)
                 or isinstance(chunk, RawMessageStartEvent)
                 or isinstance(chunk, RawMessageStopEvent)
+            ):
+                return True
+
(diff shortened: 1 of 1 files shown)


cpsievert:
Here's a minimal example derived https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching

```python
from anthropic import AsyncAnthropic
from app_utils import load_dotenv

from shiny.express import ui

load_dotenv()
llm = AsyncAnthropic()

chat = ui.Chat(id="chat")
chat.ui()
chat.update_user_input(value="Analyze the major themes in 'Pride and Prejudice'")


@chat.on_user_submit
async def _():
    response = await llm.beta.prompt_caching.messages.create(
        model="claude-3-5-sonnet-20241022",
        max_tokens=1024,
        stream=True,
        system=[
            {
                "type": "text",
                "text": "You are an AI assistant tasked with analyzing literary works. Your goal is to provide insightful commentary on themes, characters, and writing style.\n",
            },
            {
                "type": "text",
                "text": "<the entire contents of 'Pride and Prejudice'>",
                "cache_control": {"type": "ephemeral"},
            },
        ],
        messages=[{"role": "user", "content": chat.user_input()}],
    )
```

shiny/ui/_chat_normalize.py
```diff
@@ -185,14 +185,28 @@ def can_normalize_chunk(self, chunk: Any) -> bool:
...
+            # types, so we need to check for them separately. If this import fails or
+            # errors, then we'll end up in the Exception handler and return False.
+            from anthropic.types.beta.prompt_caching import (
+                RawPromptCachingBetaMessageStartEvent,
             )
+
+            if isinstance(chunk, RawPromptCachingBetaMessageStartEvent):
+                return True
+
+            return False
```
cpsievert, line 208: This change to `can_normalize_chunk()` looks good.

If we want to also support this feature for the non-streaming case, we'll want to also update `can_normalize()` to know about `PromptCachingBetaMessage`

PR #1791, opened 2024-12-04
Ignore type error in test introduced by ollama 0.4.2

cpsievert:
Gets `pyright` passing on CI.

ollama must not use `pyright` for type checking because `ollama.chat` doesn't have valid `pyright` type hints

https://github.com/ollama/ollama-python/commit/b50a65b27d6c4e5bac42fcf763234e6a4be128b6


cpsievert:
Plyawright failures are due to different problems with airmass example. Merging

PR #1821, opened 2025-01-15
README: fix broken installation link

WillGiang:
https://shiny.posit.co/py/docs/install.html - > https://shiny.posit.co/py/docs/install-create-run.html


gadenbuie (approved):
Thanks!

PR #1831, opened 2025-01-29
Run black (fix CI)

cpsievert:
Black _just_ did a release, this simply runs `black .` on the codebase


cpsievert:
Other failures are due to `airmass` example tests 😞

PR #1842, opened 2025-02-05
Add support for async server function

mconflitti-pbc:
Coming across a use case using chatlas where I need to be able to await an async function in my server function but shiny does not currently support this.

Simple addition to `_session.py` to handle both sync and async.

### DIFF
--- a/shiny/_app.py
+++ b/shiny/_app.py
@@ -6,7 +6,7 @@ import secrets
 from contextlib import AsyncExitStack, asynccontextmanager
 from inspect import signature
 from pathlib import Path
-from typing import Any, Callable, Mapping, Optional, TypeVar, cast
+from typing import Any, Awaitable, Callable, Mapping, Optional, TypeVar, cast
 
 import starlette.applications
 import starlette.exceptions
@@ -57,8 +57,8 @@ class App:
         returns a UI definition, if you need the UI definition to be created dynamically
         for each pageview.
     server
-        A function which is called once for each session, ensuring that each session is
-        independent.
+        A sync or async function which is called once for each session, ensuring that
+        each session is independent.
     static_assets
         Static files to be served by the app. If this is a string or Path object, it
         must be a directory, and it will be mounted at `/`. If this is a dictionary,
@@ -104,13 +104,13 @@ class App:
     """
 
     ui: RenderedHTML | Callable[[Request], Tag | TagList]
-    server: Callable[[Inputs, Outputs, Session], None]
+    server: Callable[[Inputs, Outputs, Session], Awaitable[None] | None]
 
     def __init__(
         self,
         ui: Tag | TagList | Callable[[Request], Tag | TagList] | Path,
         server: (
-            Callable[[Inputs], None] | Callable[[Inputs, Outputs, Session], None] | None
+            Callable[[Inputs], Awaitable[None] | None] | Callable[[Inputs, Outputs, Session], Awaitable[None] | None] | None
         ),
         *,
         static_assets: Optional[str | Path | Mapping[str, str | Path]] = None,
--- a/shiny/session/_session.py
+++ b/shiny/session/_session.py
@@ -633,7 +633,9 @@ class AppSession(Session):
                             self._manage_inputs(message_obj["data"])
 
                             with session_context(self):
-                                self.app.server(self.input, self.output, self)
+                                result = self.app.server(self.input, self.output, self)
+                                if isinstance(result, Awaitable):
+                                    await result
 
                         elif message_obj["method"] == "update":
                             verify_state(ConnectionState.Running)


schloerke:
Is your use case for setting up data?  When are you needing the result of the information? (I guess, can you give a tiny motivation example that chatlas would use?)

Why not put it behind an async reactive calculation? Then the value can be used accordingly.

Ex:

From 
```python
async def server(input):
    data = await get_data()
```

to

```python
def server(input):

    @reactive.calc
    async def data():
        return await get_data()
```

mconflitti-pbc:
> Is your use case for setting up data? When are you needing the result of the information? (I guess, can you give a tiny motivation example that chatlas would use?)
> 
> Why not put it behind an async reactive calculation? Then the value can be used accordingly.
> 
> Ex:
> 
> From
> 
> ```python
> async def server(input):
>     data = await get_data()
> ```
> 
> to
> 
> ```python
> def server(input):
> 
>     @reactive.calc
>     async def data():
>         return await get_data()
> ```

Inside of chatlas, the server function is defined for you and I want to be able to access the session headers to get the user session token when using chat.app()

in chat.app()
```python
  async def server(input, output, session):  # noqa: A002
      if session_callback:
          await session_callback(session)
```

May be too niche of a problem, but was surprised to see shiny didnt support both. And afaict, no reason that it can't.

shiny/_app.py
```diff
@@ -104,13 +104,15 @@ def server(input: Inputs, output: Outputs, session: Session):
...
+    server: Callable[[Inputs, Outputs, Session], Awaitable[None] | None]
 
     def __init__(
         self,
         ui: Tag | TagList | Callable[[Request], Tag | TagList] | Path,
         server: (
-            Callable[[Inputs], None] | Callable[[Inputs, Outputs, Session], None] | None
+            Callable[[Inputs], Awaitable[None] | None]
+            | Callable[[Inputs, Outputs, Session], Awaitable[None] | None]
+            | None
```
schloerke, line 115: Because this can be defined with an async function, can we use [`wrap_async`](https://github.com/posit-dev/py-shiny/blob/3d1108c576ce426eec9efc019b3e7e7757f207c8/shiny/_utils.py#L264) when storing the fn value below? 

This will always upgrade the function to async. Then in `_session.py`, we will always `await` the server execution as the run method is already an async method.

schloerke (changes requested):

schloerke:
> was surprised to see shiny didnt support both. And afaict, no reason that it can't.

Fair!

shiny/_app.py
schloerke: The current `cast` calls are lying as to what values are already there. Please update them accordingly.

By wrapping all server functions to be async, we won't need to check if we need to await or not (as we'll _always_ need to await).

mconflitti-pbc: Sweet. Makes sense to me.

mconflitti-pbc: Though, now looking into support server function to be async, would need to go down the rabbit hole of ensure that we tackle other side effects like the module.server() decorator which is not async. Could leave it only supporting sync i suppose.. curious on thoughts from @wch.

schloerke (approved):

shiny/_utils.py
```diff
@@ -362,10 +362,10 @@ def is_async_callable(
...
 
 
-# def not_is_async_callable(
-#     obj: Callable[P, T] | Callable[P, Awaitable[T]]
-# ) -> TypeGuard[Callable[P, T]]:
-#     return not is_async_callable(obj)
+def not_is_async_callable(
+    obj: Callable[P, T] | Callable[P, Awaitable[T]],
+) -> TypeGuard[Callable[P, T]]:
+    return not is_async_callable(obj)
```
wch, line 368: Instead of adding this function, can we change `is_async_callable()` to a `TypeIs`?

shiny/_app.py
```diff
@@ -123,7 +124,9 @@ def __init__(
         self,
         ui: Tag | TagList | Callable[[Request], Tag | TagList] | Path,
         server: (
-            Callable[[Inputs], None] | Callable[[Inputs, Outputs, Session], None] | None
+            Callable[[Inputs], Awaitable[None] | None]
+            | Callable[[Inputs, Outputs, Session], Awaitable[None] | None]
```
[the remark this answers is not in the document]
wch: Agreed, they are different types and need separate entries for correctness.

shiny/_utils.py
```diff
@@ -262,15 +262,15 @@ def private_seed() -> Generator[None, None, None]:
 
 
 def wrap_async(
-    fn: Callable[P, R] | Callable[P, Awaitable[R]],
+    fn: Callable[P, R] | Callable[P, Awaitable[R]] | Callable[P, Awaitable[R] | R],
```
wch, line 265: I think the change on this line shouldn't be necessary, after making the changes above in the signature of `__init__`.

shiny/_utils.py
```diff
@@ -262,15 +262,15 @@ def private_seed() -> Generator[None, None, None]:
...
 ) -> Callable[P, Awaitable[R]]:
     """
     Given a synchronous function that returns R, return an async function that wraps the
     original function. If the input function is already async, then return it unchanged.
     """
 
     if is_async_callable(fn):
-        return fn
+        return cast(Callable[P, Awaitable[R]], fn)
```
wch, line 273: Is this change necessary?

shiny/_utils.py
wch: I just saw the comment below about how `TypeIs` didn't work. Can you add a comment explaining the problem? Otherwise we'll keep coming back to this and asking `TypeIs` in the future.

wch:
I wonder if there's a way to define a generic that handles the sync/async function cases. I asked Claude and it gave me something, but it will require a closer look to make sure it's right, and perhaps it should go in a separate PR that puts it in types.py.

```py
from typing import TypeVar, Awaitable, Protocol, Callable, Union, ParamSpec, cast, overload

# Define a covariant TypeVar for the return type
R_co = TypeVar('R_co', covariant=True)

# Define a parameter specification for flexible argument handling
P = ParamSpec('P')

# More descriptive type alias
MaybeAwaitableCallable = Union[
    Callable[P, R_co],
    Callable[P, Awaitable[R_co]]
]

async def wrap(fn: MaybeAwaitableCallable[P, None], *args: P.args, **kwargs: P.kwargs) -> int:
    # Implementation remains the same
    result = fn(*args, **kwargs)
    # Note: We would use is_async_callable() instead of isinstance()
    if isinstance(result, Awaitable):
        await result
    return 42


## Example usage

def sync_function(a: int, b: str) -> None:
    print(f"Sync: {a}, {b}")

async def async_function(a: int, b: str) -> None:
    print(f"Async: {a}, {b}")


async def test() -> None:
    # Both of these are valid
    await wrap(sync_function, 1, "test")
    await wrap(async_function, 2, "async")
```

Issue #740, opened 2023-09-26
Maintain calculation results in session state


jcheng5:
@vnijs had a request to have the results of a long-running reactive calculation "persist" even through navigating away and back. In talking this through in person, maybe the equivalent of Shiny for R's `bindCache` (combined with bookmarkable state) would do it.

vnijs:
@GShotwell mentioned the idea of saving models, which I really liked. Being able to load them back in + show output and create additional plots, diagnostics, etc. would also be really interesting. You could do that with a separate app designed just to evaluate a particular type of saved models. However, to be able to do it in the same estimation app would be really nice (i.e., restore input values, reactive.Calc state, and output).

liuzj039:
I encountered the same issue. I would like to ask if there have been any recent developments about the `bindCache` function. Thank you very much.

sambrilleman:
Just curious if there are any updates on this, or people have any workarounds. Is there a Python equivalent of R's `bindCache`?

My use case is that I have to hit an external web API, and that API request will be the same for each user of the app. So I would prefer to cache it, and include a time component in the key to invalidate the cached result after X minutes. I've read the docs but based on what I can tell there isn't currently this type of functionality supported in Shiny for Python?

gshotwell:
@sambrilleman I bet you could get this behaviour by using a reactive outside of the server function. This will let all of the users of the app share the same API call, which you could invalidate on a schedule with `invalidate_later`. Alternatively you could use `reactive.poll` to call a cheap API call regularly and only do the full API refresh when something changed. Here's an example I think 

https://github.com/posit-dev/py-shiny-templates/blob/main/monitor-database/shared.py#L26
https://github.com/posit-dev/py-shiny-templates/blob/main/monitor-database/app-core.py#L4

sambrilleman:
Legend! Thanks @gshotwell , we'll give that approach a go 😀 I guess I didn't realise we can use a reactive outside of the context of a server. I guess if it's out there on its own, not part of the server, it just forms its own reactive dependency graph or something like that 🤔 (i.e. only linked to the invalidation/polling like you mention). Interesting that it still fires on start up as part of the App though. I guess because its still called from the server (even if not declared there) so must be called/loaded by the app atleast once. Anyhow ignore me if I'm talking some gibberish, pretty new to Shiny. I'll take a closer look at your example, and we can just copy that behaviour, thanks!

gshotwell:
[...]

PR #1870, opened 2025-02-25
feat: Add support for bookmarking via `shiny.bookmark`

### LINKED ISSUE #1225: How to bookmark page state?
That is a thing in the R version, is this feature hidden? Or not planned for development?

schloerke:
Fixes #1225
Related #740
Related #1905 

Working demo app:
```python
from starlette.requests import Request

from shiny.bookmark import BookmarkState


# App UI **must** be a function to ensure that each user restores their own UI values.
def app_ui(request: Request):
    return ui.page_fluid(
        ui.markdown(
            "Directions: "
            "\n1. Change the radio buttons below"
            "\n2. Refresh your browser."
            "\n3. The radio buttons should be restored to their previous state."
            "\n4. Check the console messages for bookmarking events."
        ),
        ui.hr(),
        ui.input_radio_buttons(
            "letter",
            "Choose a letter (Store in Bookmark 'input')",
            choices=["A", "B", "C"],
        ),
        ui.input_radio_buttons(
            "letter_values",
            "Choose a letter (Stored in Bookmark 'values' as lowercase)",
            choices=["A", "B", "C"],
        ),
        "Selection:",
        ui.output_code("letters"),
    )


    # Exclude `"letter_values"` from being saved in the bookmark as we'll store it manually for example's sake
    # Append or adjust this list as needed.
    session.bookmark.exclude.append("letter_values")

    lowercase_letter = reactive.value()

    @reactive.effect
    @reactive.event(input.letter_values)
    async def _():
        lowercase_letter.set(input.letter_values().lower())

    @render.code
    def letters():
        return str([input.letter(), lowercase_letter()])

    # When the user interacts with the input, we will bookmark the state.
    @reactive.effect
    @reactive.event(input.letter, lowercase_letter, ignore_init=True)
    async def _():
        await session.bookmark()

    # Before saving state, we can adjust the bookmark state values object
    @session.bookmark.on_bookmark
    async def _(state: BookmarkState):
        print("Bookmark state:", state.input, state.values, state.dir)
        with reactive.isolate():
            state.values["lowercase"] = lowercase_letter()

    # After saving state, we will update the query string with the bookmark URL.
    @session.bookmark.on_bookmarked
    async def _(url: str):
        print("Bookmarked url:", url)
        await session.bookmark.update_query_string(url)

    @session.bookmark.on_restore
    def _(state: BookmarkState):
        print("Restore state:", state.input, state.values, state.dir)

        # Update the radio button selection based on the restored state.
        if "lowercase" in state.values:
            uppercase = state.values["lowercase"].upper()
            # This may produce a small blip in the UI as the original value was restored on the client's HTML request, _then_ a message is received by the client to update the value.
            ui.update_radio_buttons("letter_values", selected=uppercase)

    @session.bookmark.on_restored
    def _(state: BookmarkState):
        # For rare cases, you can update the UI after the session has been fully restored.
        print("Restored state:", state.input, state.values, state.dir)


# Make sure to set the bookmark_store to `"url"` (or `"server"`)
# to store the bookmark information/key in the URL query string.
app = App(app_ui, server, bookmark_store="url")
```


docs/_quartodoc-core.yml
```diff
@@ -98,6 +98,23 @@ quartodoc:
...
+        - bookmark.RestoreState
+        - kind: page
+          path: bookmark_integration
+          summary:
+            name: "Integration"
+            desc: "Decorators to set save and restore directories."
+          flatten: true
+          contents:
+            - bookmark.set_save_dir
+            - bookmark.set_restore_dir
```
wch, line 117: Have the ability to set these at the `App` level.

shiny/session/_session.py
```diff
@@ -1357,6 +1399,83 @@ def __contains__(self, key: str) -> bool:
     def __dir__(self):
         return list(self._map.keys())
 
+    _serializers: dict[
+        str,
+        Callable[
+            [Any, Path | None],
+            Awaitable[Any | Unserializable],
+        ],
+    ]
```
wch, line 1408: Move this to top of class defintion.

shiny/session/_session.py
```diff
@@ -1173,14 +1211,17 @@ def __init__(self, parent: Session, ns: ResolvedId) -> None:
...
+
     def _is_hidden(self, name: str) -> bool:
         return self._parent._is_hidden(name)
 
     def on_ended(
         self,
         fn: Callable[[], None] | Callable[[], Awaitable[None]],
     ) -> Callable[[], None]:
-        return self._parent.on_ended(fn)
+        return self._root.on_ended(fn)
```
wch, line 1224: Try to do this with `_parent` instead of `_root`.

shiny/bookmark/_utils.py
```diff
@@ -0,0 +1,27 @@
...
+import os
+from typing import Any
+
+import orjson
+
+
+def is_hosted() -> bool:
+    # Can't look at SHINY_PORT, as we already set it in shiny/_main.py's `run_app()`
+
+    # TODO: Barret: Q: How to support shinyapps.io? Or use `SHINY_PORT` how R-shiny did
```
wch, line 12: Make a PR to change the name `SHINY_PORT` to something else, to avoid this naming conflict.

docs/_quartodoc-core.yml
schloerke: [`a50860e` (#1870)](https://github.com/posit-dev/py-shiny/pull/1870/commits/a50860e21c5a86af723e757f04829fa4457e2549)

shiny/session/_session.py
schloerke: Root has been removed. Everything is now through the parent

PR #1894, opened 2025-03-08
JS build script updates

wch:
This updates and streamlines the build.ts script. Note that this PR builds on #1893, so until that PR is merged, the diff here will look larger than the real changes compared to #1893. (The only relevant changes here are to build.ts.)

Some notable things:

- Metafiles are generated by default now.
- The metafile now contains information for all of the built assets, not just one.
- Previously, many of the assets had their own settings for `sourcemap`, `minify`, and `metafile`. I don't know why that would be desirable, so I removed that.

The merged metafile, when viewed at https://esbuild.github.io/analyze/, looks like this:

image


js/build.ts
```diff
@@ -21,7 +63,7 @@ async function bundle_helper(
       bundle: true,
       minify: minify,
       sourcemap: true,
```
wch, line 65:
```suggestion
      sourcemap: minify,
```

schloerke (approved):

PR #1924, opened 2025-03-17
fix: Revert - fix(session): Always have calling session be the parent session when making scoped session

schloerke:
This reverts commit 14b92d4e8f273560ba63a9c4e1206a7f4b0aa9e1 via #1923

From conversations with @jcheng5 , the proxy sessions should have a single "parent" object which is "Root" session, they should not recurse up the session family tree during calls. 

In a followup PR, I'll rename `._parent` to `._root_session` to avoid confusion.

From chat w/ @jcheng5 
> If there are reasons we want nested session scopes to actually be related (and there are reasons we might--like we could add `.destroy()` methods that recursively kill their child scopes), we can do that, but we’d have to make further changes I think. 
> 
> Off the top of my head, we’d have to make sure that make_scope:
> 1. Always returns the same object when `make_scope` is passed with the same `id`, much like Python does with loaded modules. We’d need to do this with weakrefs.
> 2. If `make_scope` is passed with an id more than one level deep, we’d need to create/fetch a separate session proxy object for each level.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Fixed an issue with the `.click()` method on InputActionButton controllers in `shiny.playwright.controllers` where the method would not work as expected. (#1886)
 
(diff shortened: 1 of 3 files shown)


jcheng5:
LGTM if @wch approves

schloerke:
Approval received via chat: 

> > In a followup PR, I'll rename ._parent to ._root_session to avoid confusion.
>
> @wch: That is a better name :+1:

Merging

PR #1928, opened 2025-03-19
chore: Add support for Narwhals `Binary` dtype in `serialize_dtype`

FBruzzesi:
# Description

[Narwhals#2243](https://github.com/narwhals-dev/narwhals/pull/2243) is going to introduce `Binary` dtype, this PR prepares for that by serializing `Binary` to `"binary"`


schloerke:
Thank you!

PR #1930, opened 2025-03-19
feat(session): Add `.on_end(fn)` support

### LINKED ISSUE #1929: [Feature]: On session disconnect, bookmark the state
### Category

Other

### Scope

Quality of Life

### Problem

When the Shiny session disconnects due to timeout, all state is lost on the server. When the user returns, the session state is reset.

I'd like for the state to persist between sessions	

### Solution

We should behave like Shiny Assistant and try to save the state as a last ditch effort before the session goes away.

This would be an alternative to updating the query string rapidly.

### Alternatives (Optional)

_No response_

### Example (Optional)

```python

```

### Impact (Optional)

_No response_

### Contribution? (Optional)

None

schloerke:
Fixes #1929 

These callbacks will execute just before the websocket is closed. Doesn't mean it's guaranteed to work, but it's better than `session.on_ended` (which is definitely after the websocket is closed).

https://github.com/user-attachments/assets/b9d8c0a7-e381-4bca-99bc-6fd2f1c45245

```python
from starlette.requests import Request

from shiny.bookmark import BookmarkState


# App UI **must** be a function to ensure that each user restores their own UI values.
def app_ui(request: Request):
    return ui.page_fluid(
        ui.markdown(
            "Directions: "
            "\n1. Change the radio buttons below"
            "\n2. Changing the radio button will cause a server error."
            "\n3. The app will save the state to the URL before disconnecting."
            "\n4. Refresh, your browser. The radio button selection should persist."
        ),
        ui.hr(),
        ui.input_radio_buttons(
            "letter",
            "Pick a letter",
            choices=["A", "B", "C"],
        ),
        "Selection:",
        ui.output_code("letter_out"),
    )


    @render.code
    def letter_out():
        return input.letter()

    @reactive.effect
    @reactive.event(input.letter, ignore_init=True)
    async def _():
        # Simulate a server error when the radio button is changed
        raise RuntimeError("Simulated server error")

    # When the session ends, update the query string with the latest bookmark information
    @session.on_end
    async def _():
        await session.bookmark.update_query_string()

    # Reset query string when restoring state
    @session.bookmark.on_restored
    async def _(state: BookmarkState):
        await session.bookmark.update_query_string(session.clientdata.url_pathname())


# Make sure to set the bookmark_store to `"url"` (or `"server"`)
# to store the bookmark information/key in the URL query string.
app = App(app_ui, server, bookmark_store="url")

```


schloerke:
While this PR addresses a serverside-error, it does not fix being able to use a WS before a disconnect occurs.

Given the marginal gain, I'm closing the PR.

PR #1940, opened 2025-03-24
fix(Chat): fix statically rendered custom icons

cpsievert:
Follow up to #1853

Apparently statically rendered custom icons were never working as intended. Here's an example:

```python
import faicons

from shiny.express import ui

welcome = """
**Hello!** How can I help you today?

Here are a couple suggestions:

* <span class="suggestion">Tell me a joke</span>
* <span class="suggestion submit">Tell me a story</span>
"""

ui.br()

chat = ui.Chat(id="chat")
chat.ui(
    messages=[welcome],
    icon_assistant=faicons.icon_svg("otter"),
)
```

Currently, you get the default robot icon, when you should get an otter in the startup message.

This got overlooked because `ui.Chat()`'s `messages` are dynamically rendered (and thus, are included in the server-side state), but `.ui(messages)` is static (and thus is not)

### DIFF
--- a/shiny/ui/_chat.py
+++ b/shiny/ui/_chat.py
@@ -1434,6 +1434,14 @@ def chat_ui(
 
     id = resolve_id(id)
 
+    icon_attr = None
+    if icon_assistant is not None:
+        icon_attr = str(icon_assistant)
+
+    icon_deps = None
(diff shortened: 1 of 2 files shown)


gadenbuie (approved):
Thanks!

PR #1947, opened 2025-03-29
fix(input_select): Decouple `input_select()` and `input_selectize()`

### LINKED ISSUE #1687: [bug]: `ui.input_select()` and `ui.input_selectize()`'s interfaces should be de-coupled
Currently, `ui.input_select()` is a base function for `ui.input_selectize()`. 

This requires that `ui.input_select()` have extra arguments:
* `selectize:bool = False`; Enables `selectize.js`
* `remove_button: Optional[bool] = None`; Adds remove button plugin for `selectize.js`
* `options: Optional[dict[str, Jsonifiable | JSEval]] = None`; Options supplied to `selectize.js`

In addition, `ui.input_select(choices=)` has type support for HTML tags. While it currently doesn't throw an error, I believe a type error should be displayed.

Ex:     
```python
ui.input_select( 
	"basic_select2",
	"Default select",
	{
		"a": "Choice A",
		"b": ui.tags.div(ui.tags.strong("Choice B1"), ui.tags.strong("Choice B2")),
	},
)
```

This is displays as if the second choice has a value of `" Choice B1AChoice B2 "`, not the HTML tags that is represents.

-------------------------

Proposal:

* Move the current definition of `ui.input_select()` to `input_select_impl()`. Have both `ui.input_select()` and `ui.input_selectize()` call `input_select_impl()` directly.
* Deprecate and eventually remove `selectize=`, `remove_button=`, and `options=` from `input_select()`. 
* Create separate typing for `input_select(choices=)` where the eventual [type definition of `_Choices = Mapping[str, TagChild]
`](https://github.com/posit-dev/py-shiny/blob/14d820c52f3f0da9d1b30a19bdafaa1eaf4a3a9f/shiny/ui/_input_select.py#L23) becomes `Mapping[str, str]`

ErdaradunGaztea:
Fixes #1687

This PR follows the suggestions made in the issue and repurposes current `input_select()` implementation as `_input_select_impl()`. The new `input_select()` keeps the same signature but raises deprecation warnings upon using any of the three Selectize-related parameters.

Additionally, passing anything that's not a string (i.e. a HTML tag or `None`) as an option label raises another deprecation warning. The final goal would be to replace `_Choices = Mapping[str, TagChild]` with `_Choices = Mapping[str, str]`, but I guess that comes after a period of deprecation.

---

It's my first PR to this project, so if I missed some important part of the PR process, please let me know.


shiny/ui/_input_select.py
```diff
@@ -282,6 +340,17 @@ def _normalize_choices(x: SelectChoicesArg) -> _SelectChoices:
         return x
 
 
+def _contains_html(x: _SelectChoices) -> bool:
+    for v in x.values():
+        if isinstance(v, Mapping):
+            if _contains_html(cast(_OptGrpChoices, v)):
```
schloerke, line 346: At this point, `v` is the `_Choices` values of `_OptGrpChoices` Mapping.

```suggestion
            if _contains_html(cast(_Choices), v)):
```

shiny/ui/_input_select.py
```diff
@@ -282,6 +340,17 @@ def _normalize_choices(x: SelectChoicesArg) -> _SelectChoices:
...
+                return True
```
schloerke, line 347: While the typings wouldn't recurse forever, the recursive nature of the function gives me a little pause.

Could we instead perform it directly as we know the recursive depth should only ever be 1 and we'd only check for "not strings"?

```suggestion
            # Check the `_Choices` values of `_OptGrpChoices`
            for vv in v.values():
                if not is instance(vv, str):
                    return True
```

shiny/ui/_input_select.py
```diff
@@ -192,6 +195,55 @@ def input_select(
...
+        warn_deprecated(
+            "`remove_button` parameter of `input_select()` is deprecated. "
+            "Use `input_selectize()` instead."
+        )
+
+    if options is not None:
+        warn_deprecated(
+            "`options` parameter of `input_select()` is deprecated. "
+            "Use `input_selectize()` instead."
+        )
```
schloerke, line 214: For all three parameters, please change the parameter type to `MISSING_TYPE` and default the value to `DEPRECATED` ([link](https://github.com/posit-dev/py-shiny/blob/785197861d209e04a2498972143660aaa10f6a96/shiny/types.py#L48))

This will clearly signal that it should not be used. (Similar to how `lifecycle::deprecated()` is used in R packages)

To check if the value is present, use code similar to `if isinstance(selectize, MISSING_TYPE):` for the three cases.

Thank you

schloerke (changes requested):
> It's my first PR to this project

Thank you for your PRs!

Please be sure to add an entry in the CHANGELOG.md file linking to your PR. ([Related example](https://github.com/posit-dev/py-shiny/blob/785197861d209e04a2498972143660aaa10f6a96/CHANGELOG.md?plain=1#L841))

schloerke:
The app test failing in https://github.com/posit-dev/py-shiny/actions/runs/14320952734/job/40142214665?pr=1947#step:5:1519 is due to a change in this PR. 


```bash
pytest ./tests/playwright/ai_generated_apps/bookmark/input_select
```

Please update the test app or fix the helper method. Thank you!

shiny/ui/_input_select.py
ErdaradunGaztea: With your other suggestion, this one is not relevant anymore, but you are right, that should have been `_Choices`!

ErdaradunGaztea:
> The app test failing in https://github.com/posit-dev/py-shiny/actions/runs/14320952734/job/40142214665?pr=1947#step:5:1519 is due to a change in this PR.
> 
> ```shell
> pytest ./tests/playwright/ai_generated_apps/bookmark/input_select
> ```
> 
> Please update the test app or fix the helper method. Thank you!

I made a mistake in the code, that's why it wasn't passing; fixed that. I've addressed all your other points, too. Please tell me if the updated approach with the `DEPRECATED` sentinel is how you do it, because there aren't any examples in the code to compare to (`navbar_options_resolve_deprecated()` is the only one).

shiny/ui/_input_select.py
```diff
@@ -2,7 +2,8 @@
 
 from __future__ import annotations
 
-from ..types import Jsonifiable
+from .._deprecated import warn_deprecated
+from ..types import Jsonifiable, MISSING_TYPE, DEPRECATED
```
schloerke, line 6: Lint
```suggestion
from ..types import DEPRECATED, MISSING_TYPE, Jsonifiable
```

PR #1948, opened 2025-03-31
[WIP] fix(DataGrid): Allow columns with empty titles to be used in `DataGrid`

### LINKED ISSUE #1844: Support for Empty Column in DataGrid
# Description:

I am trying to use the DataGrid component in the Shiny Express library to display a DataFrame that includes an empty column. However, it seems that the DataGrid does not support empty column names. Below is an example code demonstrating the issue:

```python
import pandas as pd

n = 6
df = pd.DataFrame(
    {
        "": range(n),
        "b": range(n, n * 2),
        "c": range(n * 2, n * 3),
        "d": range(n * 3, n * 4),
        "e": range(n * 4, n * 5),
    }
)

ui.h2("Data Frame with Styles applied to 4 cells")

@render.data_frame
def my_df():
    return render.DataGrid(df)

```

**Current Behavior**: The DataGrid fails to render the DataFrame when an empty column name is present.

**Expected Behavio**r: The DataGrid should support DataFrames with empty column names and render them appropriately.

ErdaradunGaztea:
Fixes: #1844 

The issue stems from the fact that TanStack uses column headers to create unique IDs, and an empty string is not a valid ID. Thus, my solution is to create a unique ID for the column with an empty header. Truth be told, there's no particular reason not to assign IDs equal to their corresponding headers for all the other columns, other that trying to change as little as possible.

Finding a unique ID is done through `findUnusedColumnName()` which was somewhat inspired by how {tibble} solves column name issues. In this case, the algorithm appends a space until it arrives at a string that doesn't exist in the list of column names. Again, the choice of space is arbitrary. What is important though is to keep in mind that this approach will fail if there is more than one column with an empty name. However, `DataGrid` currently doesn't work with duplicated names anyway, and generating unique IDs for all columns would probably be the solution.

Side note, I suspect that these large diffs in JS files are due to me using Windows, though I made sure that the output doesn't replace any LF with CRLF.


js/data-frame/index.tsx
```diff
@@ -1018,6 +1019,21 @@ const ShinyDataGrid: FC<ShinyDataGridProps<unknown>> = ({
...
 
+/**
+ * Solves the problem arising when there's more than one column with the same name (or when
+ * the column name is illegal, like in the case of an empty name) by returning a name with
+ * enough spaces appended so that it becomes unique.
+ * @param columns Names of existing columns.
+ * @param initName String to start the search with.
+ * @returns A string that does not exist in columns and thus can be used as column ID.
+ */
+function findUnusedColumnName(columns: ReadonlyArray<string>, initName = " "): string {
```
schloerke, line 1030: Can we move `initName` into the function as this will only hold true if the `initName` is `" "`. 

If other column names can be expanded, then they could both resolve to the same `"   "` string.

However, if we know that `""` is the only one being expanded and there are no duplicate names, then this function is only called once... which is safe.

Changing to something more specific...
```suggestion
function emptyColumnNameToId(columns: ReadonlyArray<string>): string {
  let initName = " ";
```

tests/playwright/shiny/bugs/1844-datagrid-empty-column/app.py
```diff
@@ -0,0 +1,22 @@
+import pandas as pd
+
+from shiny import App, Inputs, reactive, render, ui
+
+app_ui = ui.page_fluid(
+    ui.output_data_frame("df1")
```
schloerke, line 6: lint

```suggestion
    ui.output_data_frame("df1"),
```

tests/playwright/shiny/bugs/1844-datagrid-empty-column/app.py
```diff
@@ -0,0 +1,22 @@
...
+)
+
+
+def server(input: Inputs):
+    df = reactive.Value(pd.DataFrame({
+        "": [1, 2],
+        "A": [4, 5],
+        " ": [7, 8]
+    }))
```
schloerke, line 15: lint

```suggestion
    df = reactive.Value(
        pd.DataFrame(
            {
                "": [1, 2],
                "A": [4, 5],
                " ": [7, 8],
            }
        )
    )
```

schloerke (changes requested):
If the tanstack `id` is only browser specific, then this is great. 

If the id is used for updates to the server (ex column sorting input values), then we should raise an error server-side before it is even sent to the browser as the column names will not match.

------

Narwhals has a restriction of non-duplicated column names. It's fair if we add non-empty column names for tanstack.

tests/playwright/shiny/bugs/1844-datagrid-empty-column/test_1844_datagrid_empty_column.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import annotations
+
+from playwright.sync_api import Page
+
+from shiny.playwright import controller
+from shiny.run import ShinyAppProc
+
+
+def test_row_selection(page: Page, local_app: ShinyAppProc) -> None:
```
schloerke, line 9: Thank you for the test!!

-----------------

```suggestion
def test_empty_column_name(page: Page, local_app: ShinyAppProc) -> None:
```

schloerke:
Closing this PR as it is superseded by #2421. 

Thank you @ErdaradunGaztea for the Issue and PR!

PR #1950, opened 2025-04-01
chore(uvicorn): Remove uv source of uvicorn PR

schloerke:
With https://github.com/encode/uvicorn/pull/2602, the expected behavior is achieved and this PR's changes in `_main.py` may be reverted.

---------------------------

We are close to a release and can not have a uv source in the `pyproject.toml`. Therefore, we're removing it and extending the reload-excludes to exclude any file within  `shiny_bookmarks` up to 5 sub-folders deep. This should be a near 100% coverage.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -134,9 +134,6 @@ doc = [
     "griffe>=1.3.2",
 ]
 
-[tool.uv.sources]
-# https://github.com/encode/uvicorn/pull/2602
-uvicorn = { git = "https://github.com/schloerke/uvicorn", branch = "reload-exclude-abs-path" }
 
 [project.urls]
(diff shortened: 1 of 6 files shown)


schloerke:
Working reprex app
```python
import pandas as pd

from shiny import reactive
from shiny.express import app_opts, input, render, session, ui
from shiny.types import FileInfo

app_opts(bookmark_store="server")


@session.bookmark.on_bookmarked
async def _(url: str):
    await session.bookmark.update_query_string(url)


@reactive.effect
@reactive.event(input.file1, ignore_init=True)
async def _():
    await session.bookmark()


ui.input_file("file1", "Choose File", multiple=True)


@render.code
def parsed_file():
    return str(input.file1())
```

PR #1960, opened 2025-04-08
feat(bookmark): Use helper method to set default bookmark methods on an `App` object directly

schloerke:
Notes: `app` can be a `Starlette` or `shiny.App` object

### DIFF
--- a/shiny/__init__.py
+++ b/shiny/__init__.py
@@ -30,6 +30,9 @@ if _is_pyodide:
 else:
     from ._main import run_app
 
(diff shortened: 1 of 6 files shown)


schloerke:
After conversation with @jcheng5 , we should close this PR as the other external interfaces (Connect et al.) should only be able to expect an ASGI compatible interface. While we can traverse `starlette` routes, we should not limit the hosting environments to only being able to support `starlette` and Shiny. 

Therefore, it is best to just let the hosting environment set the default bookmark state _globally_ within the Shiny bookmark module.

Closing PR

PR #1961, opened 2025-04-08
feat(bookmark): Add `session.bookmark.update_on_session_disconnect()`

schloerke:
This PR is currently a POC w/ an example app.

Features:
* Added `on_session_disconnect()` that accepts a URL and immediately sends it to the browser.

### DIFF
--- a/shiny/api-examples/bookmark_on_session_disconnect/app.py
+++ b/shiny/api-examples/bookmark_on_session_disconnect/app.py
@@ -1,0 +1,86 @@
+import textwrap
+
+from starlette.requests import Request
+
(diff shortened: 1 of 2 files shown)


schloerke:
This method requires constant creation of the bookmark URL within the server to function as a last resort within the browser. 

I believe this it is the worst of all bookmarking worlds. 

If the user refreshes their page, they do not get the behavior. They **only** get a bookmarked site if the server has an error or the server session times out.  If the user performs any bookmarkable state action, their URL bar is not updated. 

Since the user is constantly creating bookmark URLs, the server will create _many_ bookmark folders that the user can not utilize.

---------------

I'm going to close this PR as I can't defend the final outcome.

PR #1978, opened 2025-04-22
feat(clientdata): `id` is now optional on `clientdata.output_*()` methods

cpsievert:
See the changes to `CHANGELOG` for a description of what this PR adds.

A minimal example:

```python
import numpy as np

from shiny.express import render, session, ui

ui.input_dark_mode()

@render.plot
def myplot():
    plt.figure()
    plt.hist(np.random.normal(size=100))  # type: ignore

    bg_color = session.clientdata.output_bg_color()
    plt.title(f"bgcolor: {bg_color}")
```

Screenshot 2025-04-22 at 6 18 30 PM

Screenshot 2025-04-22 at 6 18 52 PM


shiny/render/renderer/_renderer.py
```diff
@@ -404,3 +408,19 @@ def get_sync_fn(self) -> Callable[[], IT | None]:
...
+        if session is None:
+            yield
+            return
+
+        old_id = session._current_output_id
+        try:
+            session._current_output_id = self._renderer.output_id
+            yield
+        finally:
+            session._current_output_id = old_id
```
schloerke, line 426: I like this approach!, but I feel it is in the wrong place.

👀 for a different location.

schloerke (changes requested):

shiny/session/_session.py
```diff
@@ -188,6 +188,9 @@ class Session(ABC):
     user: str | None
     groups: list[str] | None
 
+    # Internal state for current_output_id()
+    _current_output_id: str | None = None
```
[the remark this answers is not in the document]
schloerke: Use the renderer as a placeholder

```suggestion
    _current_renderer: Renderer | None = None
```

CHANGELOG.md
```diff
@@ -7,13 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
+### New features
+
+* The `.output_*()` methods of the `ClientData` class (e.g., `session.clientdata.output_height()`) can now be called without an `id` inside a output renderer. (#1978)
+* The `Session` class gains a `.current_output_id()` method. It returns the ID of the currently executing output renderer (if any). (#1978)
```
schloerke, line 13:
```suggestion
```

shiny/render/renderer/_renderer.py
cpsievert: IIRC, we ended up not finding a better place?

cpsievert:
@schloerke my memory is a little fuzzy, but I think we had agreed this was OK as is?

schloerke:
It is good, but I do not want to alter `AsyncValueFn`. 

I'm almost done with a fix.

schloerke:
@cpsievert This is ready to go. Please take a look at the changes. 

Changes:
* Module support has been added / tested
* Module support has been added to `ClientData` 
* We only add the current output name into the `ClientData` class

schloerke:
@cpsievert There's an unrelated failure from shinytemplates (@karangattu is fixing) that is preventing auto merging. Please admin merge when ready.

cpsievert:
Ahh, thank you for the fixes!! 🍻

PR #1986, opened 2025-04-29
chore: update to latest shiny.js; cut autoresize logic that now comes bundled with shiny.js

cpsievert:
Handles fallout from https://github.com/rstudio/shiny/pull/4210 (that should get merged first, then this)


scripts/_pkg-sources.R
```diff
@@ -1,6 +1,6 @@
 list(
   bslib = "rstudio/bslib@main",
-  shiny = "rstudio/shiny@main",
+  shiny = "rstudio/shiny@feat/textAreaAutoResize",
```
cpsievert, line 3: Once https://github.com/rstudio/shiny/pull/4210 is merged, revert this change, and re-build dependencies

```suggestion
  shiny = "rstudio/shiny@main",
```

PR #2006, opened 2025-06-04
fix: Expanding timeout test

elnelson575:
Issue:
Some tests are timing out on GHA machines.

Fix: Expand how long we wait for these tests to complete.


schloerke (approved):
Thank you!

PR #2010, opened 2025-06-08
Value.get: Accept "default" parameter

pfalcon:
The idea is similar to e.g. Python's dict.get() method, which can take an optional default if a given key doesn't exist in the dict.

This change would allow to users to "bootstrap" otherwise circular dependencies in simple cases. For example, it allows to turn this code:

```

@render.express
def myrender():
    ui.input_select("sel", None, ["All", "So", "Much"], selected="All")
    input.sel()
```

which doesn't render anything nor produces errors (underneath it throws SilentException which isn't reported) into following code:

```

@render.express
def myrender():
    ui.input_select("sel", None, ["All", "So", "Much"], selected=input.sel.get("All"))
    input.sel.get("All")
```

which works as expected (initial selected value is "All", and if user chooses another, it's preserved across myrender() run).

### DIFF
--- a/shiny/reactive/_reactives.py
+++ b/shiny/reactive/_reactives.py
@@ -126,9 +126,15 @@ class Value(Generic[T]):
     def __call__(self) -> T:
         return self.get()
 
-    def get(self) -> T:
+    def get(self, default: T | MISSING_TYPE = MISSING) -> T:
         """
-        Read the reactive value.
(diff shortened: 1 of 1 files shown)


cpsievert:
Thanks for the PR, @pfalcon! There's a bit more nuance to fully address this feature, so I'll be taking it over to #2100

PR #2020, opened 2025-06-27
Grab new shiny JS/CSS assets

cpsievert:
Covers this diff https://github.com/rstudio/shiny/compare/ca41c08...460a93a


shiny/ui/_input_action_button.py
```diff
@@ -64,7 +64,15 @@ def input_action_button(
...
-        None if icon is None else " ",
+        # The separator element helps us distinguish between the icon and label
+        # when dynamically updating the button/link (see rstudio/shiny#4242)
+        tags.span(class_="shiny-icon-separator") if icon else None,
+        # Adds a space between the icon and label
+        # TODO: this space gets removed when icon/label are updated dynamically,
+        # which is not ideal. The 'right' way to do this would be to either
+        # add a CSS class to the separator element, or wrap both the icon and
+        # label in a container element.
+        " " if icon and label else None,
```
cpsievert: This obviously isn't ideal, and long-term this isn't the approach to spacing that we'd want, but I'm thinking I'll just keep it to retain the current behavior.

cpsievert: Ugh, the old JS logic would (unconditionally, for both button and links!!) add a space when updating to either the icon or label

https://github.com/rstudio/shiny/blob/33dc41c/srcts/src/bindings/input/actionbutton.ts#L73

This is a big bummer because, unless we add the space in the server-side update, the space gets removed on update 😭 

cc @gadenbuie

cpsievert: Before we make a decision here, I'm gonna look into https://github.com/rstudio/shiny/pull/4249 as a possible direction

PR #2025, opened 2025-07-08
fix(examples): Remove stat_smooth to fix example tests

karangattu:
The issue happened because `stat_smooth` was trying to draw these trend lines, but some groups didn't have enough data points to create a meaningful line.


karangattu:
This might have fixed itself and maybe not required anymore. Hence closing

PR #2038, opened 2025-07-23
feat(api): Deprecate `panel_well()` in favor of `card()`

### LINKED ISSUE #1704: UserWarning incorrectly emitted when using negative column widths for ui.layout_columns
**Minimal reproducible example**
```py

app_ui = ui.page_fluid(
    ui.layout_columns(
        ui.output_text("greeting"),
        col_widths=(-1, 1),
    ),
)

def server(inputs, outputs, session):
    @render.text
    def greeting():
        return "Hello, world!"

app = App(app_ui, server)
```
**Output**
UserWarning: More column widths than children at breakpoint 'sm', extra widths will be ignored.

**Expected output**
No warning, as negative widths shouldn't be counted when comparing the number of widths vs. the number of children

cpsievert:
This PR also fixes #1704 (since I wanted to use `layout_columns()` in the updated examples, but it was preventing tests from passing)


tests/playwright/examples/example_apps.py
```diff
@@ -91,6 +91,7 @@ def get_apps(path: str) -> typing.List[str]:
     "FutureWarning: use_inf_as_na option is deprecated",
     "pd.option_context('mode.use_inf_as_na",  # continutation of line above,
     "RuntimeWarning: invalid value encountered in dot",  # some groups didn't have enough data points to create a meaningful line
+    "UserWarning: More column widths than children at breakpoint",
```
cpsievert: @karangattu / @schloerke running `shiny run examples/airmass/app.py` now gives a (known) false positive warning:

```python
/Users/cpsievert/github/py-shiny/shiny/ui/_layout_columns.py:209: UserWarning: More column widths than children at breakpoint 'sm', extra widths will be ignored.
```

This change here doesn't seem to be the correct way to ignore the warning -- any idea what I'm doing wrong?

karangattu: I suspect it needs the entire warning message `UserWarning: More column widths than children at breakpoint 'sm', extra widths will be ignored.`

karangattu: nevermind that is not the issue... investigating!

karangattu: The issue was that the server side showed this warning
```sh
UserWarning: More column widths than children at breakpoint 'sm', extra widths will be ignored.
  warn(
  ```
  with `warn(` on the newline.

karangattu: ~We can make this more robust by keeping the entire message and placing the `warn` on a new line, ensuring future `warn`s aren’t ignored.~ the error filtering logic checks each line individually for substrings

cpsievert: Thanks @karangattu, good to know for the future. This got me curious enough that I just decided to fix the false-positive warning in [64f65f9](https://github.com/posit-dev/py-shiny/pull/2038/commits/64f65f998d99bf262e8a9550a0be7eb4fa42f63c)

PR #1711, opened 2024-10-01
[WIP] chore!: Drop python 3.8. Add python 3.13

schloerke:

PR #2040, opened 2025-07-24
chore(py3.13): Update changelog and test on less jobs

schloerke:
Followup to #1711


schloerke:
Goal of PR achieved. Merging

PR #705, opened 2023-08-24
Docstrings for module.ui and module.server

jcheng5:

PR #1974, opened 2025-04-14
docs: docstring for core shiny.modules.ui and shiny.module.server

chendaniely:
Adding missing docstrings for `shiny.module`

PR #2044, opened 2025-07-29
docs: update module server and ui to incorporate #705

chendaniely:
incorporated the text from #705 and #1974 to improve on the module server and ui docstrings.

closes #705


cpsievert:
Thanks!

Issue #2052, opened 2025-08-06
Selectize options are overwritten or silently dropped


elnelson575:
During our investigation of https://github.com/posit-dev/py-shiny/issues/1703 we discovered that there are a subset of issues arising from how and in what order we set `options` in `update_selectize`.

For example, in this example, the user-provided options are dropped silently:
```
from shiny import reactive
from shiny.express import ui

ui.input_selectize(
    "foo",
    "Foo",
    choices=[],
    multiple=True,
)


@reactive.effect
def _():
    ui.update_selectize(
        "foo",
        choices=[f"Foo & {i}" for i in range(1000)],
        options={"placeholder": "Select an option"},
    )
```

Two permutations that need investigation are:
`update_selectize(..., server=True, options = {"placeholder": "Select an option"})`
and
`update_selectize(..., options = {"placeholder": "Select an option"})`

**UPDATE:**
Based on the difference in behavior between server and client, it seems like this is an us issue.

We need to decide what the behavior should be when you specify options in `update_selectize`. 
Should this add additional options to some default list? 
Replace the list entirely?
We should investigate what the current behavior is in R Shiny. This should be consistent
We were looking at `selectInput.ts` 289 and how that differs from the preceding code.

elnelson575:
Initial fix for overwriting: move `cfg` piece below we check if `server=True`

cpsievert:
@elnelson575 #2053 will fix the silent dropping issue -- will you create another issue for the no-x/double-x thing?

elnelson575:
@cpsievert New issue created here: https://github.com/posit-dev/py-shiny/issues/2055

Issue #1703, opened 2024-09-25
Passing options to update_selectize() inside a module breaks update


thohan88:
Based on the [update_selectize example](https://shiny.posit.co/py/api/core/ui.update_selectize.html) from the docs.

When put inside a module, the update logic breaks if `options` are passed to `update_selectize()`. No choices are displayed.

```python
from shiny import App, module, reactive, ui

@module.ui
def reprex_selectize_ui():
    return ui.input_selectize("x", "Server side selectize", choices=[], multiple=True)


@module.server
def reprex_selectize_server(input, output, session, starting_value = 0):
    @reactive.effect
    def _():
        ui.update_selectize(
            "x", 
            choices=[f"Foo {i}" for i in range(10000)],
            server=True,
            # options={"placeholder": "Search"},
            # If uncommented, this breaks the update, and no choices are displayed
        )

app_ui = ui.page_fluid(
    reprex_selectize_ui("reprex_selectize")
)


def server(input, output, session):
    reprex_selectize_server("reprex_selectize")


app = App(app_ui, server)

PR #2053, opened 2025-08-06
fix: Fix 1703 Options selectize not working in modules

elnelson575:
Fixes an issue as reported in https://github.com/posit-dev/py-shiny/issues/1703 where passing `options` into `update_selectize()` when `server=True` breaks the selectize menu (choices do not load, search/filtering does not occur).

Additional problems discovered while addressing this are documented here: https://github.com/posit-dev/py-shiny/issues/2052


cpsievert:
Thanks! Looks good, please update the CHANGELOG and maybe worth a test?

tests/playwright/shiny/components/selectize/app.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@module.ui
+def reprex_selectize_ui():
+    return ui.input_selectize("x", "Server side selectize", choices=[], multiple=True)
+
+
+@module.server
+def reprex_selectize_server(
+    input: Inputs, output: Outputs, session: Session, starting_value: int = 0
```
cpsievert:
```suggestion
    input: Inputs, output: Outputs, session: Session
```

elnelson575:
We're moving this over here and adding a test for it.

Initial fix for overwriting: 
move 
```  
if options is not None:
        cfg = tags.script(
            json.dumps(options),
            type="application/json",
            data_for=resolve_id(id),
            data_eval=json.dumps(extract_js_keys(options)),
        )
        session.send_input_message(id, {"config": cfg.get_html_string()})
```
to before
``` 
    if not server:
        return update_select(
            id, label=label, choices=choices, selected=selected, session=session
        )
```
in `_input_update.py`

CHANGELOG.md
```diff
@@ -41,6 +41,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* Fixed missing selectize options when using a module and specifying `options`. (#1703)
```
```suggestion
* Fixed numerous issues related to programmatically updating selectize options. (#2053)
   * `update_selectize(options=...)` no longer gets ignored when `server=False` (the default).
   * `update_selectize(options=...)` now works as expected in a module.
```

PR #2057, opened 2025-08-12
fix: Address datepicker issues

### LINKED ISSUE #1689: input_date component: value and max cannot be the same
If I set max and value to the same value the date string is not rendered, just an empty input field. It seems max date needs to be bigger than value but the docs say:

"Max: The maximum allowed date."

Python 3.11.6
shiny==1.1.0

min_date = datetime.now()
max_date = min_date + relativedelta(days=10)

ui.input_date(
    "start_date_picker",
    "Start Date:",
    value=max_date,
    min=min_date,
    max=max_date,
    format="dd.mm.yyyy",
    language="de"
)

### LINKED ISSUE #1713: [Improvement] Allow None Value for input_date() and update_date()
Currently, when setting the `value `parameter of `input_date()` to `None`, the default behavior in `py_shiny` is to display the current date, rather than an empty input. This differs from the behavior in `R's Shiny`, where an empty input is shown when `NULL `is specified.

While a workaround exists by setting the value parameter to `pd.NaT`, this approach does not apply to` update_date()`. If the value parameter for `update_date()` is explicitly set to `None`, no changes occur, and the same workaround is ineffective.

**Proposed Improvement**

It would be beneficial to allow users to define a `None `value for both `input_date()` and `update_date()` components. This change would align `py_shiny` more closely with the expected behavior in `R's Shiny` and improve user experience.

elnelson575:
Closes #1713
Closes #1689


shiny/ui/_input_date.py
```diff
@@ -310,6 +310,9 @@ def _date_input_tag(
 def _as_date_attr(x: Optional[date | str]) -> Optional[str]:
     if x is None:
         return None
-    if isinstance(x, date):
-        return str(x)
-    return str(date.fromisoformat(x))
+    if isinstance(x, str):
+        if x == "":
+            return x
```
cpsievert: Let's make sure to document this behavior in the docstring

shiny/ui/_input_date.py
```diff
@@ -310,6 +310,9 @@ def _date_input_tag(
 def _as_date_attr(x: Optional[date | str]) -> Optional[str]:
     if x is None:
         return None
-    if isinstance(x, date):
-        return str(x)
-    return str(date.fromisoformat(x))
+    if isinstance(x, str):
+        if len(x) == 0:
+            return x
```
cpsievert, line 315: Let's make sure to document this behavior

elnelson575: Added the empty string bit into the doc string, documented additional behavior in `update_date` and `input_date`. Notably, inputs appear to be sticky on the Bootstrap Datepicker side. That is, if we send an empty string, the value disappears, but the display date remains.

tests/playwright/shiny/components/datepicker/app.py
```diff
@@ -0,0 +1,93 @@
...
+from dateutil.relativedelta import relativedelta
+
+from shiny import App, Inputs, Outputs, Session, reactive, render, ui
+
+min_date = date.fromisoformat("2011-11-04")
+max_date = min_date + relativedelta(days=10)
+
+# Our input requires strings to be in the format "YYYY-MM-DD"
+str_min = min_date.strftime("%Y-%m-%d")
+str_max = max_date.strftime("%Y-%m-%d")
```
cpsievert:
```suggestion
```

shiny/ui/_input_date.py
elnelson575: Fixed

shiny/ui/_input_update.py
```diff
@@ -475,21 +474,36 @@ def update_date(
 
     Note
     ----
-    {note}
+    A special value of `""` (an empty string) can be used to clear the value, min, and
```
cpsievert: Should keep this note and add this note to the parameter description

```suggestion
    {note}
```

shiny/ui/_input_update.py
```diff
@@ -464,7 +464,6 @@ def update_date(
         An input label.
     value
         The starting date. Either a `date()` object, or a string in yyyy-mm-dd format.
-        If ``None`` (the default), will use the current date in the client's time zone.
```
cpsievert, line 467: Add this back

shiny/ui/_input_date.py
```diff
@@ -310,6 +310,9 @@ def _date_input_tag(
...
+        x = date.fromisoformat(x)
+    # Using strftime here to ensure we just just a date, regardless of if H:M:S are included
```
cpsievert:
```suggestion
    # Ensure we return a date (not datetime) string
```

tests/playwright/shiny/components/date_range/app.py
cpsievert, line 1: Move to `tests/playwright/shiny/inputs`

shiny/ui/_input_update.py
elnelson575: This is the update, so we're not supplying a default update value (ex. if None is passed to update, it will be dropped) Correct?

cpsievert: Ah yes, new version looks good 👍

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improvements
 
+* Added support for blank values in `input_date()`, `input_date_range()`, `update_date()`, and `update_date_range()` for values, mins, and maxes to allow for providing blank input boxes and removing previously set values. (#1713) (#1689)
```
```suggestion
* `input_date()`, `input_date_range()`, `update_date()`, and `update_date_range()` now supports `""` for values, mins, and maxes. In this case, no date will be specified on the client. (#1713) (#1689)
```

PR #2063, opened 2025-08-19
fix: Updating nav cards to have full screen option

elnelson575:


shiny/ui/_navs.py
```diff
@@ -708,6 +711,7 @@ def navset_card_tab(
     sidebar: Optional[Sidebar] = None,
     header: TagChild = None,
     footer: TagChild = None,
+    full_screen: bool = False,
```
cpsievert: Bring up closer to `sidebar`

PR #2067, opened 2025-08-21
test(narwhals): Update Playwright test path in Makefile

karangattu:
This pull request changes the path for Playwright tests in the `narwhals-test-integration` target to point to the test that tests shiny integration with various data processing libraries.
This way it will make this test suite more robust.


schloerke:
cc @FBruzzesi

FBruzzesi:
Thanks for the ping @schloerke - I will close the issue in Narwhals. Just FYI I think it's been a while anyway that we don't get a failure. Not sure if this tells you anything useful though 😇

PR #2069, opened 2025-08-25
fix: include_css and include_js can use files in same dir

elnelson575:
Part 1 of 3:
`include_js` and `include_css` should be able to both load files from the same dir in the same app.


tests/playwright/shiny/bugs/2061-css-js-inclusion/www/customjs.js
```diff
@@ -0,0 +1 @@
+console.log("heyo")
```
cpsievert: Have this append a message in the UI, like 'Passed'. Then have a test to assert the message has been added.

tests/playwright/shiny/bugs/2061-css-js-inclusion/app.py
```diff
@@ -0,0 +1,27 @@
...
+from shiny import App, Inputs, Outputs, Session, ui
+
+js_file = Path(__file__).parent / "js" / "customjs.js"
+css_file = Path(__file__).parent / "css" / "style.css"
+
+# Define the UI
+app_ui = ui.page_fluid(
+    ui.include_css(css_file, method="link_files"),
+    ui.include_js(js_file, method="link_files"),
+    ui.h1("Simple Shiny App with External CSS"),
```
cpsievert, line 23: If you wanted to also test that `link_files` does indeed copy other files (and are available for request, you coudl do:

```suggestion
    ui.tags.script(src="js/customjs2.js"),
    ui.h1("Simple Shiny App with External CSS"),
```

Then have the script do something similar to the other script (and have the test check the DOM state)

tests/playwright/shiny/bugs/2061-css-js-inclusion/js/customjs.js
```diff
@@ -0,0 +1,4 @@
+const newParagraph = document.createElement('p');
+newParagraph.textContent = 'Heyo!';
+const bodyElement = document.body;
+bodyElement.appendChild(newParagraph);
```
cpsievert:
```suggestion
document.body.appendChild(newParagraph);
```

tests/playwright/shiny/bugs/2061-css-js-inclusion/css/style.css
```diff
@@ -0,0 +1,19 @@
+@import url("more.css");
```
cpsievert: If you do the `ui.script()` thing, you could get rid of this (to avoid confusion as to why it's here)

tests/playwright/shiny/bugs/2061-css-js-inclusion/app.py
elnelson575: Do you mean define an HTML dependency directly? I don't see a ui.script() function anywhere in the docs.

elnelson575: Ah, got it, tags.script. 👍🏻

tests/playwright/shiny/bugs/2061-css-js-inclusion/app.py
```diff
@@ -0,0 +1,33 @@
+from pathlib import Path
+
+from shiny import App, Inputs, Outputs, Session, ui
+
+js_file = Path(__file__).parent / "js" / "customjs.js"
+# We don't know what the temp dir will be called until we make it and that's what we need to reference
+# in the app (ex. we can't just reference where it is pre-copy)
+js2_file = (
+    ui.include_js(js_file, method="link_files").attrs["src"][:-11] + "customjs2.js"
+)
```
cpsievert: Hmm, why this additional `include_js()` call? I think you can just have the relative file path (like I suggested)?

elnelson575: With:
```
ui.include_css(css_file, method="link_files"),
ui.include_js(js_file, method="link_files"),
ui.tags.script(src="js/customjs2.js"),
```
like you suggested above, it attempts to source the file from the root of where the app is running, vs. the root of the temp_dir. We don't want to run the app from inside the temp dir, but we also do not know the temp dir before it's created, hence the call to get the temp dir. Unless there's something I'm not thinking of to get around that?
See below:
Screenshot 2025-08-28 at 9 00 40 PM
Screenshot 2025-08-28 at 9 00 11 PM

cpsievert: Ahh, right, ok. How about we do things like this?

```python
custom_css = ui.include_css(
    Path(__file__).parent / "css" / "style.css",
    method="link_files",
)

custom_js = ui.include_js(
    Path(__file__).parent / "js" / "customjs.js",
    method="link_files",
)

# path where the JS file's parent directory is mounted
href = custom_js.get_dependencies()[0].source_path_map()["href"]

# Define the UI
app_ui = ui.page_fluid(
    custom_css,
    custom_js,
    ui.tags.script(src=href + "/customjs2.js"),
    ui.h1("Simple Shiny App with External CSS"),
    ui.div(
        ui.p("This is a simple Shiny app that demonstrates ui.include_css()"),
        ui.p("The styling comes from an external CSS file!"),
        class_="content",
    ),
)
```

elnelson575: Sure, just updated.

shiny/ui/_include_helpers.py
```diff
@@ -36,10 +36,10 @@ def include_js(
     method
         One of the following:
 
-        * ``"link"`` is the link to the CSS file via a :func:`~shiny.ui.tags.link` tag. This
+        * ``"link"`` is the link to the JS file via a :func:`~shiny.ui.tags.link` tag. This
```
cpsievert:
```suggestion
        * ``"link"`` is the link to the JS file via a :func:`~shiny.ui.tags.script` tag. This
```

shiny/ui/_include_helpers.py
```diff
@@ -48,7 +48,7 @@ def include_js(
           example, if the app's source is located at ``/app/app.py``, then ``path``
           should be somewhere like ``/app/css/custom.css`` (and all the other relevant
           accompanying 'safe' files should be located under ``/app/css/``).
-        * ``"inline"`` is the inline the CSS file contents within a
+        * ``"inline"`` is the inline the JS file contents within a
           :func:`~shiny.ui.tags.style` tag.
```
cpsievert:
```suggestion
          :func:`~shiny.ui.tags.script` tag.
```

shiny/ui/_include_helpers.py
```diff
@@ -36,10 +36,10 @@ def include_js(
...
         One of the following:
 
-        * ``"link"`` is the link to the CSS file via a :func:`~shiny.ui.tags.link` tag. This
+        * ``"link"`` is the link to the JS file via a :func:`~shiny.ui.tags.link` tag. This
           method is generally preferable to ``"inline"`` since it allows the browser to
           cache the file.
-        * ``"link_files"`` is the same as ``"link"``, but also allow for the CSS file to
+        * ``"link_files"`` is the same as ``"link"``, but also allow for the JS file to
           request other files within ``path``'s immediate parent directory (e.g.,
           ``@import()`` another file). Note that this isn't the default behavior because
```
cpsievert:
```suggestion
          ``import`` another file). Note that this isn't the default behavior because
```

shiny/ui/_include_helpers.py
```diff
@@ -48,7 +48,7 @@ def include_js(
           example, if the app's source is located at ``/app/app.py``, then ``path``
           should be somewhere like ``/app/css/custom.css`` (and all the other relevant
```
cpsievert:
```suggestion
          should be somewhere like ``/app/js/custom.js`` (and all the other relevant
```

cpsievert (approved):

cpsievert:
@elnelson575 for future reference, it would be great if all PRs include a `Closes #n` as well as an update to the `CHANGELOG.md`

PR #2072, opened 2025-08-29
fix(deps): Restrict `pytest-rerunfailures` version to below 16

schloerke:
Related: https://github.com/pytest-dev/pytest-rerunfailures/issues/302


schloerke:
Merging as PRs are in better shape and I don't want to wait

PR #2078, opened 2025-09-03
fix: Clean up tmp files for HTML dependencies after app closes

elnelson575:


shiny/_app.py
```diff
@@ -214,6 +215,18 @@ def __init__(
                 cast("Tag | TagList", ui), lib_prefix=self.lib_prefix
             )
 
+    def __del__(self):
```
elnelson575, line 219: This is the part distinct to this PR

PR #2070, opened 2025-08-26
fix: Make outside users able to read tmp files

elnelson575:
Closes: #2061 
Related to: #2079 

This PR updates the temp directory into which included files are saved to resolve issues with file permissions on multi-user systems.

Testing:
- [x] Include Files Playwright test
- [x] Not Include Files Playwright test

PR #2079, opened 2025-09-03
fix: Remove tmp files after app closes

elnelson575:
Related to: #2061 and #2070 

This PR removes dirs created for JS and CSS inclusions from the tmp directory on application close.
example:
Screenshot 2025-09-04 at 10 23 43 AM


CHANGELOG.md
```diff
@@ -53,6 +53,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* Fixed several issues related to including custom js and css files
+    * JS and CSS file loaded with `include_js()` or `include_css()` will be loaded into a new temporary directory and cleaned up properly on program exit. (#2061)
+    * JS and CSS files in the same directory will be loaded correctly. (#729)
```
cpsievert: I think we remove this and add a new one at the bottom of this section

wch:
There can be multiple Python processes running the same app. For example, Connect and shinyapps.io have settings to control when to spawn a new process.

If I'm correctly understanding how temp dirs are created and how this PR works: If a temp directory is shared across multiple processes, then when the load decreases and the server stops one of the processes, then it would cause the shared temp directory be deleted, which could cause problems for the other process.

There's also the case of multiple App objects in the same process. Some of the pages on the Shiny web site have multiple Shinylive apps, and those all run in the same process, so if the user were to, say, edit and restart one of those apps, it would cause the shared temp dir to be deleted, which could cause problems.

cpsievert:
@wch thanks for pointing this out. I suppose we could scope the `include_js()`/`include_css()` tempdirs to the process-level, but doing that also feels more trouble than it's worth at this point. I think we should probably just drop this and move on, sorry @elnelson575.

PR #2094, opened 2025-10-08
chore(python version): Drop Python 3.9 support and add Python 3.14 support

karangattu:
Update minimum required Python version to 3.10 in pyproject.toml and scripts. Add Python 3.14 to supported/tested versions in CI workflows and remove Python 3.9 from all relevant matrices and classifiers. Remove test skip logic for Python 3.9 where no longer needed.

Reference: https://www.python.org/downloads/release/python-3140/

- [x] `pyogrio` has a release that supports 3.14; Support added in https://github.com/geopandas/pyogrio/commit/e88af711a6bd893bca14f58686bc65fef30bace7; Support for min version 3.10 added in https://github.com/geopandas/pyogrio/commit/56a5961e32fa37bc72def28a654c6bfec0285c27

### DIFF
--- a/tests/pytest/test_render_data_frame.py
+++ b/tests/pytest/test_render_data_frame.py
@@ -1,4 +1,4 @@
-from typing import Any, cast
+from typing import Any, Callable, cast
 
 import pandas as pd
 import pytest
@@ -96,8 +96,6 @@ def test_as_selection_modes():
 
 @pytest.mark.asyncio
 async def test_as_data_frame_none():
-
-    from typing import Any, Callable
 
     async def expect_data_is_not_none_error(fn: Callable[[], Any]):
         fn_async = wrap_async(fn)


.github/workflows/pytest.yaml
```diff
@@ -17,13 +17,14 @@ jobs:
     strategy:
       matrix:
         # "3.10" must be a string; otherwise it is interpreted as 3.1.
-        python-version: ["3.13", "3.12", "3.11", "3.10", "3.9"]
+        python-version: ["3.14", "3.13", "3.12", "3.11", "3.10"]
         os: [ubuntu-latest, windows-latest, macOS-latest]
         exclude:
+          - python-version: ${{ github.event.pull_request.draft && '3.14' }}
```
schloerke:
```suggestion
```

.github/workflows/pytest.yaml
```diff
@@ -108,19 +108,16 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.13", "3.12", "3.11", "3.10", "3.9"]
+        python-version: ["3.14", "3.13", "3.12", "3.11", "3.10"]
         browser: ["chromium", "firefox", "webkit"]
         exclude:
+          - python-version: ${{ github.event.pull_request.draft && '3.14' }}
```
schloerke:
```suggestion
```

.github/workflows/pytest.yaml
```diff
@@ -157,13 +154,14 @@ jobs:
...
```
schloerke:
```suggestion
```

.github/workflows/pytest.yaml
```diff
@@ -222,13 +218,14 @@ jobs:
...
```
schloerke:
```suggestion
```

scripts/generate-imports.py
```diff
@@ -5,8 +5,8 @@
 from types import ModuleType
 from typing import Union, cast
 
-if sys.version_info < (3, 9):
-    raise RuntimeError("This script requires Python 3.9 or later")
+if sys.version_info < (3, 10):
+    raise RuntimeError("This script requires Python 3.10 or later")
```
schloerke:
```suggestion
if sys.version_info < (3, 9):
    raise RuntimeError("This script requires Python 3.9 or later")
```

PR #2099, opened 2025-10-15
feat: Add `input_submit_textarea()` and `update_submit_textarea()`

cpsievert:
Related https://github.com/rstudio/bslib/pull/1204

Adds `input_submit_textarea()`, a new input control that's akin to `input_textarea()`, but delays the sending of an input value until a corresponding button is pressed.

Here is a hello world example:

```python
import time

ui.input_submit_textarea("text", placeholder="Enter some input...")


@render.text
def value():
    if "text" in input:
        # Simulate processing time
        time.sleep(2)
        return f"You entered: {input.text()}"
    else:
        return "Submit some input to see it here."
```

### DIFF
--- a/shiny/playwright/controller/_input_fields.py
+++ b/shiny/playwright/controller/_input_fields.py
@@ -213,6 +213,28 @@ class _ExpectAutocompleteAttrM:
         )
 
 
+class _ExpectRowsAttrM:
+    """A mixin class for the rows attribute."""
+
+    def expect_rows(
+        self: UiBaseP,
+        value: AttrValue,
+        *,
+        timeout: Timeout = None,
+    ) -> None:
+        """
+        Expect the `rows` attribute to have a specific value.
+
+        Parameters
+        ----------
+        value
+            The expected value of the `rows` attribute.
+        timeout
+            The maximum time to wait for the expectation to be fulfilled. Defaults to `None`.
+        """
+        _expect_attribute_to_have_value(self.loc, "rows", value=value, timeout=timeout)
+
+
 class InputText(
     _SetTextM,
     _ExpectTextInputValueM,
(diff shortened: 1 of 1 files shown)


shiny/playwright/controller/_input_fields.py
```diff
@@ -966,3 +966,110 @@ def expect_autoclose(
...
+        Clicks the submit button.
+
+        Parameters
+        ----------
+        timeout
+            The maximum time to wait for the click. Defaults to `None`.
+        """
+        self.loc_button.click(timeout=timeout)
+
+    def expect_rows(self, value: AttrValue, *, timeout: Timeout = None) -> None:
```
karangattu: since `expect_rows()` is duplicated in both `InputSubmitTextarea` and `InputTextArea`, we could make it a mixin similar to this
```python
class _ExpectRowsAttrM:
    """A mixin class for the rows attribute."""

    def expect_rows(
        self: UiBaseP,
        value: AttrValue,
        *,
        timeout: Timeout = None,
    ) -> None:
        """
        Expect the `rows` attribute to have a specific value.

        Parameters
        ----------
        value
            The expected value of the `rows` attribute.
        timeout
            The maximum time to wait for the expectation to be fulfilled. Defaults to `None`.
        """
        _expect_attribute_to_have_value(self.loc, "rows", value=value, timeout=timeout)
```
so then the classes will look like
```python
  class InputTextArea(
      _SetTextM,
      WidthContainerStyleM,
      _ExpectTextInputValueM,
      _ExpectPlaceholderAttrM,
      _ExpectAutocompleteAttrM,
      _ExpectSpellcheckAttrM,
      _ExpectRowsAttrM,
      UiWithLabel,
  ):
      """Controller for :func:`shiny.ui.input_text_area`."""
      .
      .
      .
  class InputSubmitTextarea(
      _SetTextM,
      WidthContainerStyleM,
      _ExpectTextInputValueM,
      _ExpectPlaceholderAttrM,
      _ExpectRowsAttrM,
      UiWithLabel,
  ):
      """Controller for :func:`shiny.ui.input_submit_textarea`."""
```

gadenbuie (approved):
Looks great, thank you!

PR #2155, opened 2026-01-21
feat: Add Toolbar

elnelson575:
Add Toolbar Capabilities, featuring:
1) Toolbar container
2) Toolbar Button
3) Toolbar Select
4) Toolbar Divider/Spacer
5) Input bindings for 2 & 3
6) Playwright controllers for the above

Port of: https://github.com/rstudio/bslib/pull/1276
For minor label & icon fixes: https://github.com/rstudio/bslib/pull/1285 was merged in to bslib.

**Demo:**

https://github.com/user-attachments/assets/30388497-b35e-4c86-9c0f-da9b1445779f

**To Do:**

- [x] Port toolbar container
    - [x] Add special cases to the docs
      - Checked each of the example cases
	- [x] Unit tests
	- [x] Playwright tests
- [x] Port toolbar divider & spacer
	- [x] Unit tests
	- [x] Playwright tests
- [x] Port toobar button
	- [x] Unit tests
	- [x] Playwright tests
- [x] Port toobar select
	- [x] Unit tests
	- [x] Playwright tests
- [x] Add special use cases app (ex. toolbars in labels, toolbars in text inputs, etc.)
	- [x] Unit tests
	- [x] Playwright tests
- [x] Playwright controllers
- [x] Accessibility check on kitchen sink
- [x] Review generated docs including how they render

Outside of this PR:
- Add documentation page
- Draft language-agnostic blog post


docs/_quartodoc-core.yml
```diff
@@ -174,6 +182,8 @@ quartodoc:
         - ui.update_action_button
         - ui.update_action_link
         - ui.update_task_button
+        - ui.update_toolbar_input_button
+        - ui.update_toolbar_input_select
```
gadenbuie, line 186: I'd personally lean toward putting this with the other Toolbar functions, but I'm realizing we're not entirely consistent about this currently (see navbars which has one update function in its section and another in the update section).

docs/_quartodoc-testing.yml
```diff
@@ -46,6 +46,8 @@ quartodoc:
         - playwright.controller.InputTaskButton
         - playwright.controller.InputText
         - playwright.controller.InputTextArea
+        - playwright.controller.ToolbarInputButton
+        - playwright.controller.ToolbarInputSelect
```
gadenbuie, line 50: Do we not have `playwright.controller.Toolbar`? Should we?

shiny/ui/_input_code_editor_bundle.py
```diff
@@ -4,7 +4,7 @@
 
 from typing import Literal
 
-VERSION_PRISM_CODE_EDITOR = "0.9.0.9002"
+VERSION_PRISM_CODE_EDITOR = "0.10.0.9000"
```
gadenbuie, line 7: Oops, this isn't the right prism version. Can you open an issue to remind me to fix this?

shiny/ui/_toolbar.py
```diff
@@ -0,0 +1,897 @@
...
+    _render_choices,
+)
+from ._tooltip import tooltip as ui_tooltip
+from .css import CssUnit, as_css_unit
+
+
+@add_example()
+@add_example("app-numeric-input.py")
+@add_example("app-text-area.py")
+@add_example("app-submit-textarea.py")
```
gadenbuie: I like that there are a variety of examples, but can you please check that this looks okay when rendered in the docs site?

shiny/ui/_toolbar.py
```diff
@@ -0,0 +1,897 @@
...
+    )
+    ```
+    """
+    width_css = as_css_unit(width) if width is not None else None
+    gap_css = as_css_unit(gap) if gap is not None else None
+
+    # Build style string manually to preserve CSS custom property names with underscores
+    style_parts: list[str] = []
+    if gap_css is not None:
+        style_parts.append(f"--_divider-gap: {gap_css}")
```
gadenbuie, line 144: Can you just call `as_css_unit(gap)` inline here?

shiny/ui/_toolbar.py
```diff
@@ -0,0 +1,897 @@
...
+    return div(
+        {
+            "class": "bslib-toolbar-divider",
+            "aria-hidden": "true",
+            "style": style_str,
+        },
+    )
+
+
+@add_example()
```
gadenbuie, line 159: I'm surprised this doesn't error because I don't see an `api-examples/toolbar_spacer/app.py` in the changes. I think it's worth moving the example app from the docstring into a small `app.py` file.

shiny/ui/_toolbar.py
```diff
@@ -0,0 +1,897 @@
...
+        ui.toolbar_input_button(id="help", label="Help", icon=icon_svg("circle-question")),
+        align="left",
+        width="100%"
+    )
+    ```
+    """
+    return div({"class": "bslib-toolbar-spacer", "aria-hidden": "true"})
+
+
+@add_example()
```
gadenbuie, line 189: Same as above

shiny/ui/_toolbar.py
```diff
@@ -0,0 +1,897 @@
...
+        id="edit",
+        label="Edit",
+        show_label=True,
+        icon=faicons.icon_svg("pencil")
+    )
+    ```
+    """
+
+    if "_add_ws" not in kwargs:
+        kwargs["_add_ws"] = True
```
gadenbuie, line 271: Why defaulting to `_add_ws = True`?

shiny/ui/_toolbar.py
```diff
@@ -0,0 +1,897 @@
...
+
+    # Compute show_label default: True if no icon, False if icon is provided
```
gadenbuie, line 273: I think all the comments in this function are very close to repeating the code verbatim and can be dropped

gadenbuie: This seems to be true for the rest of the file, which could use a comment clean-up pass

shiny/ui/_toolbar.py
```diff
@@ -0,0 +1,897 @@
...
+        return x
+    if isinstance(x, (list, tuple)):
+        return " ".join(_extract_text(item) for item in x)
+    if isinstance(x, Tag):
+        if hasattr(x, "children"):
+            return _extract_text(x.children)
+    return ""
+
+
+@add_example()
```
gadenbuie, line 363: Same as above

shiny/ui/_toolbar.py
```diff
@@ -0,0 +1,897 @@
...
+            "showLabel": show_label,
+            "icon": icon_processed,
+            "disabled": disabled,
+        }
+    )
+
+    session.send_input_message(id, message)
+
+
+@add_example()
```
gadenbuie, line 438: Same

shiny/ui/_toolbar.py
```diff
@@ -0,0 +1,897 @@
...
+        _render_choices(choices_normalized, selected),
+        {
+            "id": select_id,
+            "class": "form-select form-select-sm bslib-toolbar-select",
+            "data-shiny-no-bind-input": True,
+        },
+    )
+
+    # Always create icon element for consistent DOM structure (even if empty)
+    # This allows dynamic icon updates via update_toolbar_input_select()
```
gadenbuie, line 544: This is a good comment! Same for the next two comments, which explain more of the "why" than repeating the "what"

shiny/ui/_toolbar.py
```diff
@@ -0,0 +1,897 @@
...
+        select_tag,
+        {
+            "id": resolved_id,
+            "class": "bslib-toolbar-input-select shiny-input-container",
+        },
+        kwargs,
+    )
+
+
+@add_example()
```
gadenbuie, line 599: one last one

docs/_quartodoc-testing.yml
elnelson575: I didn't initially create one-- it would have basically just:
```
expect_align()
expect_width()
expect_gap()
```
Yes? I could see an argument for it since it'd be relatively lightweight, but there isn't a *lot* to interact with/test.
Align is covered by bounding box tests, width less so, and gap less so. Those are, I believe, covered in the pytest tests where it tests for attributes in the html though.

shiny/ui/_toolbar.py
elnelson575: This thwarted me for a bit. Basically, because that HTMLtools function uses `css()` which converts underscores to hyphens in key names, such that`--_divider-gap` and `--_divider-width` become `---divider-gap`, which isn't valid and does not match the bslib class.

shiny/ui/_toolbar.py
elnelson575: Did a quick pass, thanks or flagging! There were a few I kept because it wasn't immediately evident to me what was going on with the MISSING_TYPE situation and it would have taken me an extra moment if I hadn't been familiar with the bslib implementation. I did pull the other similar one up so we're not reiterating it and cleaned up the rest of the file though.

shiny/ui/_toolbar.py
elnelson575: This is the behavior in the existing `_input_action_button()` and `_input_task_button()` implementations-- I think to make the HTML easier to read during debugging?

shiny/ui/_input_code_editor_bundle.py
elnelson575: Done! https://github.com/posit-dev/py-shiny/issues/2160

shiny/ui/_toolbar.py
elnelson575: So they render in sequence-- It doesn't look bad, but thinking on it, the render time might be gross, so maybe better just to make the kitchen sink one a little more comprehensive and only leave that. ✅

shiny/ui/_toolbar.py
elnelson575: Can do!

docs/_quartodoc-core.yml
elnelson575: I don't have strong feelings on this beyond that we should be consistent.

shiny/ui/_toolbar.py
```diff
@@ -0,0 +1,676 @@
...
+
+
+def _extract_text(x: TagChild) -> str:
+    """Extract text content from a TagChild for validation."""
+    if isinstance(x, str):
+        return x
+    if isinstance(x, (list, tuple)):
+        return " ".join(_extract_text(item) for item in x)
+    if isinstance(x, Tag):
+        if hasattr(x, "children"):
```
cpsievert, line 358: I'm a bit surprised that this line is needed?

elnelson575: Yeah, it's a bit overly cautious, just defends against a malformed tag for the most part. Since this is used to grab readable text for labels and warn if there is none, I tended towards caution since an unlabelled icon is essentially broken accessibility-wise.

PR #2165, opened 2026-02-11
feat: Add ui.icon() for inline SVG icons (FontAwesome + Bootstrap)

### LINKED ISSUE #2159: [Feature]: Icon function
### Category

UI Components

### Scope

Major Feature

### Problem

Claude keeps hallucinating a ui.icon() function mimicking bslib's. Right now we import icons from faicons and it feels a bit clunky to have to import a separate package to use icons.
We should consider making ui.icon() a reality.

### Solution

Expect it would look something like:
```python
ui.input_action_button("id", "my label", *, icon=ui.icon("star"))
```

We should make sure the icons lack default margins, so the alignment is handled by the surrounding component.

### Alternatives (Optional)

_No response_

### Example (Optional)

```python

```

### Impact (Optional)

_No response_

### Contribution? (Optional)

None

elnelson575:
Closes #2159

## Summary

- Adds `ui.icon()` for rendering FontAwesome and Bootstrap Icons as inline SVGs
- Both icon sets are **bundled directly** — no external packages required (eliminates `faicons` dependency)
- Bootstrap Icons are bundled to match parity with R's bslib, which includes Bootstrap Icons out of the box
- API: `icon(name, *, lib="fa", size=None, fill=None, class_=None, id=None, variant=None, **kwargs)`
  - `lib`: `"fa"` (default) or `"bs"`
  - `variant`: selects FontAwesome style (`"solid"`, `"regular"`, `"brands"`, etc.)
  - `size`: semantic sizes (`"xs"`, `"sm"`, `"lg"`, `"xl"`, `"2xl"`, etc.) or any CSS unit (`"2em"`, `"24px"`)
  - SVG presentation attributes (`stroke`, `fill_opacity`, etc.) pass through via `**kwargs`
- Full accessibility support: decorative (default, `aria-hidden`) and semantic (`aria-label`) modes
- Drop-in replacement for `faicons.icon_svg()` — same layout defaults (`margin_left`, `margin_right`, `position`)
- Icons are lazy-loaded from JSON on first use and cached for the process lifetime

## Test plan

- [ ] `tests/pytest/test_icon.py` covers BS and FA icons, all params, a11y modes, unknown icon errors, semantic sizes, stroke/fill attrs
- [ ] Example apps: `shiny/api-examples/icon/app-core.py` and `app-express.py`
- [ ] Verify `ui.icon("star")` works in both core and express apps
- [ ] Verify `ui.icon("github", variant="brands")` renders correctly
- [ ] Verify `ui.icon("star", lib="bs", variant="solid")` emits a warning

🤖 Generated with [Claude Code](https://claude.com/claude-code)


shiny/ui/_icon.py
```diff
@@ -0,0 +1,308 @@
...
+    position: Optional[str],
+    title: Optional[str],
+    a11y: str,
+    id: Optional[str],
+) -> Tag:
+    """Create a FontAwesome icon using the faicons package."""
+    try:
+        from faicons import icon_svg
+    except ImportError:
+        raise ImportError(
```
elnelson575: Consider if we want to bundle this in and deprecate the package vs. add it as a hard requirement

elnelson575: If we're worried about boosting the wheel size, we could measure with/without bundling the faicons in there

elnelson575: Via Claude:
  - Shiny wheel today (published 1.5.0): 3,636 KB
  - Shiny with full fa + bs bundled: 4,439 KB → +803 KB (+22%)
  vs.
  - Dash: 7,063 KB — 59% larger than our proposed size

PR #2176, opened 2026-02-18
feat: Save shinylive assets as GHA artifact

elnelson575:
This updates the build process to run the workflow on manual trigger OR commit to main to save the shinylive assets saved with the branch (or main)'s version of pyshiny to an artifact.
This can then be pulled into the py-shiny-site process to allow for local previewing before the PyPI release.

Workflow will work like:

Trigger | Quartodoc | Build Shinylive | Save Artifact | Install Assets | Build Site | Deploy
-- | -- | -- | -- | -- | -- | --
PR | ✅ | ❌ | ❌ | ❌ | ❌ | ❌
Push to docs-** | ✅ | ✅ | ❌ | ✅ | ✅ | ❌
Push to main | ✅ | ✅ | ✅ | ✅ | ✅ | ✅
Manual (build_only = false) | ✅ | ✅ | ✅ | ✅ | ✅ | ❌*
Manual (build_only = true) | ❌ | ✅ | ✅ | ❌ | ❌ | ❌

See https://github.com/posit-dev/py-shiny-site/pull/344 for details

### DIFF
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -1,7 +1,7 @@
 name: Build API docs and Shinylive for GitHub Pages
 
-# Allow for `main` branch to build full website and deploy (with shinylive, saved as artifact)
-# Allow branches that start with `docs-` to build full website, but not deploy (with shinylive, not saved as artifact)
+# Allow for `main` branch to build full website and deploy (saving dev shinylive assets as an artifact)
+# Allow branches that start with `docs-` to build full website, but not deploy (builds dev shinylive assets, no artifact)
 # Allow for PRs to build quartodoc only. (No shinylive, no site build, no deploy)
 # Manual workflow_dispatch can build only shinylive (skip docs) or full build
 


.github/workflows/build-docs.yaml
```diff
@@ -3,9 +3,16 @@ name: Build API docs and Shinylive for GitHub Pages
 # Allow for `main` branch to build full website and deploy
 # Allow branches that start with `docs-` to build full website, but not deploy
 # Allow for PRs to build quartodoc only. (No shinylive, no site build, no deploy)
+# Manual workflow_dispatch allows building shinylive from any branch
 
 on:
   workflow_dispatch:
+    inputs:
+      build_shinylive:
+        description: 'Build shinylive assets (slow, ~10 min)'
```
gadenbuie: Does enabling this *add* 10 minutes? I thought we were basically already doing this?

elnelson575: Nope, so this means that we only build and save the shinylive assets when this job is manually triggered. 
Before, we were just rebuilding on merge/push to main (github.event_name != 'pull_request') in the original workflow. Rather than enabling it on all branches on all pushes, it seemed more space efficient to just build in situations where you want the built artifact to test (otherwise we're building and storing a lot of artifacts we don't need)

.github/workflows/build-docs.yaml
```diff
@@ -1,11 +1,18 @@
 name: Build API docs and Shinylive for GitHub Pages
 
-# Allow for `main` branch to build full website and deploy
-# Allow branches that start with `docs-` to build full website, but not deploy
+# Allow for `main` branch to build full website and deploy (with shinylive, saved as artifact)
+# Allow branches that start with `docs-` to build full website, but not deploy (with shinylive, not saved as artifact)
```
gadenbuie:
```suggestion
# Allow for `main` branch to build full website and deploy (saving dev shinylive assets as an artifact)
# Allow branches that start with `docs-` to build full website, but not deploy (builds dev shinylive assets, no artifact)
```

PR #2182, opened 2026-02-24
feat: Implement R Shiny-style concurrency model

jcheng5:
## Summary

- **Remove global reactive lock**: Replace with per-session `_reactive_lock` so sessions run independently
- **Concurrent effect execution**: Effects within a session now run via `asyncio.gather()` instead of sequentially
- **Cycle start action queue**: Defer input processing while effects are busy, keeping input values stable (R Shiny-style `cycleStartAction` / `busyCount` pattern)
- **Deprecate `reactive.lock()`**: Returns a no-op lock with deprecation warning for backward compatibility
- **Remove `_request_flush()`**: Was already a no-op; py-shiny's per-session message loops don't need a global flush registry

Detailed implementation notes and comparison against R Shiny source are in `plans/CONCURRENCY_IMPLEMENTATION_NOTES.md`.

## Test plan

- [x] All 392 unit tests pass
- [x] Pyright type checking passes with 0 errors
- [x] Formatting passes
- [ ] Playwright end-to-end tests
- [ ] Manual testing with example apps (including `examples/asyncness/`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/asyncness/app.R
+++ b/examples/asyncness/app.R
@@ -5,7 +5,7 @@ library(promises)
 
 ui <- fluidPage(
   p(
-    "Launch this app in two different tabs. In one tab, press the button and see if ",
+    "Launch this Shiny for R app in two different tabs. In one tab, press the button and see if ",
     "the clock stops in the other tab."
   ),
   input_task_button("block", "Perform blocking operation"),
@@ -17,22 +17,24 @@ server <- function(input, output, session) {
     message("[", substr(session$token, 1, 5), "] ", ...)
   }
   output$fast <- renderText({
-    invalidateLater(1000)
+    invalidateLater(100)
     format(Sys.time(), "%H:%M:%S")
   })
 
-  observe({
-    invalidateLater(1000)
-    msg("Fast observer is running")
-  })
+  # observe({
+  #   invalidateLater(1000)
+  #   msg("Fast observer is running")
+  # })
 
   # Observe button clicks and perform slow operation
-  observeEvent(input$block, {
-    msg("Starting slow operation")
-    async_sleep(3) %...>%
-      {
-        msg("Slow operation completed")
-      }
+  lapply(seq_len(3), function(i) {
+    observeEvent(input$block, {
+      msg("Starting slow operation: ", i)
+      async_sleep(3) %...>%
+        {
+          msg("Slow operation completed: ", i)
+        }
+    })
   })
 }
 
--- a/examples/asyncness/app.py
+++ b/examples/asyncness/app.py
@@ -6,7 +6,7 @@ from shiny import App, reactive, render, ui
 
 app_ui = ui.page_fluid(
     ui.p(
-        "Launch this app in two different tabs. In one tab, press the button and see if"
+        "Launch this Shiny for Python app in two different tabs. In one tab, press the button and see if"
         " the clock stops in the other tab."
     ),
     ui.input_task_button("block", "Perform blocking operation"),
@@ -17,15 +17,19 @@ app_ui = ui.page_fluid(
 def server(input, output, session):
     @render.text
     def fast():
-        reactive.invalidate_later(1)
+        reactive.invalidate_later(0.1)
         return time.strftime("%H:%M:%S", time.localtime())
 
-    @reactive.effect
-    @reactive.event(input.block)
-    async def slow():
-        print("Starting slow operation")
-        await asyncio.sleep(3)
-        print("Slow operation finished")
+    def make_effect(i):
+        @reactive.effect
+        @reactive.event(input.block)
+        async def _():
+            print(f"Starting slow operation {i}")
+            await asyncio.sleep(3)
+            print(f"Slow operation {i} finished")
+
+    for i in range(3):
+        make_effect(i + 1)
 
 
 app = App(app_ui, server, static_assets=Path(__file__).parent / "www")


jcheng5:
The apps in `examples/asyncness` and the `plans/` documents don't have to stay, but I wanted them available for the PR review.

I considered other concurrency schemes besides the one we designed for Shiny for R, but in the end, I think those were a really nice set of tradeoffs so this PR ended up in a very similar place.

shiny/reactive/_core.py
```diff
@@ -128,24 +128,8 @@ def __init__(self) -> None:
         )
         self._next_id: int = 0
         self._pending_flush_queue: PriorityQueueFIFO[Context] = PriorityQueueFIFO()
```
schloerke, line 132: This handles priority ordering... but should we also handle priority when resolving the queue's items?

shiny/reactive/_core.py
```diff
@@ -174,9 +158,50 @@ def on_flushed(
...
+                    # Log the exception but don't re-raise; effects handle
+                    # their own errors internally
+                    traceback.print_exception(
+                        type(result), result, result.__traceback__
+                    )
+
+        # If any tasks invalidated new effects during their async portions,
+        # flush those too (recursive).
+        if not self._pending_flush_queue.empty():
+            await self._flush_concurrent()
```
schloerke: If two effects that invalidate each other via two reactive values being set... 

```python
x = reactive.value(0)
y = reactive.value(0)

@reactive.effect
def _A():
    y.set(x() + 1)
@reactive.effect
def _B():
    x.set(y() + 1)
```

... with these lines, wont the flush never resolve, starving everything else?

schloerke: This shinyapp.io app ([[LINK](https://shinylive.io/py/editor/#code=NobwRAdghgtgpmAXGKAHVA6VBPMAaMAYwHsIAXOcpMAMwCdiYACAZwAsBLCbDOAD1R04LFkw4xUxOmTERUAVzJ4mQiABM4dZfI4AdCPUatO3MRKkyhUQmQ4A3OPvGTpTKC2wRCHYvv06MLgUyAH0WABsODToACl1IeOV4gDlEpgAGZQBGdMymACZ0gEo-CAABVWiMEg19DRomMj4yGKLEfSZOlTgyeToIJhp4iAAqfLFRECDFDAhWphGCgF940r4mAF5u61sHDDsocPk4GOL9bE3tm3s4fcPj05KIfQq4HZveGho4G313T0ITHqTBCAEFWu0Bl0oAB3KAcGT-Lw+DAROBwVAxLJPLpMQRcFrxeSoNRQChMbDxHFdDgNaZkWbzAB8BXSkNxnR4LB6MT48wA1ExsS8rNc9nAvj8yH8PF4gRKQQAhCEdaFwhFuWXeYio8LozFZDAAVmpnXx5DiYGJpPJfCpqs6fFRPOwAqFTz8YAAGu0wKVXupNNViLVAw0+CFbGQ9SqoZ0hL1+qwyLE+UUPc8wABNX3+ypBmqOMMUyMImNtB3dRMDFgpmKu9NgJYAXSAA)]) shows that it currently starves the app. 

So ... we don't need to address this?

Or we allow for a second pass on the queue and let the flush finish. The remaining effects will be calculated on a followup flush.

shiny/session/_session.py
```diff
@@ -672,37 +684,51 @@ def verify_state(expected_state: ConnectionState) -> None:
...
 
+                            # Progress messages may have queued up; let them drain
+                            await asyncio.sleep(0)
+
+                            await reactive_flush()
+
                             # TODO: Remove this call to reactive_flush() once https://github.com/posit-dev/py-shiny/issues/1889 is fixed
                             # Workaround: Any `on_flushed()` calls from bookmark's `on_restored()` will be flushed here
                             if self.bookmark.store != "disable":
                                 await reactive_flush()
```
schloerke: Can this be removed? You're doing the reactive flush every time

shiny/session/_session.py
```diff
@@ -672,37 +684,51 @@ def verify_state(expected_state: ConnectionState) -> None:
                             with session_context(self):
                                 self.app.server(self.input, self.output, self)
 
+                            # Progress messages may have queued up; let them drain
+                            await asyncio.sleep(0)
+
+                            await reactive_flush()
```
schloerke, line 724: I feel like this cadence needs to be a function. The purpose could be documented and be consistently re-used

shiny/reactive/_core.py
schloerke: Can this instead use `_start_cycle` loop? That may delay followup calls a few ticks

shiny/session/_session.py
```diff
@@ -1076,14 +1098,77 @@ async def _flush(self) -> None:
...
+
+        When count reaches 0, sends 'idle' message and starts processing any queued
+        cycle start actions (deferred input updates, etc.).
+        """
         self._busy_count -= 1
         if self._busy_count == 0:
             self._send_message_sync({"busy": "idle"})
+            # Start processing queued actions now that no effects are busy
+            if self._cycle_start_action_queue:
+                asyncio.create_task(self._start_cycle())
```
schloerke, line 1210: This cadence is important. I'd like a helper method

shiny/reactive/_core.py
jcheng5: This is intentional--it's to avoid only-temporarily-correct values from being shown in the UI. It's definitely debatable whether this is always the right behavior, I think it goes hand in hand with wanting to have each output flushed to the client as soon as it's ready (instead of all the outputs being held until they're all ready).

jcheng5: See comment above.

shiny/reactive/_core.py
```diff
@@ -174,9 +158,50 @@ def on_flushed(
...
+
+        Relies on CPython's event loop processing the ready queue in FIFO order:
+        ``create_task`` adds the coroutine's first step, then ``sleep(0)`` adds
+        our resume callback after it, so the task's sync portion always runs
+        before we dequeue the next context.
+        """
+        tasks: list[asyncio.Task[None]] = []
+        while not self._pending_flush_queue.empty():
+            ctx = self._pending_flush_queue.get()
+            task = asyncio.create_task(ctx.execute_flush_callbacks())
```
jcheng5: According to [this](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task), `create_task` has an `eager_start` parameter that would let us skip the `asyncio.sleep(0)`.

jcheng5: Apparently eager_start is only since 3.14.

jcheng5: We could have a helper function that, today, calls `create_task` + sleep, but can change to `eager_start` when 3.14 becomes our minimum (or dispatch based on current Python version if we care that much).

shiny/reactive/_core.py
```diff
@@ -174,9 +158,50 @@ def on_flushed(
...
+                if isinstance(result, BaseException):
+                    # Log the exception but don't re-raise; effects handle
+                    # their own errors internally
+                    traceback.print_exception(
+                        type(result), result, result.__traceback__
+                    )
+
+        # If any tasks invalidated new effects during their async portions,
+        # flush those too (recursive).
+        if not self._pending_flush_queue.empty():
```
jcheng5: Should this be an outer while loop instead of recursion? To remove any possibility of stack overflow?

shiny/reactive/_core.py
```diff
@@ -291,14 +316,40 @@ def on_flushed(
...
-    For example, :func:`~shiny.reactive.lock` makes it safe to set a
-    :class:`~reactive.value` and call :func:`~shiny.reactive.flush` from a different
-    :class:`~asyncio.Task` than the one that is running the Shiny
-    :class:`~shiny.Session`.
+    .. deprecated::
+        The global reactive lock has been removed. Reactive locking is now handled
+        per-session. This function returns a no-op lock for backward compatibility.
     """
-    return _reactive_environment.lock
+    warnings.warn(
```
jcheng5, line 332: Did a search of GitHub and found these Shiny apps that rely on reactive.lock.

* https://github.com/posit-dev/shiny-assistant/blob/cae561f9f94cfda0988f8da3a83411be0962b607/shinyapp/app.py#L311
* https://github.com/jcheng5/py-sidebot/blob/55d13900b8a7b50c327609830df4877080fad19d/app.py#L311
* https://github.com/parlance-labs/langfree/blob/6aee32f08b8b71b2af077249c1e9dc47f35e0b6d/langfree/shiny.py#L93
* https://github.com/wch/chatstream/blob/ef48867ec15ff5946552d1c887cc42763fced876/chatstream/__init__.py#L458
* https://github.com/amcrabtree/cell-sidebot/blob/3c58c7aa931e5f6da24008448c3deca543762720/app.py#L239
* https://github.com/viksrc/tickplot/blob/917427ae2c73e1fe903fb68cbfec545026440592/app.py#L277
* https://github.com/matheus-rech/shiny-test/blob/6292b9f5885a4a2eb559fabec2a3bda2dda08fa8/app.py#L260
* https://github.com/zyf0717/pi-pulse/blob/4a04697abc50a08dbcaef2f42ada9b09e5752568/server.py#L61

shiny/reactive/_core.py
```diff
@@ -359,12 +410,32 @@ async def _task(ctx: Context, deadline: float) -> None:
...
+                    if isinstance(root, AppSession):
+                        root_session = root
+                except (AttributeError, TypeError):
+                    pass
+
+            if root_session is not None:
+
+                async def _do_invalidate() -> None:
+                    nonlocal cancellable
+                    async with root_session._reactive_lock:
```
jcheng5, line 439: What if we add `Session.reactive_lock()` so callers don't have to do the walking up the scopes themselves? (The express stub session should no-op or throw.)

shiny/reactive/_core.py
```diff
@@ -359,12 +410,32 @@ async def _task(ctx: Context, deadline: float) -> None:
...
+                        cancellable = False
+                        ctx.invalidate()
+                        await flush()
+
+                await root_session._cycle_start_action(_do_invalidate)
```
jcheng5, line 444: I don't like that this uses a private. I think the philosophy should be that 3rd parties could implement something just like `invalidate_later` themselves, without accessing any private members.

shiny/session/_session.py
```diff
@@ -1076,14 +1098,77 @@ async def _flush(self) -> None:
...
+
+    async def _cycle_start_action(
```
jcheng5, line 1212: This should be renamed to include a more active verb; as is, it sounds like "start" is the verb and it's not.

schloerke:
> ### Test plan
> * Playwright end-to-end tests

Currently, https://github.com/posit-dev/py-shiny/blob/299bdbeeadc121f4d6cc15e2eff0f22dccf48adf/tests/playwright/shiny/session/flush/test_on_flush.py#L11-L16 is failing

schloerke:
Please have claude get `uv run make check-fix` to run successfully

shiny/reactive/_extended_task.py
```diff
@@ -146,22 +146,21 @@ def _done_callback(self, task: asyncio.Task[R]) -> None:
         self._task = None
 
         async def _impl():
-            async with lock():
```
jcheng5, line 196: Make sure it's OK that we removed this. Ultrathink.

shiny/session/_session.py
```diff
@@ -572,6 +571,14 @@ def __init__(
         self._flush_callbacks = _utils.AsyncCallbacks()
         self._flushed_callbacks = _utils.AsyncCallbacks()
 
+        # Per-session reactive lock (replaces the old global lock)
+        self._reactive_lock: asyncio.Lock = asyncio.Lock()
+
+        # Cycle start action queue: holds callbacks to run when no effects are busy.
```
jcheng5:
```suggestion
        # Cycle start action queue: Stores callbacks while effects are busy; they
        # can run when everything in this session is idle.
```

shiny/session/_session.py
```diff
@@ -584,6 +591,10 @@ async def _run_session_ended_tasks(self) -> None:
             return
         self._has_run_session_ended_tasks = True
 
+        # Clear any queued cycle start actions and reset busy count
+        self._cycle_start_action_queue.clear()
```
jcheng5, line 609: Is this the only cleanup we need to do? Make sure we don't leak.

shiny/session/_session.py
```diff
@@ -584,6 +591,10 @@ async def _run_session_ended_tasks(self) -> None:
...
+        self._busy_count = 0
```
jcheng5, line 610: Should this be set to 0? Why? Calls to increment/decrement might still be outstanding?

jcheng5: It does affect whether `_decrement_busy_count()` calls `_start_cycle`, so proceed carefully.

jcheng5: Better yet, `_start_cycle` could check if we're ended (just in case).

shiny/reactive/_extended_task.py
```diff
@@ -146,22 +146,21 @@ def _done_callback(self, task: asyncio.Task[R]) -> None:
...
-                if len(self._invocation_queue) > 0:
-                    next_invocation = self._invocation_queue.pop(0)
-                    next_invocation()
-                    await flush()
+            if len(self._invocation_queue) > 0:
+                next_invocation = self._invocation_queue.pop(0)
+                next_invocation()
+                await flush()
 
         asyncio.create_task(_impl())
```
jcheng5, line 212: I think this might need to be

```
        asyncio.create_task(session._cycle_start_action(_impl()))
```

I think that's the equivalent to `async with reactive.lock()` now??

shiny/session/_session.py
```diff
@@ -572,6 +571,14 @@ def __init__(
         self._flush_callbacks = _utils.AsyncCallbacks()
         self._flushed_callbacks = _utils.AsyncCallbacks()
 
+        # Per-session reactive lock (replaces the old global lock)
+        self._reactive_lock: asyncio.Lock = asyncio.Lock()
```
jcheng5, line 588: Do we need both this AND discipline around always using start_cycle_action? Don't they both achieve the same effect--of serializing unrelated "callbacks" that are for the same session?

shiny/reactive/_extended_task.py
jcheng5: Or `session._reactive_lock()`, feels like one or the other.

jcheng5: Maybe... `reactive.lock()` should be deprecated but ALSO call `get_current_session()._reactive_lock()`???

shiny/session/_session.py
```diff
@@ -636,8 +647,9 @@ def verify_state(expected_state: ConnectionState) -> None:
                         )
                         return
 
-                    async with lock():
-                        if message_obj["method"] == "init":
+                    if message_obj["method"] == "init":
+                        # Init must run immediately (session setup)
+                        async with self._reactive_lock:
```
jcheng5, line 677: The `update` method does BOTH self._reactive_lock and _cycle_start_action; should we change this to be consistent?? Or do we just know it doesn't matter because the session is just starting??

shiny/session/_session.py
```diff
@@ -672,37 +684,51 @@ def verify_state(expected_state: ConnectionState) -> None:
...
+
+                    elif "tag" in message_obj and "args" in message_obj:
+                        # Dispatch messages (e.g. uploadInit, uploadEnd,
+                        # makeRequest) run immediately, NOT deferred through the
+                        # cycle start action queue. This matches R Shiny, where
+                        # only input updates are deferred. Dispatches need prompt
+                        # handling for things like ExtendedTask invocations and
+                        # data frame cell edits via makeRequest.
+                        message_other = typing.cast(ClientMessageOther, message_obj)
+                        async with self._reactive_lock:
```
jcheng5, line 754: Are we sure this needs a reactive lock (and/or cycle_start_action)?

shiny/session/_session.py
```diff
@@ -672,37 +684,51 @@ def verify_state(expected_state: ConnectionState) -> None:
...
-                        # https://github.com/posit-dev/py-shiny/issues/1381
-                        await asyncio.sleep(0)
-
-                        self._request_flush()
-
-                        await reactive_flush()
+                    else:
+                        raise ProtocolError(
+                            f"Unrecognized method {message_obj['method']}"
+                        )
```
jcheng5, line 738: Question unrelated to this line of code: do download handlers also need to be locked?

shiny/session/_session.py
```diff
@@ -1047,9 +1072,6 @@ def on_flushed(
     ) -> Callable[[], None]:
         return self._flushed_callbacks.register(wrap_async(fn), once)
 
-    def _request_flush(self) -> None:
-        self.app._request_flush(self)
```
jcheng5, line 1105: It's fine that we're removing this BUT we REALLY ought to be having Session._flush exit early if it finds nothing. Currently it's sending no-op packets to the client, that's wasteful already but in a concurrent session world it is fully broken as it will can outputs to un-grey too early.

shiny/reactive/_core.py
jcheng5, line 444: IMPORTANT: As written, invalidate_later has no higher priority than inputs. Claude claims that in Shiny for R, invalidate_later is not blocked by the start cycle, so effectively it's higher priority than inputs.

jcheng5:
<details>
<summary>Conversation with Claude about possible redundancy between session._reactive_lock and cycle_start_action</summary>

**❯ Is it possible that the session._reactive_lock introduced by this PR, is entirely redundant with the _busy_count/_cycle_start_action_queue (or vice
  versa)? Is it possible that both play exactly one role, which is to serialize actions against a single session?**

⏺ Let me examine both mechanisms to understand their roles.

⏺ Explore(Explore reactive lock and busy count)
  ⎿  Done (25 tool uses · 56.0k tokens · 1m 28s)
  (ctrl+o to expand)

⏺ Great question. Let me verify the exact timing relationship between the lock and busy_count by reading the critical code paths.

⏺ Read 2 files (ctrl+o to expand)

⏺ Now let me check whether reactive_flush() runs all effects to completion inline (which would mean busy_count always returns to 0 before the lock is
  released).

⏺ Searched for 1 pattern, read 1 file (ctrl+o to expand)

⏺ Let me also check invalidate_later to see how it interacts.

⏺ Searched for 1 pattern, read 1 file (ctrl+o to expand)

⏺ Now I have a complete picture. Here's my analysis:

  Short answer: They overlap heavily, but neither is fully redundant with the other.

  The overlap

  For the common case — serializing input-update-triggered reactive flushes — they do the same thing. Here's why:

  reactive_flush() (specifically _flush_concurrent) runs all effects to completion (including cascading invalidations) via asyncio.gather() + recursive
  re-flush before returning. Since the lock is held for the entire manage_inputs → flush cycle, busy_count is always back to 0 by the time the lock is
  released. So the queue's "wait until effects settle" behavior is identical to the lock's "wait until the lock is free" behavior.

  If your only concern were serializing input updates against each other, the lock alone would be sufficient.

  What the lock adds that the queue doesn't

  Dispatch messages (uploadEnd, makeRequest, etc.) bypass the queue entirely and go straight through the lock (line 722). Without the lock, a dispatch message
   could interleave with a queued input update at any await point during reactive_flush().

  What the queue adds that the lock doesn't

  1. Dispatch priority over input updates. Dispatches acquire the lock directly, so they can jump ahead of queued input updates. Without the queue
  (lock-only), everything would be FIFO in the lock's waiter list. This is the R Shiny behavior: dispatches (like ExtendedTask invocations) run promptly,
  while input updates wait for the current cycle to finish.
  2. Non-blocking websocket loop. Input updates return immediately after appending to the queue (_cycle_start_action doesn't await the lock). Without the
  queue, the websocket loop would block on the lock for each input update, preventing it from receiving further messages until the current flush completes.
  3. Deferred inputs during HTTP requests. _handle_request (line 872) increments busy_count without holding the lock. So during a file upload, the queue
  defers input updates even though the lock is free. Without the queue, input updates would sail through the lock unimpeded during an upload.
  4. Client busy/idle signaling. The busy_count drives the {"busy": "busy"}/{"busy": "idle"} messages to the client. This is orthogonal to serialization but
  coupled to the mechanism.

  Could you collapse them into one?

  Theoretically yes, but each direction loses something:

  - Queue only (drop the lock): You'd need to route dispatches through the queue too, losing their prompt-execution semantics. Or you'd need a different
  serialization mechanism for dispatches.
  - Lock only (drop the queue): You'd lose dispatch priority, websocket loop responsiveness, the HTTP-request deferral, and client busy/idle signaling. The
  first two are the most material.

  Bottom line

  They both serialize reactive operations against a session, and for the happy path they're doing the same work. But the queue provides scheduling policy
  (what order things run, what gets deferred) while the lock provides mutual exclusion (nothing runs concurrently). Those are complementary concerns that
  happen to produce the same outcome in the simple case.

✻ Brewed for 5m 12s

**❯ Is it possible some of these different behaviors are bugs, not intentional?**

⏺ Yes, absolutely. Let me reconsider each "difference" with fresh eyes:

  Dispatch blocking the websocket loop — Almost certainly a bug. Dispatches acquire the lock inline (line 722), which means conn.recv() can't be called again
  until the lock is available and the dispatch+flush completes. If a long-running flush is in progress, the websocket loop is frozen. In R Shiny this isn't an
   issue because the message-receive path is never blocked—R's single-threaded event loop processes messages that are already received and queued.

  Dispatch "priority" over input updates — This isn't a designed priority system, it's just a side effect of dispatches waiting on the lock (in asyncio's FIFO
   waiter list) while input updates wait in the separate action queue. The ordering falls out accidentally from using two different queueing mechanisms. In R
  Shiny, dispatches run "immediately" only in the sense that they're not routed through startCycle—but there's no preemption story because everything is
  naturally serialized.

  HTTP requests deferring input updates — Almost certainly an unintended interaction. The _increment_busy_count in _handle_request (line 872) predates this PR
   and existed purely for client busy/idle signaling. Now that busy_count also gates the action queue, a file upload inadvertently blocks all input
  processing. In R Shiny, file uploads don't defer input updates this way.

  invalidate_later going through the queue — Worth scrutinizing. In R Shiny, invalidateLater schedules a later() callback that fires at the next event-loop
  iteration. It does not go through startCycle. In the Python code (line 435), it goes through _cycle_start_action, which means it queues behind any pending
  input updates. If three input updates are queued and then an invalidate_later timer fires, the invalidation waits for all three to process first. That's
  different from R, where the later() callback would fire at the top of the event loop.

  The overall picture: I think the honest answer to your original question is that the two mechanisms are largely doing the same job, and the behavioral
  differences between them aren't intentional design choices but emergent consequences of having two overlapping serialization systems. The lock was likely
  added for safety ("we're in async Python, we need mutual exclusion") and the queue was added for R Shiny semantics ("we need startCycle/busy_count"),
  without fully reckoning with the fact that they interact in subtle, untested ways.

  If you wanted to clean this up, I think the question to answer is: which mechanism should be the primary serialization strategy, with the other either
  removed or demoted to a narrow supporting role?

✻ Cooked for 1m 17s

**❯ Your claim about invalidateLater is incorrect--it does use cycleStartAction**

⏺ You're right, I apologize for the incorrect claim. If R Shiny's invalidateLater also routes through cycleStartAction, then the Python implementation on line
   435 is faithful to R's behavior and that one is correct.

  That narrows the list of likely-accidental behaviors to:

  1. Dispatch blocking the websocket receive loop — Still suspect. The async with self._reactive_lock on line 722 freezes the message loop until the lock is
  free and the dispatch+flush completes.
  2. HTTP requests deferring input updates — Still suspect. _handle_request bumping busy_count now has a side effect it didn't used to have.
  3. Dispatch "priority" — Still just a side effect of two different queueing mechanisms, not a designed ordering guarantee.

</details>

shiny/session/_session.py
jcheng5: Barret says: yes

jcheng5: It ended up being a bit more complicated than this; it's now locked for the handler's synchronous execution, but if it returns a generator, the lock is released between iterations. We'll need to document this.

shiny/session/_session.py
```diff
@@ -672,37 +683,51 @@ def verify_state(expected_state: ConnectionState) -> None:
...
-                        # https://github.com/posit-dev/py-shiny/issues/1381
-                        await asyncio.sleep(0)
-
-                        self._request_flush()
-
-                        await reactive_flush()
+                    else:
+                        raise ProtocolError(
+                            f"Unrecognized method {message_obj['method']}"
+                        )
```
jcheng5, line 738: Also related to download handlers: the browser doesn't prompt for a download destination until the iterator is finished, I found that surprising.

shiny/session/_session.py
```diff
@@ -944,6 +944,22 @@ async def _handle_request_impl(
...
+                        # of the content generator. This means the lock is
+                        # acquired-released-acquired-released for each chunk yielded.
+                        #
+                        # TODO: Consider whether we need to hold the reactive lock for
+                        # the entire streaming iteration, or only for the first chunk.
+                        # The common pattern is: do some reactive/async computation
+                        # before the first yield, then stream pure I/O after that. If
+                        # so, we could acquire the lock once, yield the first chunk,
+                        # then release it and stream the rest without the lock. This
+                        # would reduce contention for long-running downloads.
```
schloerke, line 1010: Needs to be addressed

shiny/session/_session.py
```diff
@@ -963,34 +991,62 @@ async def _handle_request_impl(
...
-                                with session_context(self):
-                                    with isolate():
-                                        async for chunk in contents:
-                                            if isinstance(chunk, str):
-                                                yield chunk.encode(download.encoding)
-                                            else:
-                                                yield chunk
-
+                                async with self._reactive_lock:
+                                    with session_context(self), isolate():
```
schloerke:
```suggestion
                                    with session_context(self):
```

shiny/reactive/_core.py
schloerke: When items are added to the queue, priority is handled then. 

Resolving

shiny/reactive/_core.py
schloerke: There isn't a common enough use case throughout the repo. Leaving it as is for now

shiny/reactive/_core.py
schloerke: 073b9e54

shiny/session/_session.py
schloerke: `sleep` then `await flush()` does not occur within the code except for here.

Given it is only here within the WS traffic, I'm going to resolve this thread

shiny/session/_session.py
schloerke: Fixed by not sending a message if it is empty

shiny/reactive/_core.py
```diff
@@ -299,14 +324,40 @@ def on_flushed(
...
-    :class:`~reactive.value` and call :func:`~shiny.reactive.flush` from a different
-    :class:`~asyncio.Task` than the one that is running the Shiny
-    :class:`~shiny.Session`.
+    .. deprecated::
+        The global reactive lock has been removed. Reactive locking is now handled
+        per-session. This function returns a no-op lock for backward compatibility.
     """
-    return _reactive_environment.lock
+    warnings.warn(
+        "reactive.lock() is deprecated. The reactive lock is now per-session. "
```
schloerke, line 333: From IRL convo, maybe this should also perform a real session lock as that is what the user is expecting.

shiny/session/_session.py
```diff
@@ -1115,29 +1169,93 @@ async def _flush(self) -> None:
         try:
             omq = self._outbound_message_queues
 
-            message: dict[str, object] = {
-                "values": omq.values,
-                "inputMessages": omq.input_messages,
-                "errors": omq.errors,
-            }
+            if not omq.is_empty():
```
jcheng5, line 1172: This is NOT correct, IIRC sometimes empty messages need to be sent to clear out "recalculating" state. (I think req with cancel_output is one?)

jcheng5: We actually already know which outputs need to clear out their recalculating state; they're tracked using the `{"recalculating":{"name":"foo","status":"recalculated"}}` message, but the CSS changes are not actually carried out until the next `{"errors":{},"values":{},"inputMessages":[]}` style message is received. We have `{"busy":"idle"}` which could be a more reliable trigger. It's currently being sent before the e/v/iM, so maybe it would be weird for the CSS to clear before the outputs are rendered; we could swap the order perhaps.

PR #2246, opened 2026-05-19
chore: Upgrade bslib to 0.11.0 and Bootstrap to 5.3.8, add brite theme

elnelson575:
## Summary

**Depends on:**
- rstudio/bslib#1308 — radio/checkbox specificity fix (must be merged into bslib before vendoring is final)
- #2256 — reorder bundle so bslib wins cascade ties against shiny.scss

Vendors updated assets from `rstudio/bslib@main` via `make upgrade-html-deps`:

- **bslib** `0.10.0.9000` → `0.11.0.9000`
- **Bootstrap** `5.3.1` → `5.3.8`
- **prism-code-editor** `0.10.0.9000` → `0.11.0.9000`
- **New Bootswatch theme: `brite`** — added upstream in Bootswatch, now vendored and exposed as a valid `ShinyThemePreset`

This was produced as a side-effect of running `make upgrade-html-deps` to verify the fix in #2242. Split into a separate PR because the Bootstrap minor version bump warrants independent review.

## Risk

The Bootstrap `5.3.1` → `5.3.8` bump may include CSS changes that affect existing app styling. Review the vendored SCSS/CSS diffs before merging.

## Test plan

- [x] Visually verify representative Shiny apps across a few themes
- [x] Run Playwright tests
- [x] Confirm `brite` theme renders correctly
- [x] Manually verify regression fix (IntersectionObserver)

<details>
<summary>Breaking changes evaluation</summary>

**1. shiny.js — IntersectionObserver fix (rstudio/shiny#4388)**

- Adds `resolveObservableTarget()` to walk up to the nearest non-`.shiny-html-output` ancestor before attaching the `IntersectionObserver`
- **Risk: none.** Pure bug fix for `uiOutput` visibility detection. No API surface changes.

**2. CSS — checkbox/radio label margin (rstudio/bslib#1308 + #2256)**

- SCSS source: `.shiny-input-checkboxgroup` → `.shiny-input-checkboxgroup.shiny-input-checkboxgroup` (specificity bump)
- Cascade order now correct: shiny.scss inline rule comes before bslib's doubled rule, so bslib wins for all group variants
- **Visible change:** The `margin-top` on `label ~ .shiny-options-group` shifts from `−10px` (shiny.scss) to `calc(−0.15em − var(--bs-border-width))` (~−3.4px) for both stacked and inline groups
- **Risk: low.** This is a visual bug fix — the gap between the group label and its options was too large before. Any user CSS overriding this `margin-top` with a single-class selector would now need two classes to win, but that's an edge case for a spacing property.

**3. Everything else** — version string bumps in JSON files and JS headers only; no behavioral changes.

**Summary:** No breaking changes. The one user-visible effect is the corrected label-to-options gap on checkbox and radio groups in BS5 apps, which is intentional.

</details>


shiny/www/shared/_version.json
```diff
@@ -1,5 +1,5 @@
 {
   "note!": "Generated by scripts/htmlDependencies.R: do not edit by hand",
   "package": "shiny",
-  "version": "1.13.0.9000 (rstudio/shiny@13c073a14b3ce585faedfa648b7111312b25ec17)"
+  "version": "1.13.0.9000 (rstudio/shiny@22de425e850030741faf46c2cbc60c1dca4e75d9)"
```
cpsievert: Let's make sure to include any additional Python changes we need for this, specifically https://github.com/rstudio/shiny/pull/4371

elnelson575: So the changes brought in by re-vendoring bslib and shiny are nonbreaking, so my plan was actually to do this one first, then bring in my in-progress py-shiny branch which adds things like the `enabled` param. Right now, even with the changes on this branch, the py-shiny behavior should be unchanged.

I am holding off until after a few more things merge into bslib though for the re-vendor-- ex. the radio buttons spacing fix is almost done and I'd like that to go in if we're re-vendoring assets anyway.

cpsievert: I just approved that spacing PR, feel free to merge if you're able to. Would also be great if we can get in this fairly serious regression that was just fixed upstream in https://github.com/rstudio/shiny/pull/4388 asap

elnelson575: Perfect, I'll finish this off now 👍🏻

cpsievert (approved):
Thanks!

PR #2261, opened 2026-05-22
fix(test): Combine scroll checks into a single stabilized bottom script

karangattu:
Merge the previous two JS helpers into a single SETTLED_AT_BOTTOM_SCRIPT used by the test to avoid a race between settle-then-check. The new script: simplifies variable names, increases bottom tolerance (20px) to account for subpixel rounding/layout shifts, and requires consecutive stable reads at the bottom across multiple polls before returning true. The test now uses one wait_for_function with 250ms polling (removing the extra short wait), reducing flakiness in auto-scroll assertions.


cpsievert:
@karangattu is this mergable? Should I review it?

karangattu:
@cpsievert yes, I was holding off if the tests were passing without this fix. But feel free to review it, if you think this would be a good direction for the fix.

PR #2317, opened 2026-07-03
fix: Make launch-browser independent of Uvicorn log level

### LINKED ISSUE #569: logging.disable(logging.INFO) conflicts with Shiny's logging and also prevents it from launching browser on start ("--launch-browser")
I am currently using the `logging` package in my code, and Shiny also depends on it for message logging. So how can I disable logging only in my code? `logging.disable(logging.INFO)` not only disables Shiny logs but also prevents it from launching the browser on start ("--launch-browser" doesn't work).

The answer is not to use the root logger, create a new one instead. However, blocking "--launch-browser" while disabling the root logger is weird.

karangattu:
## Summary

Fixes #569 by replacing the `--launch-browser` readiness signal from Uvicorn log interception with a startup callback.

This makes browser launch work when users run with `--log-level=warning` or disable INFO logging.

## Changes

- Add `ShinyServer`, a small `uvicorn.Server` wrapper that runs an `on_started` callback after startup.
- Mirror `uvicorn.run()` locally so Shiny can pass that callback through reload and non-reload modes.
- Replace `LaunchBrowserHandler` with direct browser opening after startup.
- Remove the reload log-handler path and call `_autoreload.reload_end()` from startup instead.
- Add regression tests and a changelog entry.

Manually tested the following cases

`shiny run app.py --launch-browser --log-level=warning` → browser opens
`shiny run app.py --reload --launch-browser` → browser opens once


tests/pytest/test_main.py
```diff
@@ -1,9 +1,101 @@
...
+        captured_kwargs.update(kwargs)
+
+    monkeypatch.setattr(_main, "_run_uvicorn", run_uvicorn)
+
+    def launch_browser(host: str, port: int) -> None:
+        browser_calls.append((host, port))
+
+    monkeypatch.setattr(_main._launchbrowser, "launch_browser", launch_browser)
+
+    _main.run_app(
```
schloerke, line 44: Both new tests call `run_app(...)` with the default `dev_mode=True`, which executes `os.environ["SHINY_DEV_MODE"] = "1"` (`shiny/_main.py`) and never cleans it up — `run_app` normally never returns, but here `_run_uvicorn` is mocked, so the env var leaks into the pytest process. Any later test in the same process that renders UI then gets an extra `shiny-devmode` HTMLDependency (`shiny/html_dependencies.py`). CI passes today, but it's a latent test-ordering landmine.

Please pass `dev_mode=False` here and in `test_reload_uses_startup_callback_not_log_handler` below (or use `monkeypatch.setenv`/`delenv` to isolate the env).

shiny/_main.py
```diff
@@ -58,6 +65,66 @@ def main() -> None:
...
+    config = Config(app, **kwargs)
+    config_attrs = cast(Any, config)
+    server = ShinyServer(config=config, on_started=on_started)
+
+    if (config_attrs.reload or config_attrs.workers > 1) and not isinstance(app, str):
+        logging.getLogger("uvicorn.error").warning(
+            "You must pass the application as an import string to enable "
+            "'reload' or 'workers'."
+        )
+        sys.exit(1)
```
schloerke: Minor: uvicorn's own `run()` exits this error path with `STARTUP_FAILURE` (3) — and `_UVICORN_STARTUP_FAILURE = 3` is already defined above. Using the constant here would keep exit-code fidelity with upstream:

```suggestion
        sys.exit(_UVICORN_STARTUP_FAILURE)
```

shiny/_main.py
```diff
@@ -58,6 +65,66 @@ def main() -> None:
...
+
+    try:
+        if config.should_reload:
+            sock = config.bind_socket()
+            ChangeReload(config, target=server.run, sockets=[sock]).run()
+        elif config_attrs.workers > 1:
+            sock = config.bind_socket()
+            Multiprocess(config, target=server.run, sockets=[sock]).run()
+        else:
+            server.run()
```
schloerke: Minor mirror gap: current uvicorn `run()` calls `config.load_app()` eagerly in the non-reload path so import errors fail fast before binding; omitting it means app-load errors surface later inside `server.serve()`. Same traceback either way, just noisier — worth a line if you want exact parity.

schloerke (commented):
This is a clear improvement over the old log-string sniffing, and I verified the `_run_uvicorn()` mirror against uvicorn 0.50.2's `run()` — the reload/workers/uds/startup-failure logic matches. Pickling in reload mode is safe (`reload_end` is module-level; the `partial(launch_browser, ...)` path only runs in non-reload), the reload branch keeps autoreload websocket notifications working even without `--launch-browser`, and the port-conflict fallback (`reload = False`) correctly falls through to the browser-launch assignment.

A few comments inline — the `dev_mode` test leak is the one I'd like fixed before merge; the rest are take-it-or-leave-it.

One more thing to be aware of (not a defect): this PR stops setting `SHINY_BROWSER_HOST`/`SHINY_BROWSER_PORT` entirely. They're internal-only per a grep of the codebase, but they were visible to running user apps, so any out-of-tree code reading them will break silently.

tests/pytest/test_main.py
schloerke: Fixed in 8e8bc947 — both tests now pass `dev_mode=False` to `run_app()`, so `SHINY_DEV_MODE` is no longer set in the pytest process.

shiny/_main.py
schloerke: Fixed in d8b130a8 — this path now exits with `_UVICORN_STARTUP_FAILURE` (3), matching upstream `uvicorn.run()`.

shiny/_main.py
schloerke: Addressed in d8b130a8 — restructured to mirror uvicorn.run()'s `if reload/workers: ... else: load_app()` shape so import errors fail fast before the socket binds.

One deviation: `Config.load_app()` was only added in recent Uvicorn releases (it's absent through 0.35, present in 0.50), and py-shiny pins `uvicorn>=0.16.0`, so the call is guarded with `hasattr(config, "load_app")`. Older versions keep loading the app inside `serve()` as before.

PR #2335, opened 2026-07-08
ci: Test oldest supported Python and dependency versions

schloerke:
Adds a single GHA job (`test-oldest-deps` in `pytest.yaml`) that runs the unit tests on ubuntu with the oldest supported Python (3.10) and every runtime dependency resolved to its declared lower bound, via `uv pip compile --resolution lowest-direct`. Test tooling resolves normally (the `test` extra has unbounded deps that lowest-resolution would pin to unbuildable 2010-era releases), constrained so it cannot upgrade the pinned runtime deps.

## Lower-bound fixes the new job flagged

- **`starlette` / `prompt-toolkit` were unbounded**, so "lowest" resolved to starlette 0.1.0 (2018) and prompt-toolkit 2.0.1. Now bounded to `starlette>=0.17.1` (the historical pin from before #1123, still passes the full unit suite) and `prompt-toolkit>=3.0.0`.
- **`opentelemetry-api`/`opentelemetry-sdk` `>=1.20.0` was too old.** Two real incompatibilities, found by bisecting versions against the unit suite:
  - `Logger.emit(body=..., severity_text=..., ...)` keyword arguments only exist since **1.38.0**. Below that, shiny's otel log emission raises `TypeError`, which the broad `except Exception` in `emit_otel_log()` silently swallows — value-update logging just never emits.
  - Below sdk **1.24.0**, `Span.record_exception()` formatted the stack trace from the *active* exception instead of the exception object passed in, so `sanitize_errors=True` leaked the unsanitized error message into span events.

  Both floors raised to 1.38.0.
- `tests/pytest/test_otel_value_logging.py` imports `InMemoryLogRecordExporter`, which only exists in opentelemetry-sdk ≥ 1.39.0; it now falls back to the pre-rename `InMemoryLogExporter` (same API) so it can run at sdk 1.38.

## Verification

Reproduced the job's install steps locally in a clean Python 3.10 venv: all **859 unit tests pass** with runtime deps at their minimums (starlette 0.17.1, htmltools 0.7.0, websockets 13.0, uvicorn 0.16.0, narwhals 1.10.0, markdown-it-py 1.1.0, otel 1.38.0, …). The full suite also still passes with latest deps.


pyproject.toml
```diff
@@ -51,7 +51,7 @@ classifiers = [
 dependencies = [
     "typing-extensions>=4.10.0",
     "uvicorn>=0.16.0;platform_system!='Emscripten'",
-    "starlette",
+    "starlette>=0.17.1",
```
schloerke, line 54: **Why this bound was added:** `starlette` had no lower bound, so the oldest-deps job's `--resolution lowest-direct` resolved it to starlette 0.1.0 (2018), which predates the ASGI interface shiny is built on — the job couldn't even construct a working environment. `>=0.17.1` restores the historical floor this project pinned before #1123 un-pinned it, and it is the oldest version I verified: the full unit suite (859 tests) passes against starlette 0.17.1.

pyproject.toml
```diff
@@ -63,18 +63,20 @@ dependencies = [
     "packaging>=20.9",
     "watchfiles>=0.18.0;platform_system!='Emscripten'",
     "questionary>=2.0.0;platform_system!='Emscripten'",
-    "prompt-toolkit;platform_system!='Emscripten'",
+    "prompt-toolkit>=3.0.0;platform_system!='Emscripten'",
```
schloerke, line 66: **Why this bound was added:** `prompt-toolkit` had no lower bound, so lowest resolution picked prompt-toolkit 2.0.1 (2018). shiny imports it directly (`from prompt_toolkit.document import Document` in `shiny/_main_create_custom.py`) and pairs it with `questionary>=2.0.0`, whose supported range is prompt_toolkit 2–3. `>=3.0.0` is the oldest version of the modern 3.x line and is what the unit suite is verified against; there's no reason to claim support for the 2.x series.

pyproject.toml
```diff
@@ -63,18 +63,20 @@ dependencies = [
...
     "python-multipart>=0.0.7;platform_system!='Emscripten'",
     "setuptools;python_version>='3.12'",
     "narwhals>=1.10.0",
     "orjson>=3.10.7",
     "shinychat>=0.1.0",
-    "opentelemetry-api>=1.20.0",
+    # 1.38.0+ required for `Logger.emit(body=..., ...)` keyword arguments
+    # (older versions silently no-op shiny's otel log emission)
+    "opentelemetry-api>=1.38.0",
```
schloerke: **Why 1.20.0 → 1.24.0:** below sdk 1.24.0, `Span.record_exception()` formatted the stack trace from the *active* exception instead of the exception object passed in, so even though shiny passes a **sanitized** exception (`shiny/otel/_span_wrappers.py`), the original unsanitized error message leaked into span events despite `sanitize_errors=True` — `test_otel_error_handling.py::test_sanitized_exception_message_in_span` fails on sdk ≤1.23, passes from 1.24.0 (and each sdk requires its exact matching api version). Separately, the oldest-deps job found that shiny's `Logger.emit(body=..., ...)` keyword call only exists since opentelemetry 1.38.0 — on older versions the `TypeError` was silently swallowed and log records were dropped; rather than requiring 1.38.0 (released 2025-10, too new for a runtime floor), `emit_otel_log()` now falls back to constructing the `LogRecord` manually on older APIs.

pyproject.toml
```diff
@@ -63,18 +63,20 @@ dependencies = [
...
 ]
 
 [project.optional-dependencies]
 theme = ["libsass>=0.23.0", "brand_yml>=0.2.0"]
-otel = ["opentelemetry-sdk>=1.20.0"]
+otel = ["opentelemetry-sdk>=1.38.0"]
```
schloerke: **Why 1.20.0 → 1.24.0:** the sdk floor moves with the `opentelemetry-api>=1.24.0` runtime floor (sdk releases pin their exact api version). 1.24.0 is where `Span.record_exception()` stopped formatting the stack trace from the active exception, which leaked unsanitized error messages despite `sanitize_errors=True` on older sdks.

pyproject.toml
```diff
@@ -116,7 +118,7 @@ test = [
     "dask[dataframe]",
     "pyarrow",
     "pyarrow-stubs",
-    "opentelemetry-sdk>=1.20.0",
+    "opentelemetry-sdk>=1.38.0",
```
schloerke: **Why 1.20.0 → 1.24.0:** same reasoning as the `otel` extra above — the sdk must match the `opentelemetry-api>=1.24.0` runtime floor, and sdk <1.24.0 leaked unsanitized error messages in span exception events.

pyproject.toml
schloerke: **Decision:** floor set at **1.24.0** (not 1.38.0). Requiring 1.38.0 (released 2025-10) was judged too aggressive for a runtime dependency, so `emit_otel_log()` gained a fallback (8d8536ac): it tries the 1.38.0+ keyword form of `Logger.emit()` and, on `TypeError`, constructs the `LogRecord` manually — attaching the current span context for trace correlation and using the SDK's record type so exporters that read `record.resource` keep working. Verified: all otel unit tests pass on both 1.24.0 (fallback path) and 1.42.x (keyword path); the oldest-deps CI job pins 1.24.0 and is green. Resolving.

pyproject.toml
schloerke: **Decision:** `otel` extra floor set at **1.24.0**, matching the `opentelemetry-api` runtime floor. The `Logger.emit()` keyword-form concern that originally pushed the floor to 1.38.0 is instead handled by a fallback in `emit_otel_log()` (8d8536ac). Oldest-deps CI job pins sdk 1.24.0 and is green. Resolving.

pyproject.toml
schloerke: **Decision:** same as the `otel` extra — floor set at **1.24.0**; the pre-1.38.0 `Logger.emit()` incompatibility is handled by the fallback in `emit_otel_log()` (8d8536ac). Note the test-suite side also has a name-compat shim: `test_otel_value_logging.py` falls back to `InMemoryLogExporter` on sdk < 1.39.0. Resolving.

pyproject.toml
schloerke: **Decision:** stands at `>=0.17.1` — the historical floor, re-verified against the current test suite (859+ unit tests pass). Note the neighboring uvicorn floor was independently raised to 0.23.0 on main (#2317) and merged into this branch; the oldest-deps job now exercises starlette 0.17.1 + uvicorn 0.23.0 together.

pyproject.toml
schloerke: **Decision:** stands at `>=3.0.0` — oldest release of the 3.x line questionary supports and that provides the `prompt_toolkit.document` API shiny imports; verified by the oldest-deps CI job.

PR #2358, opened 2026-07-13
fix: render data frames with non-string column names (#2115)

### LINKED ISSUE #2115: [Bug]: Issue when rendering DataGrid with numeric column name
### Component

Reactive Programming

### Severity

### Shiny Version

1.5.1

### Python Version

3.12.12

```python
from shiny.express import render

@render.data_frame
def table():
    df = pd.DataFrame([["a", 1], ["b", 2], ["c", 3], ["d", 4]], columns=[0, 1])
    return render.DataGrid(df)
```

### Behavior

Current behavior:
I attempted to render a pandas `DataFrame` that has numeric column names and it failed with an obscure error.

Expected behavior:
The `DataFrame` should have been rendered without issue with the numeric column names.

Proposed solution/context:
This is easily worked around by naming the columns something rather than letting them just be numeric (to be fair, this was an error on my part). 

However, it took some deep digging to identify the root cause. In `shiny.render._data_frame_utils._tbl_data.serialize_frame`, the following code appears:
```python
def serialize_frame(into_data: IntoDataFrame) -> FrameJson:

    data = as_data_frame(into_data)

    type_hints = [serialize_dtype(data[col_name]) for col_name in data.columns]
```

The first line converts the pandas `DataFrame` into a narwhals `DataFrame` and the second accesses each column by name. In the case of a numeric column name, i.e. your column name is `0`, then `data[0]` returns the first *row* of the `DataFrame` and it returns it as a `DataFrame` rather than a `Series`, causing the confusing error.

A more robust way of accessing a column in a narwhals `DataFrame` would be to use the `.get_column` method, which gets around this ambiguity: ` type_hints = [serialize_dtype(data.get_column(col_name)) for col_name in data.columns]`

### Error Messages (if any)

```shell
  File "{Python_env}\shiny\session\_session.py", line 1836, in output_obs
    value = await renderer.render()
            ^^^^^^^^^^^^^^^^^^^^^^^
  File "{Python_env}\shiny\render\_data_frame.py", line 1005, in render
    payload = value.to_payload()
  File "{Python_env}\shiny\render\_data_frame_utils\_datagridtable.py", line 168, in to_payload
    **serialize_frame(self.data),
      ~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "{Python_env}\shiny\render\_data_frame_utils\_tbl_data.py", line 240, in serialize_frame
    type_hints = [serialize_dtype(data[col_name]) for col_name in data.columns]
                  ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "{Python_env}\shiny\render\_data_frame_utils\_tbl_data.py", line 186, in serialize_dtype
    dtype: DType = col.dtype
                   ^^^^^^^^^
AttributeError: 'DataFrame' object has no attribute 'dtype'
```

### Environment

```markdown
OS: Windows 11
Browser: Firefox
Dependencies:
shiny 1.5.0
narwhals 2.10.0
pandas 2.3.3
```

eeshsaxena:
Fixes #2115.

## Problem

Rendering a `render.DataGrid` / `render.data_frame` from a `DataFrame` whose column names are non-string (e.g. numeric) fails with an obscure `AttributeError`:

```python
from shiny.express import render
import pandas as pd

@render.data_frame
def table():
    df = pd.DataFrame([["a", 1], ["b", 2], ["c", 3], ["d", 4]], columns=[0, 1])
    return render.DataGrid(df)
# AttributeError: 'DataFrame' object has no attribute 'dtype'
```

## Cause

In `shiny/render/_data_frame_utils/_tbl_data.py`, columns were accessed with `data[col_name]` on the narwhals frame:

```python
type_hints = [serialize_dtype(data[col_name]) for col_name in data.columns]
```

For a numeric column name, narwhals interprets `data[0]` as positional/row access and returns a **row frame** rather than the column, so `serialize_dtype` receives a `DataFrame` instead of a `Series`. Thanks to @… (the issue reporter) for the precise diagnosis.

The same ambiguous access exists in the cell-patch scatter path (`apply_frame_patches`), where the column name comes from `nw_data.columns[...]` and can likewise be numeric:

```python
nw_data[column_name].scatter(...)
```

## Fix

Use `data.get_column(col_name)` / `nw_data.get_column(column_name)` in both places, which unambiguously accesses a column by name.

```python
>>> serialize_frame(as_data_frame(pd.DataFrame([["a", 1], ["b", 2]], columns=[0, 1])))
# now returns columns=[0, 1], typeHints=["string", "numeric"], data=[["a", 1], ["b", 2]]
```

## Testing

- Added `test_serialize_frame_numeric_column_names` in `tests/pytest/test_render_data_frame_tbl_data.py`. It fails on `main` with the `AttributeError` and passes with this change.
- Full `test_render_data_frame_tbl_data.py` passes (42 tests); ruff check + format clean.
- Added a CHANGELOG entry under Bug fixes.

An AI assistant helped implement this fix (following the root-cause and suggested approach from the issue); I reviewed the change and ran the tests locally.

### DIFF
--- a/shiny/render/_data_frame_utils/_tbl_data.py
+++ b/shiny/render/_data_frame_utils/_tbl_data.py
@@ -169,7 +169,7 @@ def apply_frame_patches(
 
     # Upgrade the Scatter info to new column Series objects
     scatter_columns = [
-        nw_data[column_name].scatter(
+        nw_data.get_column(column_name).scatter(
             scatter_values["row_indexes"], scatter_values["values"]
         )
         for column_name, scatter_values in cell_patches_by_column.items()
@@ -237,7 +237,12 @@ def serialize_frame(into_data: IntoDataFrame) -> FrameJson:
 
     data = as_data_frame(into_data)
 
-    type_hints = [serialize_dtype(data[col_name]) for col_name in data.columns]
+    # Use `get_column` rather than `data[col_name]`: for a numeric column name
+    # (e.g. `0`) the latter is interpreted as row/positional access and returns a
+    # row frame instead of the column, which then fails downstream.
+    type_hints = [
+        serialize_dtype(data.get_column(col_name)) for col_name in data.columns
+    ]
 
     # TODO-future-barret; Swich serialization to "by column", rather than "by row"
     # * This would allow for a single column to be serialized in a single operation
--- a/shiny/render/_data_frame_utils/_tbl_data.py
+++ b/shiny/render/_data_frame_utils/_tbl_data.py
@@ -150,9 +150,11 @@ def apply_frame_patches(
     # This allows for a single column to be updated in a single operation (rather than multiple updates to the same column)
     #
     # In; patches: List[Dict[row_index: int, column_index: int, value: Any]]
-    # Out; cell_patches_by_column: Dict[column_name: str, List[Dict[row_index: int, value: Any]]]
+    # Out; cell_patches_by_column: Dict[column_name: Any, List[Dict[row_index: int, value: Any]]]
     #
-    cell_patches_by_column: dict[str, ScatterValues] = {}
+    # Column names are not necessarily strings (e.g. pandas' default integer labels),
+    # so the keys are typed as `Any`.
+    cell_patches_by_column: dict[Any, ScatterValues] = {}
     for cell_patch in patches:
         column_name = nw_data.columns[cell_patch["column_index"]]
         if column_name not in cell_patches_by_column:
--- a/shiny/render/_data_frame_utils/_tbl_data.py
+++ b/shiny/render/_data_frame_utils/_tbl_data.py
@@ -152,8 +152,11 @@ def apply_frame_patches(
     # In; patches: List[Dict[row_index: int, column_index: int, value: Any]]
     # Out; cell_patches_by_column: Dict[column_name: Any, List[Dict[row_index: int, value: Any]]]
     #
-    # Column names are not necessarily strings (e.g. pandas' default integer labels),
-    # so the keys are typed as `Any`.
+    # Narwhals types `get_column(name)`'s `name` as `str`, but its docs note that pandas
+    # does allow non-string column names and that they work when passed to it, as there
+    # is no ambiguity:
+    # https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.get_column
+    # This dict is not exported, so its key type is relaxed to match.
     cell_patches_by_column: dict[Any, ScatterValues] = {}
     for cell_patch in patches:
         column_name = nw_data.columns[cell_patch["column_index"]]
@@ -239,9 +242,6 @@ def serialize_frame(into_data: IntoDataFrame) -> FrameJson:
 
     data = as_data_frame(into_data)
 
-    # Use `get_column` rather than `data[col_name]`: for a numeric column name
-    # (e.g. `0`) the latter is interpreted as row/positional access and returns a
-    # row frame instead of the column, which then fails downstream.
     type_hints = [
         serialize_dtype(data.get_column(col_name)) for col_name in data.columns
     ]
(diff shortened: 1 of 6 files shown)


tests/pytest/test_render_data_frame_tbl_data.py
```diff
@@ -366,7 +380,6 @@ def test_dtype_coverage():
     errs: list[str] = []
 
     for dtype_name in dtype_names:
-
         # Skip known types or imports that are not dtypes
```
schloerke, line 457: Unrelated hunk: this removes a blank line in `test_dtype_coverage` with no other change to that function. Could you restore it so the diff stays scoped to the fix?

shiny/render/_data_frame_utils/_tbl_data.py
```diff
@@ -169,7 +169,7 @@ def apply_frame_patches(
 
     # Upgrade the Scatter info to new column Series objects
     scatter_columns = [
-        nw_data[column_name].scatter(
+        nw_data.get_column(column_name).scatter(
```
schloerke, line 177: This half of the fix is untested. `test_serialize_frame_numeric_column_names` only exercises `serialize_frame`, so if this line regressed, CI would stay green while editable data frames with numeric column names broke on the first cell edit.

Could you add a test that runs a cell patch through `apply_frame_patches` on a frame with `columns=[0, 1]`? It fails on `main` with `AttributeError: 'DataFrame' object has no attribute 'scatter'`, so it pins the behavior this line fixes.

shiny/render/_data_frame_utils/_tbl_data.py
```diff
@@ -169,7 +169,7 @@ def apply_frame_patches(
```
schloerke, line 174: Related to this change: the accumulator a few lines up is annotated `cell_patches_by_column: dict[str, ScatterValues]` (line 155), but its keys come from `nw_data.columns[cell_patch["column_index"]]` — which this PR just established can be an `int`. The annotation is what made the original `nw_data[column_name]` look safe.

Worth widening it (`dict[Any, ScatterValues]`) in the same PR so the type no longer asserts the thing this fix disproves. Pre-existing, but this is the line that proves it wrong.

schloerke (commented):
I rebased this onto `main` and resolved the `CHANGELOG.md` conflict for you (your entry had been written against an older revision of the Bug fixes section, and several of the surrounding bullets have since landed in reworded form). The changelog bullet now sits next to the other data frame entry and mentions the patch path too. `pytest tests/pytest/test_render_data_frame_tbl_data.py` passes (42 tests) and `make check-format check-lint` is clean.

The fix itself is right, and catching the second occurrence in `apply_frame_patches` is the part that makes this the better fix — without it, rendering would work and then the first cell edit would crash the same way. I confirmed both sites against narwhals directly:

```
d[0]                       -> Narwhals DataFrame (positional row access)
d.get_column(0)            -> Narwhals Series
d[0].scatter([0], ['z'])   -> AttributeError: 'DataFrame' object has no attribute 'scatter'
```

Three small things before this goes in — see the inline comments.

tests/pytest/test_render_data_frame_tbl_data.py
schloerke: Restored in c1df6e6f — I went ahead and made these changes myself rather than sending them back to you, since I was already rebasing the branch.

The test file's diff is now purely additive.

shiny/render/_data_frame_utils/_tbl_data.py
schloerke: Added `test_apply_frame_patches_numeric_column_names` in c1df6e6f. It builds a frame with `columns=[0, 1]` and patches cells in two different columns, so the group-by-column path is exercised, then asserts the patched result and that the input frame is not mutated in place.

Verified it actually pins this line — reverting `get_column` back to `nw_data[column_name]` makes it fail with the production error:

```
E  AttributeError: 'DataFrame' object has no attribute 'scatter'
   shiny/render/_data_frame_utils/_tbl_data.py:174
```

Writing this turned up a separate typing bug, now fixed in e628f929. My first draft patched the *numeric* column with `"30"` and pandas rejected it with `LossySetitemError` — you cannot write a `str` into an `int64` column. That is the normal case, not an edge case: the browser always sends an edited cell as a `str`, so a `set_patch_fn` has to coerce it to the column's type first. But `CellValue` was `TagNode`, which excludes `int`/`float`/`bool`/`None` — so the coercion our own shipped example performs was a type error:

```python
# shiny/api-examples/data_frame_data_view/app-core.py
@df_original.set_patch_fn
def _(*, patch: render.CellPatch) -> render.CellValue:
    if patch["column_index"] in [0, 2]:
        return float(patch["value"])
    return int(patch["value"])
```

That slipped through because `api-examples/` is not among the paths pyright checks. `CellValue` is now `Union[TagNode, int, float, bool, None]`, with `test_apply_frame_patches_non_string_values` covering it.

I stopped at the JSON-native scalars on purpose. `date`/`datetime` apply to the frame just fine, but processed patches reach the client through `AppSession._send_message()`, which is stdlib `json.dumps` — a `datetime` would patch the frame and then raise on send. Including them would type-check while the app broke, so patching temporal columns stays honestly unsupported. There is a comment at the definition explaining this.

shiny/render/_data_frame_utils/_tbl_data.py
schloerke: Widened to `dict[Any, ScatterValues]` in c1df6e6f, with a note on why, and the `Out;` comment above it updated to match.

Related, from the same commit's follow-up (e628f929): `CellPatchProcessed.value` was annotated `str | CellHtml` but `maybe_as_cell_html()` only upgrades HTML-like values and passes everything else through untouched, so it could always hold any `Jsonifiable`. Pyrefly caught it the moment `CellValue` was widened. Now `Jsonifiable | CellHtml`.

schloerke:
Thank you @eeshsaxena !

PR #2362, opened 2026-07-19
fix(ui): stringify selected in update_checkbox_group/update_radio_buttons

### LINKED ISSUE #2272: [Bug]: update_checkbox_group requires different selected than input_checkbox_group
### Component

UI (ui.*)

### Severity

### Shiny Version

1.61

### Python Version

3.14.2

```python
from shiny import App, reactive, ui

DICT = {"Foo": {0: "a", 1: "b", 2: "c"}, "Bar": {0: "alpha", 1: "beta", 2: "gamma"}}

problem_h2 = """
Bug: update_checkbox_group() requires different 'selected' setup for choices 
with integer keys than input_checkbox_group()
"""

problem_str = """
input_checkbox_group accepts selected keys as ints, but update_checkbox_group does not.
The intention is to have all the options checked by default. There is no problem when 
initiating the checkbox group, but there is a problem when it updates.
\n Bonus idea:\n
It would be great if you could just pass a bool as an argument to check all boxes passed.
"""

checkbox_init = {
    "label": "Select multiple options:",
    "choices": DICT["Foo"],
    "selected": [k for k in DICT["Foo"].keys()],
}

app_ui = ui.page_fluid(
    ui.h2(problem_h2),
    ui.p(problem_str),
    ui.input_radio_buttons(
        "select_key", "Select a dictionary:", list(DICT.keys()), selected="Foo"
    ),
    ui.card(
        ui.card_header("Bugged checkbox_group"),
        ui.p(
            "After selecting a dict above, the options will be unchecked. They should be checked."
        ),
        ui.input_checkbox_group("check_group_bad", **checkbox_init),
    ),
    ui.card(
        ui.card_header("This checkbox_group works properly"),
        ui.p("But it required an extra step of converting keys to strings"),
        ui.input_checkbox_group("check_group_good", **checkbox_init),
    ),
)


    @reactive.effect
    @reactive.event(input.select_key, ignore_init=True)
    def _():
        choices = DICT[input.select_key()]
        ui.update_checkbox_group(
            "check_group_bad", choices=choices, selected=[k for k in choices.keys()]
        )
        ui.update_checkbox_group(
            "check_group_good",
            choices=choices,
            selected=[str(k) for k in choices.keys()],
        )


app = App(app_ui, server)
```

### Behavior

Current: `update_checkbox_group()` with `selected=some_dict` where the type of some_dict is dict[int, str], does not check the boxes. This is unexpected, because `input_checkbox_group()` accepts dictionaries with int keys and quietly converts them to strings. 
Expected: `update_checkbox_group()` should quietly convert dictionary integers to string the same way `input_checkbox_group()` does for the `selected` parameter. 
Current workaound: converting int keys from dict to string works just fine. e.g `selected=[str(k) for k in choices_dict.keys()]`

### Error Messages (if any)

```shell
No errors, shiny silently doesn't check any boxes when none of the `selected=` hit.
```

### Environment

```markdown
OS: Windows 11
Browser: Chrome
Dependencies: None
```

eeshsaxena:
Closes #2272.

`input_checkbox_group` accepts integer choice keys in `selected`, but `update_checkbox_group` / `update_radio_buttons` sent them to the client unchanged, so integer keys never matched the (string) option values on update. Stringify the selected values sent to the client. Adds unit tests (this update path previously had only browser tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `render.data_frame` now renders data frames whose column names are non-string (e.g. numeric) instead of failing with an obscure error. Columns are now accessed via `get_column()`, which avoids narwhals interpreting a numeric column name as positional row access. (#2115)
+
 * Fixed `--launch-browser` so it no longer depends on INFO-level Uvicorn startup logs. Browser launch now works when logs are disabled or `--log-level=warning` is used. (#569)
 
 * `ui.output_data_frame` will now consistently order the filtered columns in ascending column order. (#2093)
--- a/shiny/render/_data_frame_utils/_tbl_data.py
+++ b/shiny/render/_data_frame_utils/_tbl_data.py
@@ -169,7 +169,7 @@ def apply_frame_patches(
 
     # Upgrade the Scatter info to new column Series objects
     scatter_columns = [
-        nw_data[column_name].scatter(
+        nw_data.get_column(column_name).scatter(
             scatter_values["row_indexes"], scatter_values["values"]
         )
         for column_name, scatter_values in cell_patches_by_column.items()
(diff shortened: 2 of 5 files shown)


cpsievert:
Thanks. Appears this also includes the fix being proposed here? https://github.com/posit-dev/py-shiny/pull/2358

eeshsaxena:
You're right, and thanks for spotting it - I branched this off #2358 rather than main, so it carries those commits too. That's my mistake.

Closing this one so #2358 can go in on its own. I'll resend the `update_checkbox_group`/`update_radio_buttons` stringify fix separately, branched off main.

PR #2363, opened 2026-07-22
ci: integrate Pyrefly but also keep Pyright

karangattu:
## What changed

- add Pyrefly 1.1.1+ to the development environment and commit a migrated `pyrefly.toml`
- keep Pyright as the required type checker during the evaluation
- add a dedicated CI job that runs both checkers three times in the same Ubuntu/Python 3.14 environment
- publish median and best wall time, peak RSS, exit status, diagnostics, affected files, coverage context, top categories, and exact diagnostic-location overlap to the job summary
- upload raw JSON, stderr, and machine-readable metrics as a CI artifact

## Evaluation decision

After CI completes, use the generated job summary and artifact to decide whether the time and memory savings justify resolving or baselining the Pyrefly-only diagnostics, then follow up by changing `check-types` only if diagnostic coverage is acceptable.


karangattu:
## CI evaluation

The three-run Ubuntu/Python 3.14 comparison completed successfully.

| Criterion | Pyright 1.1.411 | Pyrefly 1.1.1 | Assessment |
| --- | ---: | ---: | --- |
| Median checker time | 25.24s | 1.76s | Pyrefly is **14.37x faster** / **93.0% less time** |
| Best checker time | 25.08s | 1.74s | Stable across the measured runs |
| Peak RSS | 1,690.0 MiB | 658.2 MiB | Pyrefly uses **61.1% less peak memory** |
| Exit code | 0 | 1 | Pyrefly is not drop-in clean yet |
| Diagnostics | 0 | 374 | Requires migration/baselining before gating |
| Files with diagnostics | 0 | 147 | Broad compatibility impact |

Diagnostic breakdown from the uploaded artifact:

- 231 diagnostics are in `shiny/`; 143 are in `tests/`.
- 232 (62.0%) are `bad-argument-type`, including recurring differences around unpacked kwargs, literal-string collections, and generic/container variance.
- 74 (19.8%) are `missing-import`, mostly test-layout resolution for `conftest` (35) and `utils.deploy_utils` (26); these look primarily configuration-related.
- The remaining categories include bad assignments (16), missing attributes (10), overload differences (7), bad returns (6), and generic/type-variable differences.

### Recommendation

Do **not** replace the required Pyright gate immediately. The performance win is large enough to continue: first fix Pyrefly test import resolution, then baseline the remaining known diagnostics and run Pyrefly as a shadow/new-error gate. Burn down or classify the baseline—especially the 232 argument-type findings—before changing `make check-types` from Pyright to Pyrefly.

At the measured Linux rate, each invocation saves about **23.5 seconds** of checker time. The full non-draft matrix currently invokes type checking on five Ubuntu and five macOS jobs; the Linux result directly suggests about **117 seconds of aggregate Ubuntu runner time** saved per workflow, while macOS should be benchmarked before projecting the total matrix savings.

[Comparison run and job summary](https://github.com/posit-dev/py-shiny/actions/runs/29890885231) includes the raw `type-checker-comparison` artifact.

karangattu:
## Pyrefly new-error gate implemented

The follow-up migration work is now included and green in CI:

- explicit Pyrefly search roots eliminate all **74 `missing-import` diagnostics** without enabling the unsafe global fallback heuristic
- a narrow Playwright-only `conftest` override handles the repository two ambiguous top-level `conftest.py` modules
- **303 known diagnostics** are committed in `pyrefly-baseline.json` (302 current findings plus the alternate Uvicorn generated-stub diagnostic seen across supported environments)
- `make check-pyrefly` now returns 0 for the baseline and fails for any new unbaselined diagnostic
- `make update-pyrefly-baseline` provides an intentional refresh path
- Pyrefly is pinned to 1.1.1 so weekly checker releases cannot silently invalidate the gate
- the workflow runs the gate on pull requests and on pushes to `main` / `rc-*`; Pyright remains behind `make check-types`

The gate was integration-tested with a temporary unbaselined assignment error: the clean tree passed, the probe failed, and the probe was removed.

### Updated CI comparison

| Criterion | Pyright 1.1.411 | Pyrefly 1.1.1 |
| --- | ---: | ---: |
| Median checker time | 25.94s | 2.54s |
| Peak RSS | 1,688.7 MiB | 653.0 MiB |
| New diagnostics | 1 | 0 |
| Baselined diagnostics | 0 | 303 |

Pyrefly remains **10.23x faster**, uses **90.2% less checker time**, and approximately **61.3% less peak memory** while now acting as a real new-error gate. The one Pyright diagnostic shown by the earlier comparison was fixed in the report dataclass; the final repository Pyright and Narwhals checks are green.

[Final comparison run](https://github.com/posit-dev/py-shiny/actions/runs/29921139716)

cpsievert:
Nice, thanks! Just curious, have we fully translated `pyrightconfig.json`  into the corresponding pyrefly config? If not, maybe that could help bring down the number of errors?

> Pyrefly is pinned to 1.1.1 

I think we'll want to keep up with the latest releases -- appears they happen on a monthly (not weekly) basis, so hopefully it isn't too painful to keep up.

> 303 known diagnostics are committed 

I'd be curious to see how well Claude does at resolving (at least most of) these. Hopefully a lot of them are similar and pretty straightforward to fix?

karangattu:
## Full config translation and diagnostic cleanup

The follow-up requested in review is now in `70ee89873`.

### Config translation

- Added `permissive-ignores = true`, so existing `# pyright: ignore` comments apply during migration.
- Added the implicit Pyright `stubPath` default as `site-package-path = ["typings"]`.
- Kept `pythonPlatform = "All"` mapped to Linux, the documented Pyrefly fallback.
- Kept the complete Pyright ignore list as `project-excludes`.
- Kept `reportPrivateUsage = "none"` mapped to `no-access = "ignore"`.
- Documented settings with no Pyrefly 1.1.1 equivalent: import cycles, unused functions, and unnecessary `isinstance`.
- Removed the empty `examples/` subconfig: examples are excluded and Pyrefly subconfigs cannot change search roots.
- Used the migrated default preset rather than Pyrefly `strict`: Pyrefly strict adds mandatory `@override` and other checks that Pyright strict leaves disabled.
- Relaxed the dependency to `pyrefly>=1.1.1`; the lockfile keeps CI reproducible while allowing routine lockfile upgrades.

### Source fixes and baseline

I fixed concrete narrowing and annotation issues in the Uvicorn compatibility path, Playwright locator handling, reactive value names, selectize option validation, nav collection mutation, bookmark state validation, and bookmark-store environment parsing.

The committed baseline dropped from **303 to 169 entries** (**44.6% fewer**). The remaining baseline is dominated by 143 `bad-argument-type` findings, primarily Pyrefly generic-invariance disagreements for htmltools attribute dictionaries and pandas/Polars/Narwhals data frames.

### Fresh local comparison

| Criterion | Pyright 1.1.409 | Pyrefly 1.1.1 |
| --- | ---: | ---: |
| Median wall time, 3 runs | 13.92s | 1.31s |
| Best wall time | 13.23s | 1.26s |
| Gate exit code | 0 | 0 |
| New diagnostics | 0 | 0 |
| Baselined diagnostics | 0 | 169 |

Pyrefly used **90.6% less checker time** and was **10.63x faster** locally.

Verification completed: Black, isort, Flake8, Pyright, Pyrefly new-error gate, comparison tests, and 62 focused nav/reactive tests.

karangattu:
The updated Ubuntu CI comparison is green: Pyright 1.1.411 took 18.74s median and 1,690.5 MiB peak RSS; Pyrefly 1.1.1 took 1.90s and 645.4 MiB. That is 9.88x faster, 89.9% less checker time, and 61.8% less peak memory. Both gates reported zero new diagnostics; Pyrefly loaded the 169-entry baseline.

karangattu:
@cpsievert take a look now.

pyrefly.toml
```diff
@@ -0,0 +1,56 @@
+# Translation of pyrightconfig.json for the parallel CI evaluation.
```
schloerke: Can we move this config content to be within `pyproject.toml`?

shiny/playwright/controller/_base.py
```diff
@@ -83,7 +82,7 @@ def not_is_missing(x: R | MISSING_TYPE) -> TypeGuard[R]:
     return not isinstance(x, MISSING_TYPE)
 
 
-def all_missing(*args: object) -> TypeGuard[MISSING_TYPE]:
+def all_missing(*args: object) -> bool:
```
schloerke: This feels wrong. Maybe keep the old one?

pyproject.toml
```diff
@@ -140,6 +140,8 @@ dev = [
     "isort>=5.10.1",
     "libsass>=0.23.0",
     "brand_yml>=0.2.0",
+    # Keep the lockfile current with Pyrefly releases and refresh the baseline as needed.
```
schloerke:
```suggestion
```

pyrefly.toml
```diff
@@ -0,0 +1,56 @@
...
+#
+# Pyrefly's default project preset is the migration target for Pyright strict;
+# its similarly named strict preset additionally enables checks (including
+# mandatory @override decorators) that Pyright strict explicitly leaves off.
+#
+# Pyrefly 1.1.1 has no diagnostics corresponding to reportImportCycles,
+# reportUnusedFunction, or reportUnnecessaryIsInstance, so those Pyright
+# disables require no Pyrefly override. The examples/ execution environment is
+# also redundant because examples/ is excluded from both checkers.
+python-platform = "linux"
```
schloerke:
```suggestion
```

pyrefly.toml
```diff
@@ -0,0 +1,56 @@
+# Translation of pyrightconfig.json for the parallel CI evaluation.
+#
+# Direct mappings:
+# - pythonPlatform=All -> linux (Pyrefly's documented cross-platform fallback)
```
schloerke:
```suggestion
```

pyrefly.toml
```diff
@@ -0,0 +1,56 @@
...
+# Pyrefly's default project preset is the migration target for Pyright strict;
+# its similarly named strict preset additionally enables checks (including
+# mandatory @override decorators) that Pyright strict explicitly leaves off.
+#
+# Pyrefly 1.1.1 has no diagnostics corresponding to reportImportCycles,
+# reportUnusedFunction, or reportUnnecessaryIsInstance, so those Pyright
+# disables require no Pyrefly override. The examples/ execution environment is
+# also redundant because examples/ is excluded from both checkers.
+python-platform = "linux"
+infer-with-first-use = false
```
schloerke: This change may require a followup PR due to errors produced. (Or just update the baseline to be fixed for later)

```suggestion
infer-with-first-use = false
implicit-any = error
```

pyrefly.toml
```diff
@@ -0,0 +1,56 @@
...
+permissive-ignores = true
```
schloerke: Make a separate PR to remove this config line.

By removing the line, we'll need to change our `# pyright: ignore` to `# pyrefly: ignore`. Try specify the specific rule being ignored, ex: `# pyrefly: ignore[bad-override]`.

Take inspiration from https://pyrefly.org/blog/too-many-type-checkers/#the-polars-story

pyrefly.toml
```diff
@@ -0,0 +1,56 @@
...
+baseline = "pyrefly-baseline.json"
```
schloerke: Let's be sure we remove outdated ignores

```suggestion
unused-ignore = "error"
unused-type-ignore = "error"
baseline = "pyrefly-baseline.json"
```

pyrefly.toml
```diff
@@ -0,0 +1,56 @@
...
+# Direct mappings:
+# - pythonPlatform=All -> linux (Pyrefly's documented cross-platform fallback)
+# - ignore -> project-excludes
+# - stubPath's Pyright default (./typings) -> site-package-path
+# - reportPrivateUsage=none -> no-access="ignore"
+#
+# Pyrefly's default project preset is the migration target for Pyright strict;
+# its similarly named strict preset additionally enables checks (including
+# mandatory @override decorators) that Pyright strict explicitly leaves off.
+#
```
schloerke: No need for these comments

pyrefly.toml
```diff
@@ -0,0 +1,56 @@
...
+# Pyrefly 1.1.1 has no diagnostics corresponding to reportImportCycles,
+# reportUnusedFunction, or reportUnnecessaryIsInstance, so those Pyright
+# disables require no Pyrefly override. The examples/ execution environment is
+# also redundant because examples/ is excluded from both checkers.
```
schloerke: Be sure to comment this in the PR. But we can remove from here

schloerke (approved):
I approve of the migration. Love the speed gains

Let's remove the script and new gha workflow. Let's update the existing `make check types` to point to `make check-pyrefly`.

For the time being, let's have a single ubuntu latest python GHA job just run `make check-pyright` until we can get shinylive typings validated even if we use pyrefly and not pyright.

pyrefly.toml
schloerke: Since we can't remove pyright today, add a comment for now. 
```suggestion
# Until pyright is completely removed, we need to maintain `permissive-ignores = true`
# Once removed, convert all `# pyright: ignore` to `#pyrefly: ignore`. Use specific rules when possible
permissive-ignores = true
```

tests/pytest/test_type_checker_comparison.py
```diff
@@ -0,0 +1,91 @@
+from scripts.compare_type_checkers import (
```
schloerke: remove this test file

PR #2369, opened 2026-07-24
fix: Rename api-examples/theme to Theme so example loads on case-sensitive filesystems

### LINKED ISSUE #370: [bug]Theme examples missing on deployed site due to case-sensitive filesystem (api-examples/theme vs Theme)
## Summary

The `express.ui.Theme` (and `ui.Theme`) API docs show an **Examples** section locally on macOS but not on the deployed Linux build.

## Root Cause

`@add_example()` in py-shiny uses `func.__name__` to locate the example directory:

```python
fn_name = func.__name__   # → "Theme" (capital T)
example_dir = os.path.join(ex_dir_found, fn_name)  # → api-examples/Theme/
```

But the actual directory on disk is `shiny/api-examples/theme/` (lowercase `t`).

- **macOS** — case-insensitive filesystem: `api-examples/Theme/` resolves to `api-examples/theme/` ✅ example found
- **Linux CI** — case-sensitive filesystem: `api-examples/Theme/` does not exist → `ExampleNotFoundException` is caught silently, a `::warning` is printed, and `@add_example()` returns without modifying the docstring → **no Examples section generated**

The silent failure is in `_docstring.py`:

```python
except ExampleNotFoundException as e:
    print(f"::warning file={file},...::{fn_name} - {e}")
    return func  # ← silently skips the example
```

## Evidence

- The warning can be observed in the CI logs for the last `main` build:  
  https://github.com/posit-dev/py-shiny-site/actions/runs/23913127706
- The fix should live in the **py-shiny** repo: either rename `api-examples/theme/` → `api-examples/Theme/`, or change `@add_example()` to do a case-insensitive directory lookup.

## Affected Pages

- `/api/express/express.ui.Theme.html`
- `/api/core/ui.Theme.html`

## Notes

This is a latent bug that exists on `main` today — not introduced by the Quarto 1.9 upgrade. It is visible now because the local (macOS) build is being compared against the deployed (Linux) build.

schloerke:
## Problem

The API docs for `ui.Theme` / `express.ui.Theme` render an **Examples** section locally on macOS but not on the deployed Linux site (posit-dev/py-shiny-site#370).

`@add_example()` locates the example directory using `func.__name__`:

```python
fn_name = func.__name__                             # "Theme" (capital T, from `class Theme`)
example_dir = os.path.join(ex_dir_found, fn_name)   # -> shiny/api-examples/Theme
```

But the directory on disk was `shiny/api-examples/theme/` (lowercase `t`). The code never does a case-insensitive lookup — it asks the filesystem for `Theme`:

- **macOS** (case-insensitive FS): resolves to `theme/` → example found ✅
- **Linux CI** (case-sensitive FS): no `Theme/` dir → `ExampleNotFoundException`, swallowed as a `::warning`, docstring returned unchanged → **no Examples section** ❌

## Fix

Rename `shiny/api-examples/theme/` → `shiny/api-examples/Theme/` so the directory matches `Theme.__name__` exactly. Every other example directory already matches its target's `__name__` including case (`Chat/`, `Value/`, `ClientData/`, …); `theme/` was the sole outlier.

## Verification

- Only `shiny/api-examples/Theme` is now tracked (no lingering `theme`).
- Case-sensitive `os.listdir` membership check: `"Theme"` present, `"theme"` absent — so the exact-name lookup succeeds on Linux.
- With `SHINY_ADD_EXAMPLES=true`, `ui.Theme.__doc__` now contains an `Examples` section.

## Follow-up (not in this PR)

The lookup failure was invisible because a missing declared example is only a `::warning`, not a build error. The `no_example()` helper's very existence implies a missing example *should* be an error — the silent `except ExampleNotFoundException: ... return func` is a weakening of that contract. Making it a hard error surfaces ~18 other functions whose declared examples are currently missing (some dirs absent entirely, some missing an express-mode example). That triage is tracked separately in posit-dev/py-shiny-site#370's follow-up issue and should be its own PR.

Fixes posit-dev/py-shiny-site#370


schloerke:
Closing without merging. The premise here was wrong: on current `main` the lookup uses the explicit `@add_example(example_name="theme")` (which takes precedence over `func.__name__`), matching the lowercase `shiny/api-examples/theme/` directory case-exactly — so it already resolves correctly on case-sensitive (Linux) filesystems. Renaming the directory to `Theme/` while leaving `example_name="theme"` would actually *break* the docs build on Linux. Keeping the lowercase folder, consistent with the other `example_name=` dirs (`input_text`, `data_frame`, `offcanvas`, `row`, `modal`, ...).

schloerke:
Closing as **superseded**. This PR renames `shiny/api-examples/theme` → `Theme` so `Theme.__name__` matches the dir on case-sensitive filesystems. `main` has since solved the same problem the other way: `Theme` is now decorated with `@add_example(example_name="theme")`, which looks up the lowercase `theme/` dir directly — so it already resolves on Linux.

Verified on `main`: `shiny/ui/_theme.py` uses `@add_example(example_name="theme")`, the `theme/` dir loads under `SHINY_ADD_EXAMPLES=true` in both modes, and all three quartodoc builds (`core`, `express`, `testing`) pass clean with no example errors. The underlying hard-error follow-up (#2370) is also complete (via #2278) and now closed.

PR #2402, opened 2026-08-01
docs: clarify panel_conditional module IDs

### LINKED ISSUE #2394: docs: panel_conditional() docstring documents an id that cannot exist
## Problem

The `panel_conditional()` docstring documents a workaround for an input id that cannot exist.

`shiny/ui/_bootstrap.py` L161-162:

> ... the input id's, but if you do use them anyway, for example, `id = "foo.bar"`, you will have to use `input["foo.bar"]` instead of ``input.foo.bar`` to read the input

But Shiny rejects any id outside `[A-Za-z0-9_]`, so `"foo.bar"` can never be created in the first place:

```python
>>> from shiny._namespaces import resolve_id
>>> resolve_id("foo.bar")
ValueError: The string 'foo.bar' is not a valid id; only letters, numbers, and underscore are permitted
```

(`shiny/_namespaces.py` — `re_valid_id = re.compile("^\\.?\\w+$")`, enforced by `validate_id()`.)

So the advice is unreachable: a reader can never be in the situation it describes.

## Additional context

The one case where bracket syntax *would* seem necessary — module namespacing, where the real DOM id is `mod-show` — is handled automatically. `panel_conditional` emits `data-ns-prefix="mod-"` and the client narrows the `input` scope, so inside a module you write the **plain** id:

```python
# Inside a module — correct:
ui.panel_conditional("input.show", ...)

# Does NOT work:
ui.panel_conditional("input['mod-show']", ...)
```

Verified both directions in a running app.

## Suggested fix

Delete that sentence. Optionally replace it with a note that module-scoped conditions use the plain (unprefixed) id, since that is the question the current text looks like it is answering.

Found while documenting `panel_conditional` on the docs site (posit-dev/py-shiny-site#414). Verified against `shiny` 1.7.1.dev6.

nightcityblade:
Fixes #2394

## Summary

- remove advice for period-containing input IDs, which Shiny rejects during ID validation
- clarify that module conditions use unprefixed, module-local input IDs because Shiny applies the namespace automatically

## Testing

- `PYTHONPATH=. python -m pytest -o addopts= tests/pytest/test_ui.py -q` (5 passed)
- Black, isort, Flake8, compileall, and `git diff --check` on the changed file

### DIFF
--- a/shiny/ui/_bootstrap.py
+++ b/shiny/ui/_bootstrap.py
@@ -160,10 +160,8 @@ def panel_conditional(
     (Be sure not to modify the input/output objects, as this may cause unpredictable
     behavior.)
 
-    You are not recommended to use special JavaScript characters such as a period . in
-    the input id's, but if you do use them anyway, for example, `id = "foo.bar"`, you
-    will have to use `input["foo.bar"]` instead of ``input.foo.bar`` to read the input
-    value.
+    Within a module, refer to inputs by their unprefixed, module-local IDs (for example,
+    ``input.foo``). Shiny applies the module namespace prefix automatically.
 
     Tip
     ---


schloerke:
Thank you, @nightcityblade !

PR #2410, opened 2026-08-02
feat: Use app author's navset id for data-tabsetid instead of random …

### LINKED ISSUE #2298: feat: Use the app author's navset id for data-tabsetid instead of a random value
## Motivation

Navsets currently mint their `data-tabsetid` (and the derived `tab-<tabsetid>-<index>` pane ids) from a random integer, even when the app author has already supplied a unique `id=` to the navset. Deriving the tabset ID from that user-supplied id would give:

- **Readable, deterministic DOM ids**: `tab-my_tabs-0` instead of `tab-7776790190029-0` — easier debugging, CSS targeting, and browser-devtools work.
- **Stable markup across renders/processes**: friendlier for snapshot tests, HTML diffing, and caching; no order- or seed-dependence.
- **Guaranteed uniqueness for id'd navsets**: Shiny input ids must already be unique within a session (and module namespacing is applied via `resolve_id()`), so the collision question disappears entirely for these navsets.

## Context

#2296 bumped the random ID pool from ~1M to 9T values (13-digit strings, kept below 2**53) after a birthday-paradox collision produced two navsets with the same `data-tabsetid` on one page — causing duplicate DOM ids, broken Bootstrap tab targeting, and CI failures ([example run](https://github.com/posit-dev/py-shiny/actions/runs/28619266837/job/84870650451), analysis in #2285). That makes collisions effectively impossible, but the ids remain random and opaque. This issue proposes the complementary improvement: don't be random at all when the author has given us a better id.

Anonymous navsets (no `id=`) and `nav_menu()` dropdowns (which have no user-facing id) would keep the random fallback.

## Prototype

```diff
--- a/shiny/ui/_navs.py
+++ b/shiny/ui/_navs.py
@@ -1617,7 +1617,10 @@ def render_navset(
     selected: Optional[str],
     context: dict[str, Any],
 ) -> tuple[Tag, Tag]:
-    tabsetid = nav_random_int()
+    # Prototype: derive the tabset ID from the user-supplied (already
+    # module-resolved) input id when available; fall back to a random ID for
+    # anonymous navsets.
+    tabsetid = id if id is not None else nav_random_int()
 
     # Separate MetadataNodes from NavSetArgs.
     metadata_args: list[MetadataNode] = []
```

By the time `render_navset()` receives `id`, it is already module-resolved (`NavSet.__init__` calls `resolve_id_or_none()`), so modules produce e.g. `data-tabsetid="mymod-inner_tabs"` / `id="tab-mymod-inner_tabs-0"` with no extra work.

**Validated locally** with the prototype applied:
- Rendering produces the expected markup for plain and module-namespaced navsets.
- All 13 Playwright tests under `tests/playwright/shiny/components/nav/` pass on chromium (kitchensink, navset_bar, page_navbar, nav insert/update paths), confirming shiny.js tab switching, `nav_insert()`'s `getTabIndex()` href parsing, and the Playwright controllers all treat the tabset ID as an opaque string and work with non-numeric values.

## Considerations

- **Snapshot churn**: markup changes for every navset that has an `id`; py-shiny's own `tests/pytest/test_navs.py` expectations and any downstream snapshot tests would need a one-time update.
- **Divergence from R**: R shiny/bslib always use random tabset ids, so rendered markup would no longer be structurally identical across languages (cosmetic only; no shared JS relies on the ID being numeric).
- **Character set**: Shiny ids are restricted enough to be safe in DOM ids and attribute selectors; shiny.js already `$escape`s the value when building selectors.
- `navset_hidden()` and `page_navbar`/`navset_bar` flow through the same `render_navset()` path, so they inherit the behavior automatically.

## Should this live in bslib or py-shiny?

py-shiny. The tabset ID generation for these components is entirely in py-shiny Python code (`shiny/ui/_navs.py` — not vendored from bslib), and the client-side JS needs no changes, so py-shiny can ship it independently. If desired, a parallel parity issue could be filed against rstudio/bslib / rstudio/shiny for the R side, but nothing there blocks this.

pevolution-ahmed:
Closes #2298

## Summary

When a navset has a user-supplied `id=`, use it as the `data-tabsetid` (and the derived `tab-<tabsetid>-<index>` pane IDs) instead of minting a random integer.

**Before:** `data-tabsetid="7776790190029"` / `id="tab-7776790190029-0"`
**After:** `data-tabsetid="my_tabs"` / `id="tab-my_tabs-0"`

Anonymous navsets (`id=None`) and `nav_menu()` dropdowns keep the random fallback.

## Changes

- **`shiny/ui/_navs.py`** One-line change in `render_navset()`: `tabsetid = id if id is not None else nav_random_int()`
- **`tests/pytest/test_navs.py`** Updated the `test_navset_pill_markup` snapshot to expect the deterministic IDs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improvements
 
-* Navsets created with an `id` (e.g. `ui.navset_tab(id="tabs")`) now use that `id` as their `data-tabsetid`, so their tab panes get stable `tab-tabs-0` style DOM ids instead of ones built from a random integer. This makes the rendered markup reproducible across renders and easier to target from custom CSS and JavaScript. Navsets without an `id`, and `ui.nav_menu()` dropdowns, keep the random ID. (#2410)
+* Navsets created with an `id` (e.g. `ui.navset_tab(id="tabs")`) now use that `id` as their `data-tabsetid`, so their tab panes get stable `tab-tabs-0` style DOM ids instead of ones built from a random integer. This makes the rendered markup reproducible across renders and easier to target from custom CSS and JavaScript. Navsets without an `id`, and `ui.nav_menu()` dropdowns, keep the random ID. (Thanks, @pevolution-ahmed!) (#2410)
 
 ### Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `playwright.controller.OutputTextVerbatim` is deprecated alongside `ui.output_text_verbatim()` and now emits a `ShinyDeprecationWarning` when constructed. Please use `playwright.controller.OutputCode` instead. (#2097)
 
+### Improvements
+
+* Navsets created with an `id` (e.g. `ui.navset_tab(id="tabs")`) now use that `id` as their `data-tabsetid`, so their tab panes get stable `tab-tabs-0` style DOM ids instead of ones built from a random integer. This makes the rendered markup reproducible across renders and easier to target from custom CSS and JavaScript. Navsets without an `id`, and `ui.nav_menu()` dropdowns, keep the random ID. (#2410)
+
 ### Bug fixes
 
 * The `ui.Theme` API reference examples now run in Shinylive. Compiling a customized theme requires `libsass`, but Shinylive only auto-loads packages it finds in an app's top-level imports and `Theme.to_css()` imports `sass` lazily, so the examples died with an `ImportError`. The example directory now declares `libsass` in a `requirements.txt`. (#2386)
--- a/shiny/ui/_navs.py
+++ b/shiny/ui/_navs.py
@@ -1626,7 +1626,9 @@ def render_navset(
     selected: Optional[str],
     context: dict[str, Any],
 ) -> tuple[Tag, Tag]:
-    tabsetid = nav_random_int()
+    # Derive the tabset ID from the user-supplied (already module-resolved)
+    # input id when available; fall back to a random ID for anonymous navsets.
+    tabsetid = id if id is not None else nav_random_int()
 
     # Separate MetadataNodes from NavSetArgs.
     metadata_args: list[MetadataNode] = []
--- a/tests/pytest/test_navs.py
+++ b/tests/pytest/test_navs.py
@@ -2,14 +2,16 @@
 
 import contextlib
 import random
+import re
 import textwrap
 from typing import Generator
 
 import pytest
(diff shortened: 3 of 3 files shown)


shiny/ui/_navs.py
```diff
@@ -1626,7 +1626,9 @@ def render_navset(
     selected: Optional[str],
     context: dict[str, Any],
 ) -> tuple[Tag, Tag]:
-    tabsetid = nav_random_int()
+    # Derive the tabset ID from the user-supplied (already module-resolved)
+    # input id when available; fall back to a random ID for anonymous navsets.
+    tabsetid = id if id is not None else nav_random_int()
```
schloerke, line 1631: @cpsievert Any reason _not_ to do this? Thank you

pevolution-ahmed:
Hi @schloerke ,

Since this is the first time contributing to Shiny, I would appreciate any guidance on what I should do following.
Thanks in advance.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `playwright.controller.OutputTextVerbatim` is deprecated alongside `ui.output_text_verbatim()` and now emits a `ShinyDeprecationWarning` when constructed. Please use `playwright.controller.OutputCode` instead. (#2097)
 
+### Improvements
+
+* Navsets created with an `id` (e.g. `ui.navset_tab(id="tabs")`) now use that `id` as their `data-tabsetid`, so their tab panes get stable `tab-tabs-0` style DOM ids instead of ones built from a random integer. This makes the rendered markup reproducible across renders and easier to target from custom CSS and JavaScript. Navsets without an `id`, and `ui.nav_menu()` dropdowns, keep the random ID. (#2410)
```
schloerke:
```suggestion
* Navsets created with an `id` (e.g. `ui.navset_tab(id="tabs")`) now use that `id` as their `data-tabsetid`, so their tab panes get stable `tab-tabs-0` style DOM ids instead of ones built from a random integer. This makes the rendered markup reproducible across renders and easier to target from custom CSS and JavaScript. Navsets without an `id`, and `ui.nav_menu()` dropdowns, keep the random ID. (Thanks, @pevolution-ahmed!) (#2410)
```

shiny/ui/_navs.py
cpsievert: Seems ok

schloerke:
@pevolution-ahmed Thank you for your PR! 

We've done all we can for now. Waiting on the logic confirmation above and then will merge.

If there are other untouched Issues that you see that should have small / direct solutions (just like this PR), I'd be happy to receive PRs any time. (Esp. if I submitted the Issue.)

shiny/ui/_navs.py
schloerke: I'll submit a PR to bslib as well

Issue #2386, opened 2026-07-29
docs: `ui.Theme` Shinylive examples fail — need `requirements.txt` with `shiny[theme]`


schloerke:
## Problem

The `ui.Theme` / `express.ui.Theme` API-reference examples are broken in Shinylive on the docs site:

- https://shiny.posit.co/py/api/core/ui.Theme.html#examples
- https://shiny.posit.co/py/api/express/express.ui.Theme.html#examples

(Also reproducible on the v1.7.0 preview build: https://pr-406--pyshiny.netlify.app/api/core/ui.Theme.html)

`shared.py` builds a customized theme:

```python
my_theme = (
    ui.Theme("shiny", include_paths=Path(__file__).parent)
    .add_defaults(bslib_dashboard_design=True)
    .add_mixins(...)
    .add_rules(...)
)
```

Rendering that theme calls `Theme.to_css()`, which hits `check_theme_pkg_installed("libsass", "sass")` and raises `ImportError`, so the app never starts.

## Root cause

`sass` is imported lazily *inside* `Theme.to_css()` (`shiny/ui/_theme.py:517-518`), not at module top level. Shinylive picks packages to load by scanning the app's top-level imports, so it never sees `sass` and never loads `libsass` — even though libsass **is** bundled in Shinylive's prebuilt pyodide packages (`libsass 0.23.0` is in `pyodide-lock.json`; added in posit-dev/shinylive#160, Nov 2024, precisely to enable `ui.Theme()` in Shinylive).

The only way to get it loaded is to list it in the app's `requirements.txt`.

## Proposed fix

Add `shiny/api-examples/Theme/requirements.txt` containing:

```
shiny[theme]
```

Why this works:

- `@add_example()` already collects every non-`app-*.py`, non-`__*` file in the example dir as a support file (`shiny/_docstring.py:188-204`), so a `requirements.txt` is bundled into the Shinylive payload automatically. Precedent: `shiny/api-examples/todo_list/requirements.txt` (contains `shinyswatch`).
- Shinylive's `_install_requirements_from_dir()` reads `requirements.txt` and micropip-installs each entry, and it explicitly understands **extras on an already-installed package** — it reads the installed dist's metadata and installs just the extra's requirements. The upstream implementation literally uses our case as its worked example (`src/hooks/usePyodide.tsx`: `# Convert requires records : 'libsass>=0.23.0; extra == "theme"'`).
[...]

PR #2418, opened 2026-08-06
docs: Register input_bookmark_button in the Express API reference

### LINKED ISSUE #2034: [Bug]: input_bookmark_button documentation not available for core or express syntax
### Component

Documentation

### Severity

### Shiny Version

1.4.0

### Python Version

3.10

```python
N/A
```

### Behavior

The Core or the express Function reference page does not have `input_bookmark_button` entry even though there is a `@add_example()` decorator on that function -> https://github.com/posit-dev/py-shiny/blob/main/shiny/bookmark/_button.py#L15

### Error Messages (if any)

```shell

```

### Environment

```markdown
macOS 15.5
```

jat255:
Fixes #2034: `[Bug]: input_bookmark_button documentation not available for core or express syntax`

## Current state of the bug

The core docs (`docs/_quartodoc-core.yml`) already had `ui.input_bookmark_button` (added by #1870), but the express docs (`docs/_quartodoc-express.yml`) are missing it. It is the *only* `input_*` symbol exported from `shiny.express.ui` with no Express config entry, and the only `ui.input_*` present in core but absent from Express.

Documenting `dir` turned up a behavioural detail worth stating in the docs: it is populated only for `store="server"`. The `"url"` store encodes state into the query string and never touches disk, so `dir` stays `None` there.

The rest of core's Bookmarking section was missing from Express too: `bookmark.restore_input`, `Bookmark`, `BookmarkState`, `RestoreState`, `set_global_save_dir_fn`, `set_global_restore_dir_fn`. Those are mode-agnostic symbols Express users need, so this PR mirrors the whole section rather than adding a single-entry section.

## History of bookmark docs changes

#1870 (`feat: Add support for bookmarking via shiny.bookmark`) added the Bookmarking section to `_quartodoc-core.yml` and the `shiny.express.ui` export, but never touched `_quartodoc-express.yml`. That same commit shipped `api-examples/input_bookmark_button/app-express.py`, so this appears to have been an unintentional omission

## Changes

- `docs/_quartodoc-express.yml`: add a "Bookmarking" section mirroring core's — `express.ui.input_bookmark_button`, then `bookmark.restore_input`, `bookmark.Bookmark`, `bookmark.BookmarkState`, `bookmark.RestoreState`, and core's `kind: page` "Integration" group wrapping `set_global_save_dir_fn` / `set_global_restore_dir_fn`. Placed immediately before "Chat interface" to match core, reusing its title/description.
- `shiny/bookmark/_button.py`: replace the roxygen link `[input_action_button()]` with `` :func:`~shiny.ui.input_action_button` `` per `.claude/references/documentation-style.md`. Also dropped the R name `bookmarkButton` from the same sentence and wrapped the long line.
- `shiny/bookmark/_bookmark.py`, `_save_state.py`, `_restore_state.py`: add the missing docstrings for `Bookmark`, `BookmarkState` and `RestoreState`. All three were registered in the reference with *no* docstring, so their rows in the Bookmarking index rendered with an empty summary column and their pages showed nothing but a signature. `BookmarkState` / `RestoreState` also had bare attribute annotations, leaving the fields app authors touch from `on_bookmark()` / `on_restore()` callbacks undocumented. Also fixes two stale references found in the same area: `on_bookmark()` pointed at `ShinySaveState` (renamed to `BookmarkState`), and `store` said "diabled".
- `CHANGELOG.md`: entry under `## [Unreleased]`, following the precedent of existing docs-only entries (#2386).

Note for review: this is the first `kind: page` group in the Express config, though core has used the pattern for a while. It builds and flattens correctly (below), but it is a new shape for this file.

Here's what it looks like built locally:

image

For comparison, the core doc page:

image

## Verification

Built the docs manually and verified they render correctly, as in the above screenshots.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `ui.input_task_button(type=None)` no longer drops the `bslib-task-button` class. Operator precedence made the `type is not None` check apply to the whole class string rather than just the Bootstrap classes, so the button rendered with `class=""`; since that class is the selector Shiny's input binding uses, the button was never bound as an input and clicking it did nothing. (#2388)
 
-* `ui.input_bookmark_button()` now appears in the Express API reference. It was the only `ui.input_*` function missing from `docs/_quartodoc-express.yml`, so its Express docs page was never generated even though `shiny.express.ui.input_bookmark_button()` is supported. Its docstring also carried a Shiny for R leftover (`[input_action_button()]`), which is now a working cross-reference. (#2418)
+* `ui.input_bookmark_button()` was added to the Express API reference. (#2418)
 
 ## [1.7.0] - 2026-07-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `ui.input_task_button(type=None)` no longer drops the `bslib-task-button` class. Operator precedence made the `type is not None` check apply to the whole class string rather than just the Bootstrap classes, so the button rendered with `class=""`; since that class is the selector Shiny's input binding uses, the button was never bound as an input and clicking it did nothing. (#2388)
 
+* `ui.input_bookmark_button()` now appears in the Express API reference. It was the only `ui.input_*` function missing from `docs/_quartodoc-express.yml`, so its Express docs page was never generated even though `shiny.express.ui.input_bookmark_button()` is supported. Its docstring also carried a Shiny for R leftover (`[input_action_button()]`), which is now a working cross-reference. (#2418)
+
 ## [1.7.0] - 2026-07-28
 
 ### New features
--- a/docs/_quartodoc-express.yml
+++ b/docs/_quartodoc-express.yml
@@ -195,6 +195,19 @@ quartodoc:
       desc: Saving and restoring app state
       contents:
         - express.ui.input_bookmark_button
+        - bookmark.restore_input
+        - bookmark.Bookmark
+        - bookmark.BookmarkState
+        - bookmark.RestoreState
+        - kind: page
+          path: bookmark_integration
+          summary:
+            name: "Integration"
+            desc: "Decorators to set save and restore directories."
+          flatten: true
+          contents:
+            - bookmark.set_global_save_dir_fn
+            - bookmark.set_global_restore_dir_fn
     - title: Dynamic UI
       desc: Dynamically show/hide UI elements.
       contents:
--- a/docs/_quartodoc-express.yml
+++ b/docs/_quartodoc-express.yml
@@ -94,6 +94,23 @@ quartodoc:
         - express.ui.insert_nav_panel
         - express.ui.remove_nav_panel
         - express.ui.update_nav_panel
+    - title: Bookmarking
+      desc: Saving and restoring app state
+      contents:
+        - express.ui.input_bookmark_button
+        - bookmark.restore_input
+        - bookmark.Bookmark
+        - bookmark.BookmarkState
+        - bookmark.RestoreState
+        - kind: page
+          path: bookmark_integration
+          summary:
+            name: "Integration"
+            desc: "Decorators to set save and restore directories."
+          flatten: true
+          contents:
+            - bookmark.set_global_save_dir_fn
+            - bookmark.set_global_restore_dir_fn
     - title: Chat interface
       desc: Build a chatbot interface
       contents:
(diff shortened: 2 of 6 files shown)


jat255:
@schloerke, While adding the missing Express section in this PR, I noticed that docstrings were missing for `Bookmark`,`BookmarkState`, and `RestoreState`.

Do you mind taking a look over the docstrings that Claude added in `shiny/bookmark/_bookmark.py`, `shiny/bookmark/_restore_state.py`, and `shiny/bookmark/_save_state.py` to sanity check them before we merge this?

shiny/bookmark/_bookmark.py
```diff
@@ -30,6 +30,21 @@
...
+    """
+    A session's bookmarking interface.
+
+    Every session exposes one of these as `session.bookmark`. It controls whether
+    bookmarking is enabled (`store`), which inputs are left out of a bookmark
+    (`exclude`), and when state is saved and restored (the `on_bookmark`,
+    `on_bookmarked`, `on_restore` and `on_restored` callback registrations).
+
+    Calling the object (`await session.bookmark()`) is equivalent to
+    `do_bookmark()`: it saves the current state and, depending on `store`, updates
```
schloerke: Expand on the function name


```suggestion
    `await session.bookmark.do_bookmark()`: it saves the current state and, depending on `store`, updates
```

docs/_quartodoc-express.yml
```diff
@@ -94,6 +94,23 @@ quartodoc:
...
+        - bookmark.RestoreState
+        - kind: page
+          path: bookmark_integration
+          summary:
+            name: "Integration"
+            desc: "Decorators to set save and restore directories."
+          flatten: true
+          contents:
+            - bookmark.set_global_save_dir_fn
+            - bookmark.set_global_restore_dir_fn
```
schloerke, line 113: I originally didn't add these as they are not express functions. While can be used in express mode, they behave similar to normal core bookmarking.

shiny/bookmark/_restore_state.py
```diff
@@ -21,9 +21,38 @@
...
 
 class RestoreState:
+    """
+    The state read back when a session is restored from a bookmark.
+
+    Shiny creates one of these while a bookmarked session starts up and passes it to any
+    `session.bookmark.on_restore()` callback. Within a module, the state is scoped to
+    that module, so names are unprefixed.
+
+    Input values are usually restored for you — :func:`~shiny.bookmark.restore_input`
```
schloerke:
```suggestion
    Input values are restored using :func:`~shiny.bookmark.restore_input` which
```

shiny/bookmark/_save_state.py
```diff
@@ -17,12 +17,33 @@
 
 
 class BookmarkState:
+    """
+    The state being written while a bookmark is saved.
+
+    Shiny creates one of these each time a bookmark is taken and passes it to any
+    `session.bookmark.on_bookmark()` callback, which may add entries to
+    `values` or write files into `dir` before the state is persisted.
```
schloerke:
```suggestion
    Shiny creates a `BookmarkState` instance each time a bookmark is taken and passes the instance to every
    `session.bookmark.on_bookmark()` callback. The registered callbacks may add entries to
    the state's `values` or write files into `dir` (if available) before the state is persisted.
```

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `ui.input_task_button(type=None)` no longer drops the `bslib-task-button` class. Operator precedence made the `type is not None` check apply to the whole class string rather than just the Bootstrap classes, so the button rendered with `class=""`; since that class is the selector Shiny's input binding uses, the button was never bound as an input and clicking it did nothing. (#2388)
 
+* `ui.input_bookmark_button()` now appears in the Express API reference. It was the only `ui.input_*` function missing from `docs/_quartodoc-express.yml`, so its Express docs page was never generated even though `shiny.express.ui.input_bookmark_button()` is supported. Its docstring also carried a Shiny for R leftover (`[input_action_button()]`), which is now a working cross-reference. (#2418)
```
schloerke:
```suggestion
* `ui.input_bookmark_button()` was added to the Express API reference. (#2418)
```

docs/_quartodoc-express.yml
schloerke: If you'd like them here, no problem. But it felt _technically_ out of place when I first wrote it.

For sure keep the `input_bookmark_button`

schloerke (approved):

docs/_quartodoc-express.yml
jat255: I don't have any strong opinion, but in API docs, I don't think verbosity is too much of an issue, so if they're useful in the context, I would keep them.

cpsievert (approved):

PR #2049, opened 2025-08-05
fix: Remove HTML Label Option

elnelson575:
Removes the option (which never really worked) to give `choices` parameter HTML labels.

Closes #1330

PR #2420, opened 2026-08-06
fix: Coerce non-string `selected` in checkbox/radio update functions

### LINKED ISSUE #2272: [Bug]: update_checkbox_group requires different selected than input_checkbox_group

jat255:
Fixes #2272

## Problem

`ui.input_checkbox_group()` and `ui.input_radio_buttons()` accept non-string choice values, such as the `int` keys of a `dict[int, str]`. The `update_*()` functions did not.

Choice values become HTML `value` attributes, so the client always sees them as strings. The constructors compared the raw choice values against the raw `selected`, so both sides agreed. `_update_choice_input()` sent `selected` without change, so the client received `{"value": [0, 1, 2]}` instead of `{"value": ["0", "1", "2"]}`.

This is not a quiet mismatch. `CheckboxGroupInputBinding.setValue` clears every box, then calls `$escape()` on each value. `$escape()` calls `.replace()`, which throws a `TypeError` on a number. This is why the reporter saw no boxes checked.

`update_radio_buttons()` and `update_select()` had the same cause and a different result. Both throw before they change anything, so a label update in the same message is lost.

## Fix

The five choice inputs (checkbox group, radio buttons, select, selectize, toolbar select) now share one normalization module, `shiny/ui/_choices.py`. Each entry point normalizes `choices` into a canonical `dict` keyed by the string form, then resolves `selected` against that dict's keys once. Everything downstream reads those resolved values, including the rendered options, the `selected` attributes, and the `value` in the `update_*()` payload. This ensures the markup and the payload cannot disagree, because neither re-derives the value.

This decision differs from a more simple fix at the wire boundary, which is the approach that gave rise to the issue in `_update_choice_input()` raised in #2272. The constructors, the select inputs, and the toolbar each carried their own comparison, leading to possible inconsistencies. Normalizing at the entry point closes the entire class of bug, but has a larger surface area as a result.

This PR also widens the type annotations, so the documented calls pass a type checker. A choice value becomes `ChoiceValue = str | int | float | bool | None`. The key positions of a choice mapping stay loosely typed as `ChoiceKey = Any`. (This is because a `Mapping` type's key parameter is invariant, so a checker demands an exact match there rather than a subtype. `Mapping[ChoiceValue, str]` therefore accepts only a literal `dict[ChoiceValue, str]`, and rejects `dict[int, str]` and `dict[str, str]` alike, even though those are valid transitively. The "proper" fix for this would likely be to add more extensive bounded `TypeVar`s, though that would expand the scope even more).

`update_radio_buttons()` collapses a list or tuple `selected` to a single value, because the binding throws on a non-empty array. An empty `selected` stays an empty list. That is the documented way to clear a selection, and the one array shape the binding accepts.

## A separate bug, found while testing this PR

`ui.update_radio_buttons(choices=[])` raised `IndexError: list index out of range` instead of clearing the choices. The shared note on the `update_*` functions documents `choices=[]` as the way to clear the set of choices. That text dates from #42, in February 2022, so the call has never worked for radio buttons. `ui.update_checkbox_group()` and `ui.update_select()` both honor it and worked fine.

(see this note in `shiny/ui/_input_update.py`, on `main`): 
image

The radio group failed because an empty group has no first option to fall back on when `selected` is omitted. It now sends the same payload as `ui.update_checkbox_group()` for the same call. The constructors still reject an empty `choices` with no `selected`.

No open issue covers this. #2400 is adjacent but different: it asks to clear the *selection* rather than the choices, and `selected=[]` already does that.

## Behavior changes (please double check these are acceptable!)

Some of these changes alter how existing apps might behave, so they're work calling out explicitly:

- `None` and `bool` choice values now render `value="None"`, `value="False"`, and `value="True"`. htmltools dropped the attribute before, or wrote `value=""` for `True`, so the browser reported the default `"on"`. The fix requires this, because the client matches on the `value` attribute and an absent attribute leaves nothing to match. Bookmarks that hold the old values now read the new strings.
  - This manifests as follows: if you have a checkbox group like
    ```
    ui.input_checkbox_group(
        "checkbox_group", 
        "Choices:",
        {None: "None key", False: "False key", True: "True key"}
    )
    ```
    And then a button that runs `ui.update_checkbox_group("checkbox_group", selected=[True, False])`, you'll get this on current `main`: 
    image 
    vs. this after this PR: 
    image
    I believe this is more correct.
- The constructors accept mixed types now, for example `choices={0: "a"}` with `selected=["0"]`. This follows from the string coercion that we're doing. It is perhaps less formally correct, though.
    - For example, on `main`, this radio would render unselected because `1 != "1"`:
        image
    - After this PR, as a consequence of the string coercion, this would now select the "b" radio box:
        image
- Choice values that collide once stringified, such as `{0: "zero", "0": "oh"}`, now raise a `ValueError`. Coercion to `str` turns two distinct values into one key, so without this error one option disappears with no message.
    - On `main`, the server would not be aware that these were different. After this PR, there's an explicit error raised:
       image
- `ui.input_radio_buttons(choices=[])` raises a `ValueError` that names `choices`, instead of an `IndexError` from a private helper. This applies to the constructors only. The `update_*()` path now clears the group instead, as described above. The fix does not require this error. It is a deliberate addition, because an empty filter result is a realistic way to reach that line.
  - On `main`, this ended up looking like the following. If you created `input_radio_buttons(choices=[])` with an empty choice list, you'd get `IndexError: list index out of range` raised in [`_generate_options()#L328`](https://github.com/posit-dev/py-shiny/blob/d4714940b8071d80e18114273aeb2294140aedff/shiny/ui/_input_check_radio.py#L328). After this PR, you'll get `ValueError: `choices` cannot be empty for a radio button group unless `selected` is given.` from [here](https://github.com/posit-dev/py-shiny/blob/aa4a301419c02f00fc04707cbb4c63f454d11b3b/shiny/ui/_input_check_radio.py#L346-L349). Likewise, calling `update_radio_buttons(choices=[])` will actually clear the options for the radio list, rather than the mysterious `IndexError`.

## Judgement-call decisions

This PR's implementation now resolves the values in `selected` against the actual choice values. i.e. `choices={1: "a"}` with `selected=[1.0]` both renders and sends `"1"`. A comparison on raw equality is not correct here. The old code gave the right answer for `1.0` and the wrong one for `"1"`, and it compared values the client never sees: the option carries `value="1"`, so a `selected` of `1.0` renders as checked but goes onto the wire as a `1.0` that no option carries.

An `update_*()` message has two halves: the regenerated options markup and the `value`. Both now come from a single coercion, so they cannot disagree. A fix that coerces only the `value` leaves the markup with nothing checked for a mixed-type call, such as `choices={0: "a"}` with `selected=["0"]`. The client then re-checks the box from the `value`, which hides the mismatch instead of removing it.

## Prior art

#2362 diagnosed this problem first and is now closed. Credit to @eeshsaxena. That PR changed `_update_choice_input()` only, which covers `update_checkbox_group()` and `update_radio_buttons()`. This PR also fixes `update_select()`, `update_selectize()`, and `toolbar_input_select()`, and it changes the input constructors. #2362 branched from #2358 rather than `main`, so it carried unrelated data-frame commits.

## Out of scope

The issue also asks for a `bool` `selected` value that checks every box. That is a separate feature request, that I think would be a departure from existing API patterns and this PR does not implement it.

## Testing

`tests/pytest/test_input_check_radio.py` covers the wire payload, the rendered markup, the new errors, and the select inputs. The stub session runs each payload through `json.dumps()`, so the tests cover the wire format and not only the dictionary.

1087 unit tests pass locally. `uv run make format check-lint check-types check-pyright` is clean, which covers both pyrefly and pyright. The Playwright input and toolbar tests pass (85 tests). Pyright also reports no errors on a file that holds the documented calls.

### DIFF
--- a/shiny/ui/_choices.py
+++ b/shiny/ui/_choices.py
@@ -30,34 +30,40 @@ V = TypeVar("V")
 
 
 def duplicate_key_error(
-    key: str, first: Any, second: Any, *, keys_can_be_optgroup_labels: bool
+    key: str, *, first: tuple[Any, Any], second: tuple[Any, Any]
 ) -> ValueError:
-    if keys_can_be_optgroup_labels:
-        what = "choice value or optgroup label"
-        reason = (
-            "A select input's `choices` (values and optgroup labels) must be unique "
-            "when converted to strings."
-        )
-    else:
-        what = "choice value"
-        reason = (
-            "Choice values must be unique when converted to strings strings."
+    """
+    Build the error for two keys whose string forms collide.
+
+    ``first`` and ``second`` are the colliding ``(key, label)`` pairs. A key whose label
+    is a ``Mapping`` heads an optgroup, so it names a group rather than a choice. Only a
+    select input's top-level ``choices`` can hold one, so the wording follows from the
+    labels rather than from the calling input.
+    """
+    (first_key, first_label), (second_key, second_label) = first, second
+    kinds = tuple(
+        "optgroup label" if isinstance(label, Mapping) else "choice value"
+        for label in (first_label, second_label)
+    )
+
+    if kinds == ("choice value", "choice value"):
+        return ValueError(
+            f"Duplicate choice value {key!r}: {first_key!r} and {second_key!r} are "
+            "distinct but are identical as strings. Choice values must be unique when "
+            "converted to strings."
         )
 
     return ValueError(
-        f"Duplicate {what} {key!r}: {first!r} and {second!r} are distinct but are "
-        f"identical as strings. {reason}"
+        f"Duplicate key {key!r} in `choices`: the {kinds[0]} {first_key!r} and the "
+        f"{kinds[1]} {second_key!r} are distinct but are identical as strings. A select "
+        "input's `choices` (values and optgroup labels) must be unique when converted "
+        "to strings."
     )
 
 
-def normalize_choices_mapping(
-    x: Mapping[Any, V], *, keys_can_be_optgroup_labels: bool = False
-) -> dict[str, V]:
+def normalize_choices_mapping(x: Mapping[Any, V]) -> dict[str, V]:
     """
     Coerce choice values (the mapping's keys) to ``str``, preserving order.
-
-    A select input's ``choices`` maps optgroup labels alongside choice values, so its
-    top level sets ``keys_can_be_optgroup_labels``, which only affects the error below.
 
     Raises
     ------
@@ -74,9 +80,8 @@ def normalize_choices_mapping(
         if str_key in normalized:
             raise duplicate_key_error(
                 str_key,
-                originals[str_key],
-                key,
-                keys_can_be_optgroup_labels=keys_can_be_optgroup_labels,
+                first=(originals[str_key], normalized[str_key]),
+                second=(key, label),
             )
         normalized[str_key] = label
         originals[str_key] = key
--- a/shiny/ui/_choices.py
+++ b/shiny/ui/_choices.py
@@ -81,7 +81,7 @@ def resolve_selected_values(selected: Any, choice_values: Mapping[str, Any]) -> 
 
     def resolve_one(value: Any) -> Any:
         if str(value) in choice_values:
-            return value
+            return choice_values[str(value)]
         for original in originals:
             if original == value:
                 return original
--- a/shiny/ui/_choices.py
+++ b/shiny/ui/_choices.py
@@ -15,7 +15,7 @@ https://github.com/posit-dev/py-shiny/pull/2420 for details.
 from __future__ import annotations
 
 from collections.abc import Collection, Iterable, Mapping
-from typing import Any, Literal, TypeVar, Union, cast
+from typing import Any, TypeVar, Union, cast
 
 # A choice value. The client only ever sees ``str(value)``, so these are the types with
 # an unambiguous string form -- the ones the choice inputs document and test.
@@ -26,29 +26,38 @@ ChoiceValue = Union[str, int, float, bool, None]
 # even `dict[str, str]`. The runtime contract is the same as `ChoiceValue`.
 ChoiceKey = Any
 
-# What a normalized mapping's keys hold. A select input's `choices` maps optgroup labels
-# alongside choice values, so its top level holds both.
-ChoiceKeyKind = Literal["choice value", "choice value or optgroup label"]
-
-_DUPLICATE_KEY_REASONS: dict[ChoiceKeyKind, str] = {
-    "choice value": (
-        "Choice values become HTML `value` attributes, so they must be unique as "
-        "strings."
-    ),
-    "choice value or optgroup label": (
-        "A select input's `choices` holds choice values and optgroup labels in one "
-        "mapping, so both must be unique as strings."
-    ),
-}
-
 V = TypeVar("V")
 
 
+def duplicate_key_error(
+    key: str, first: Any, second: Any, *, keys_can_be_optgroup_labels: bool
+) -> ValueError:
+    if keys_can_be_optgroup_labels:
+        what = "choice value or optgroup label"
+        reason = (
+            "A select input's `choices` (values and optgroup labels) must be unique "
+            "when converted to strings."
+        )
+    else:
+        what = "choice value"
+        reason = (
+            "Choice values must be unique when converted to strings strings."
+        )
+
+    return ValueError(
+        f"Duplicate {what} {key!r}: {first!r} and {second!r} are distinct but are "
+        f"identical as strings. {reason}"
+    )
+
+
 def normalize_choices_mapping(
-    x: Mapping[Any, V], *, keys_are: ChoiceKeyKind = "choice value"
+    x: Mapping[Any, V], *, keys_can_be_optgroup_labels: bool = False
 ) -> dict[str, V]:
     """
     Coerce choice values (the mapping's keys) to ``str``, preserving order.
+
+    A select input's ``choices`` maps optgroup labels alongside choice values, so its
+    top level sets ``keys_can_be_optgroup_labels``, which only affects the error below.
 
     Raises
     ------
@@ -63,10 +72,11 @@ def normalize_choices_mapping(
     for key, label in x.items():
         str_key = str(key)
         if str_key in normalized:
-            raise ValueError(
-                f"Duplicate {keys_are} {str_key!r}: {originals[str_key]!r} and {key!r} "
-                f"are distinct but are identical once converted to a string. "
-                + _DUPLICATE_KEY_REASONS[keys_are]
+            raise duplicate_key_error(
+                str_key,
+                originals[str_key],
+                key,
+                keys_can_be_optgroup_labels=keys_can_be_optgroup_labels,
             )
         normalized[str_key] = label
         originals[str_key] = key
--- a/shiny/ui/_choices.py
+++ b/shiny/ui/_choices.py
@@ -1,0 +1,197 @@
+"""
+Shared normalization for the choice-based inputs (checkbox group, radio buttons,
+select, selectize, toolbar select).
+
+Choice values become HTML ``value`` attributes, so the client only ever sees their
+string form. Everything here exists to make the server agree with that: choice values
+are coerced to ``str`` once, and ``selected`` is coerced the same way so that the
+regenerated options markup and the ``value`` sent in an ``update_*()`` message match
+by construction.
+
+See https://github.com/posit-dev/py-shiny/issues/2272.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Iterable, Mapping, TypeVar
+
+# A choice value.
+#
+# Deliberately `Any` rather than a union of the scalar types we expect. `Mapping`'s key
+# type is invariant, so a narrower annotation makes `Mapping[<narrow>, TagChild]` reject
+# `dict[int, str]` -- and, surprisingly, `dict[str, str]` as well. `Any` is also honest
+# about the runtime contract: any object is accepted and coerced with `str()`.
+ChoiceValue = Any
+
+_V = TypeVar("_V")
+
+
+def normalize_choices_indexed(
+    x: Mapping[Any, _V],
+) -> tuple[dict[str, _V], dict[str, Any]]:
+    """
+    Coerce choice values (the mapping's keys) to ``str``, preserving order.
+
+    Returns both the stringified mapping and an index from each string form back to the
+    original choice value, which :func:`resolve_selected_values` needs.
+
+    Raises
+    ------
+    ValueError
+        If two choice values collide once stringified. Silently dropping one of them
+        would make an option disappear from the rendered group with no diagnostic.
+    """
+    labels: dict[str, _V] = {}
+    originals: dict[str, Any] = {}
+
+    for key, label in x.items():
+        str_key = str(key)
+        if str_key in labels:
+            raise ValueError(
+                f"Duplicate choice value {str_key!r}: {originals[str_key]!r} and {key!r} "
+                "are distinct choices but are identical once converted to a string. "
+                "Choice values become HTML `value` attributes, so they must be unique "
+                "as strings."
+            )
+        labels[str_key] = label
+        originals[str_key] = key
+
+    return labels, originals
+
+
+def normalize_choices_mapping(x: Mapping[Any, _V]) -> dict[str, _V]:
+    """Coerce choice values (the mapping's keys) to ``str``, preserving order."""
+    return normalize_choices_indexed(x)[0]
+
+
+def resolve_selected_values(selected: Any, choice_values: Mapping[str, Any]) -> Any:
+    """
+    Replace each ``selected`` entry with the choice value it refers to.
+
+    Usually a no-op, since ``selected`` entries normally *are* choice values. It matters
+    when an entry compares equal to a choice value without stringifying identically:
+    ``selected=[1.0]`` against ``choices={1: "a"}`` resolves to ``[1]``, so it both
+    renders as and transmits ``"1"`` -- the value the option actually carries.
+
+    Without this step, matching on the string form alone would stop honoring a
+    numerically equal ``selected``, which used to mark the option checked. Matching on
+    raw equality *instead* is not an option either: that is what left the server marking
+    an option checked while sending the client a ``"1.0"`` it could never match.
+    """
+    if selected is None or not choice_values:
+        return selected
+
+    originals = list(choice_values.values())
+
+    def resolve_one(value: Any) -> Any:
+        if str(value) in choice_values:
+            return value
+        for original in originals:
+            if original == value:
+                return original
+        return value
+
+    if isinstance(selected, (str, bytes)) or not isinstance(selected, Iterable):
+        return resolve_one(selected)
+    return [resolve_one(value) for value in selected]
+
+
+def _as_raw_list(x: Any) -> list[Any]:
+    """
+    Coerce ``selected`` to a list of raw (un-stringified) choice values.
+
+    ``str`` and ``bytes`` are iterable but represent a single choice, so they are
+    treated as scalars. Any other iterable -- including the ``dict_keys``, ``set``, and
+    generator shapes that a caller may build a selection from -- is consumed as a
+    sequence. Handling them here keeps them from falling into the scalar branch, where
+    they would be stringified to their unusable ``repr``.
+    """
+    if x is None:
+        return []
+    elif isinstance(x, (str, bytes)) or not isinstance(x, Iterable):
+        return [x]
+    else:
+        return list(x)
+
+
+def normalize_selected(x: Any) -> str | list[str] | None:
+    """
+    Coerce ``selected`` to string(s), preserving scalar vs. sequence shape.
+
+    Shape matters on the wire: the checkbox group's client-side ``setValue()`` expects
+    an array while the radio group's expects a scalar, so a scalar ``selected`` must not
+    grow into a list on its way out. Tuples become lists purely for payload consistency
+    -- ``json.dumps()`` already serializes a tuple as a JSON array.
+    """
+    if x is None:
+        return None
+    elif isinstance(x, str):
+        return x
+    elif isinstance(x, bytes) or not isinstance(x, Iterable):
+        return str(x)
+    else:
+        return [str(v) for v in x]
+
+
+def normalize_selected_list(x: Any) -> list[str] | None:
+    """Coerce ``selected`` to a list of strings, for clients that expect an array."""
+    if x is None:
+        return None
+    return [str(v) for v in _as_raw_list(x)]
+
+
+def normalize_selected_scalar(x: Any) -> str | None:
+    """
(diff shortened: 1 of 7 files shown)


shiny/ui/_input_select.py
```diff
@@ -18,25 +18,41 @@
...
-    _OptGrpChoices,
+    # [0, 1, 2] or ["a", "b", "c"]
+    "list[ChoiceValue]",
+    # (0, 1, 2) or ("a", "b", "c")
+    "tuple[ChoiceValue, ...]",
+    # {"a": "Choice A", 0: "Choice B"}
+    Mapping[ChoiceValue, str],
+    # optgroup {"Group A": {"a1": "Choice A1", "a2": "Choice A2"}, "Group B": {}}
+    Mapping[ChoiceValue, Mapping[ChoiceValue, str]],
+]
```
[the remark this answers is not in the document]
jat255: The `str` label type here appears to be deliberate from a previous change: #2049 (commit `621fbafb`, "Remove HTML Label Option") deprecated and then removed support for non-string labels in `input_select()`/`input_selectize()`. That change narrowed `_Choices` from `Mapping[str, TagChild]` to `Mapping[str, str]`, removed the docstring language permitting non-string labels, and removed the deprecation warning that was `_contains_html()`'s only call site.

`_contains_html()` is leftover dead code from that deprecation; it has no callers on `main` or on this branch, so it isn't evidence of supported runtime behavior.

That said, this review usefully surfaces the dead code. @cpsievert, do you think we should remove `_contains_html()` as part of this PR, or leave that cleanup for a follow-up?

shiny/ui/_choices.py
```diff
@@ -0,0 +1,197 @@
...
+
+    originals = list(choice_values.values())
+
+    def resolve_one(value: Any) -> Any:
+        if str(value) in choice_values:
+            return value
+        for original in originals:
+            if original == value:
+                return original
+        return value
```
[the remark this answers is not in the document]
jat255: Fixed in https://github.com/posit-dev/py-shiny/pull/2420/commits/a84a2a304106b41296be84a495de4f96470b0717. The fast path now returns the canonical choice value (`choice_values[str(value)]`) instead of the entry unchanged, so the function matches its docstring.

For what it's worth, this was behaviorally a no-op: every caller routes the result through string coercion (`ChoiceSelection` compares `str(choice)`, and the `normalize_selected_*` payload builders call `str(v)`), and `str(value) in choice_values` guarantees the string forms agree. But returning the canonical value is cleaner and handles the pathological case of an object whose `__str__` collides with a key without comparing equal to the original. All 31 tests in `tests/pytest/test_input_check_radio.py` pass unchanged.

shiny/ui/_input_select.py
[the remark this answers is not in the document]
cpsievert: Yea, that seems right. https://github.com/posit-dev/py-shiny/pull/1947 was the PR that introduced it (simply to throw a deprecation warning).

shiny/ui/_choices.py
```diff
@@ -0,0 +1,197 @@
...
+
+from collections.abc import Iterable, Mapping
+from typing import Any, TypeVar, cast
+
+# A choice value.
+#
+# Deliberately `Any` rather than a union of the scalar types we expect. `Mapping`'s key
+# type is invariant, so a narrower annotation makes `Mapping[<narrow>, TagChild]` reject
+# `dict[int, str]` or `dict[str, str]`.
+ChoiceValue = Any
```
cpsievert: I'm wondering whether `Any` makes the public contract broader than intended. Since `SelectedArg` includes `ChoiceValue`, the public `selected` annotations now effectively accept any object. For example, `choices=[[1, 2]]` is accepted statically but fails at runtime because the choice is unhashable. This also appears to make `None` and arbitrary objects supported choice values.

R Shiny documents string, numeric, and logical choice values and converts those values, along with `selected`, to character values in `normalizeChoicesArgs()` and `choicesWithNames()`. Is similarly bounded support intended here, with `Mapping[Any, ...]` only addressing key-type invariance, or is support for any stringifiable Python object intentional?

shiny/ui/_choices.py
```diff
@@ -0,0 +1,197 @@
...
+    if selected is None or not choice_values:
+        return selected
+
+    originals = list(choice_values.values())
+
+    def resolve_one(value: Any) -> Any:
+        if str(value) in choice_values:
+            return choice_values[str(value)]
+        for original in originals:
+            if original == value:
```
cpsievert: The `1.0` versus `1` example explains why comparing only `str()` results may be insufficient in Python. I'm wondering about the wider consequences of using general Python equality, though. For example, `choices={1: "one"}, selected=True` selects `"one"` because `True == 1`. Equality may also return non-boolean objects or raise for array-like and custom types.

R Shiny converts both choice values and `selected` to character before comparing them. Consequently, `choices=c(one=1), selected=TRUE` does not select `"one"` because `"TRUE"` and `"1"` differ. What equivalence semantics do we want py-shiny to establish here?

cpsievert (commented):
Overall, this is a strong direction. Centralizing normalization keeps rendered options and update payloads consistent and closely follows R Shiny's character-coercion model. The radio handling and markup/payload tests are also solid. My inline questions are about defining edge cases, not the overall architecture.

shiny/ui/_input_select.py
```diff
@@ -331,12 +349,49 @@ def _update_options(
...
-        return {k: k for k in x}
-    else:
-        return x
+        return normalize_choices_mapping({k: k for k in x})
+
+    # The result may mix flat options and optgroups at the top level (e.g.
+    # `{"a": "A", "Group B": {...}}`). That matches neither arm of the
+    # `_SelectChoices` union, hence the `cast`, but `_render_choices()`
+    # checks each value with `isinstance`, so the mix is fine at runtime.
+    normalized = normalize_choices_mapping(x)
```
cpsievert: This appears to apply choice-value collision validation to every top-level key. When the associated value is a mapping, however, the key becomes an optgroup label rather than an option value. For example, `{0: {1: "one"}, "0": "flat"}` raises even though one `"0"` is an optgroup label and the other is an option value.

R Shiny treats optgroup names as labels and only the nested leaves as option values, so a group label and option value may have the same string. Are optgroup labels intended to participate in py-shiny's choice-value collision policy?

shiny/ui/_input_select.py
```diff
@@ -331,12 +349,49 @@ def _update_options(
...
+    """
+    if x is None:
+        return {}
+    elif isinstance(x, (list, tuple)):
+        return normalize_choices_indexed({k: k for k in x})[1]
+
+    index: dict[str, Any] = {}
+    for key, value in x.items():
+        if isinstance(value, Mapping):
+            index.update(normalize_choices_indexed(value)[1])
```
cpsievert: This silently replaces an earlier indexed value when two optgroups contain values with the same string form. For example, `{"A": {0: "zero"}, "B": {"0": "oh"}}` currently renders two `value="0"` options, and `selected=0` marks both with a `selected` attribute. In a single-select control, the browser ultimately selects the latter option. This differs from the new behavior within one mapping, where a stringification collision raises.

R Shiny permits duplicate underlying values, both within and across groups, although those values are inherently indistinguishable when reported by the browser. Is py-shiny intentionally adopting a stricter policy, and if so, is that policy meant to apply within each group or across the complete input?

shiny/ui/_choices.py
```diff
@@ -0,0 +1,197 @@
...
+from typing import Any, TypeVar, cast
+
+# A choice value.
+#
+# Deliberately `Any` rather than a union of the scalar types we expect. `Mapping`'s key
+# type is invariant, so a narrower annotation makes `Mapping[<narrow>, TagChild]` reject
+# `dict[int, str]` or `dict[str, str]`.
+ChoiceValue = Any
+
+_V = TypeVar("_V")
```
cpsievert: The repository convention says module-level names in private modules should not receive an additional underscore prefix because the module is already private. My reading is that this would apply to both `_V` and `_as_raw_list()`. Is there a reason these names should be treated differently here?

shiny/ui/_choices.py
jat255: Yes, the narrow set of types is what we want. `Any` was there only for the `Mapping` key position, and it spread to other annotations unintentionally.

This is fixed in 4436a04d, which added separate type vars for `ChoiceValue` and `ChoiceKey`. `ChoiceValue` is now `str | int | float | bool | None`, while mapping keys use a separate `ChoiceKey = Any`, and a comment explaining the decision. When `choices` is a list or a tuple, the annotation is now `Sequence[ChoiceValue]`. A `list` is invariant in its element type, so `list[ChoiceValue]` rejects a plain `list[str]`.

Your `choices=[[1, 2]]` example now fails a type check. Pyright rejects `choices=[[1, 2]]` for the checkbox group, the radio group, and the select inputs.

One hole stays, and it needs a runtime guard. A bare `str` is a `Sequence[str]`, so `choices="abc"` still passes a type check. Iterating it makes one choice per character. `_normalize_choices()` now raises `TypeError: "choices" must be a list, tuple, or dict.` for that case, instead of an `AttributeError` from a private helper. `None` and other non-iterables take the same path.

shiny/ui/_choices.py
jat255: Yes, I think you're correct that general Python equality is the wrong thing to use here and it was trying to get a bit too clever, IMO. It made `selected=True` name the choice value `1`, and a numpy or custom `__eq__` can raise or return a non-`bool` inside that loop.

This is fixed in 4436a04d. The string form is now the only identity, so the `==` fallback is gone. `selected=True` against `choices={1: "one"}` no longer checks the option, which matches the R Shiny implementation.

One exception was left in (`shiny/ui/_choices.py:115-118`, inside `resolve_one()`:) to prevent a regression/behavior change: On 1.7.0 and on `main`, `ui.input_checkbox_group(choices={1: "a", 2: "b"}, selected=[1.0])` renders the `1` option checked, via the raw `==` comparison.  This was implemented like so (this is pretty edge-casey, so happy to change this design, if you want):

```python
# `str(1.0)` is `"1.0"`, so an integral float misses the `1` it names. `bool` is
# a subclass of `int` rather than `float`, so `True` stays `"True"` here.
if isinstance(value, float) and value.is_integer():
    as_int = str(int(value))
    if as_int in choice_values:
        return as_int
```

`ui.input_radio_buttons(choices={1: "one"}, selected=True)` does check the option on 1.7.0 and on `main`, so this is a behavior change in that input type, which has a CHANGELOG entry.

shiny/ui/_input_select.py
jat255: The error enforces uniqueness of the _keys_ of `choices` after stringification, and those keys are a mix of choice values and optgroup labels. 

A select input's normalized `choices` is one `dict` keyed by the string form, and that dict holds optgroup labels next to flat choice values. A top-level collision drops an entry either way, so raising an error in the collision is the alternative to losing an option with no message.

You are right about R. `choicesWithNames()` returns a named list, which holds duplicates, so R renders both entries and needs no check.

I kept the error and made the message much more explicit (dbb226e2):

    Duplicate key '0' in `choices`: the optgroup label 0 and the choice value '0'
    are distinct but are identical as strings. A select input's `choices` (values
    and optgroup labels) must be unique when converted to strings.

Only the top level is affected. Each group's contents go through the same function, and those keys are all choice values, so `{"G": {0: "a", "0": "b"}}` still reports a duplicate choice value.

shiny/ui/_choices.py
jat255: Nope, that was an oversight (and just a habit of mine making all private methods prefixed with `_`). Both names drop the redundant prefix in 4436a04d: `_V` is now `V`, and `_as_raw_list()` is now `as_raw_list()`. The module is already private, so you're correct that the added underscore was superfluous.

shiny/ui/_input_select.py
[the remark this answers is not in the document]
jat255: Thanks for confirming, and for the pointer to the PR. `_contains_html()` was removed in 4436a04d. The `str` label type stays as #2049 left it, since the deprecation that `_contains_html()` served is long finished.

shiny/ui/_input_select.py
jat255: This method was removed in 4436a04d in the switch to string comparison. Its replacement, `_choice_value_strings()`, returns a `set[str]`, so nothing is overwritten across groups.

---

### More broadly...

Personally, I think silently dropping things because of the browser's limitations is odd behavior, and my expectation as a Python user would lean towards "this shouldn't be allowed, raise a loud error" rather than implicitly doing something I then have to notice. That is why I went the direction of raising. (I also appreciate this is a deviation from previous practice on the R side, so if you disagree, I'm happy to move to an approach more in line with R)

### Why there is a uniqueness rule at all

This is a consequence of where the string coercion happens. One of the key changes of this PR is that normalization happens once in `shiny/ui/_choices.py`. `main` keys the choices `dict` by the raw choice value and coerces to a string later at each point of use. This PR re-keys that `dict` by the string form once, so every consumer (`_render_choices()`, `ChoiceSelection`, `_choice_value_strings()`, `_find_first_option()`, the payload builders) reads an already-coerced form, making the rendered markup and the message `value` agree by construction.

The alternative is to keep the `dict` keyed by raw values, as `main` does, and coerce at the two boundaries instead: `value=str(k)` when rendering, and `str(v)` in the payload. Both still derive from the same `str()`, so markup and payload agree, duplicates render as they do in R, and no uniqueness rule is needed.

### Some specifics on the duplicate checking

The duplicate check runs on each `dict` in `choices` on its own, comparing only that dict's own keys: the outer dict once, then each optgroup's dict. That's where a collision loses a choice, because each dict is re-keyed to a new `dict` keyed by the string form. Values in two different optgroups are never siblings of one dict, so nothing is dropped and both options render.

Some examples, for clarity:

| `choices` value | Behavior on current `main` | after this PR |
|---|---|---|
| `{0: "zero", "0": "oh"}` | two `value="0"` options | `ValueError` (duplicate choice value) |
| `[0, "0"]` | two `value="0"` options | `ValueError` (duplicate choice value) |
| `{"G": {0: "a", "0": "b"}}` | two `value="0"` options in one group | `ValueError` (duplicate choice value) |
| `{0: {1: "one"}, "0": "flat"}` | one optgroup + one option, both keyed `"0"` | `ValueError` (label vs value) |
| `{0: {1: "one"}, "0": {2: "two"}}` | two `<optgroup label="0">` | `ValueError` (label vs label) |
| `{"A": {0: "zero"}, "B": {"0": "oh"}}` | both render, `selected=0` marks the first | both render, `selected=0` marks both |
| `{"A": {1: "one"}, "B": {1: "uno"}}` | both render, `selected=1` marks both | unchanged |
| `[0, False]` | one option, `value="0"`, label `False` | unchanged |

As I understand it, R Shiny permits duplicates in all of the above. So this PR is stricter than R most of the time, and closer to R in the two cross-group rows. The last row diverges from R in both columns, which is covered below.

Marking both options is mostly pre-existing. `main` already does it when two groups share a value at the same type, as in `{"A": {1: "one"}, "B": {1: "uno"}}`. Only mixed types change, and there `main` picked by raw type while both options render `value="0"` either way.

### Values that collide as `dict` keys still merge silently

`choices=[0, False]` names two choices, but renders one option: `value="0"` with the label `False`, and no error. `[1, True]` and `[1, 1.0]` behave the same way. In each case one choice disappears, and the option that survives takes its label from the entry that was dropped. This is because a list is turned into `{k: k for k in x}`, keyed by the raw value, and `0` and `False` are the same `dict` key in Python. Their string forms `"0"` and `"False"` are different, so keying by the string form instead keeps both options. That needs no new error, but I can see the argument for making it raise one just to be consistent, which we can fold into this work if wanted.

The same collision in a `dict` cannot be fixed. `{0: "zero", False: "oh"}` arrives here as `{0: "oh"}`, because Python merges the keys before the function is ever called, so there's nothing we can do about it at this point.

jat255:
@cpsievert thanks for your comments and sorry this got interrupted for a while. It should be ready for re-review. 

The main changes since your review: 
  - `resolve_selected()` matches `selected` to a choice by its string form only, so the general `==` fallback is gone. 
  - `ChoiceValue` is now `str | int | float | bool | None` instead of `Any`
  - the string-to-original index pattern is removed (which also removes the cross-optgroup overwrite)
  - the underscore-prefix and `_contains_html()` points are fixed.

I left two threads unresolved above about the optgroup collision-policy questions, in case you want to take a look. I'm comfortable with where things are at now, but happy to make any changes due to the slight divergence from R shiny behavior.

PR #2421, opened 2026-08-06
fix: Render data frames with an empty column name

### LINKED ISSUE #1844: Support for Empty Column in DataGrid

jat255:
Fixes #1844

Supersedes a stale draft in #1948 from last year.

## The issue

When the data frame has a column with the name `""`, a `DataGrid` shows nothing. The payload from the server is correct (it contains `"columns": ["", "b", ...]`). The fault is in the client side display. See [this example](https://shinylive.io/py/examples/#code=NobwRAdghgtgpmAXGKAHVA6VBPMAaMAYwHsIAXOcpMASxlWICcyACVKCAEygGcXe2nADoQAZo2IwWPABY0I2DHAAeqRnB586DZi3Vc4jPCwCuNESIgsAvCwBsIzqJuCMAEShkoAMUaw4ABQiLCEsIMGhkUIo0Yh6HADmgRAAlHgRkSHRAEax8RBJARDGVgBULABMaRmZ0YR5fgXJLOUVJS0sAMzVVplZYA2JzeWd7eUALD19-XCDTUUd42MsAKxToQC+IikWEGYYMhVBYB5eLL7+0TsQIgAC+pyGGNxeAPrilwbOMNivTgEpRA1dRkEyMKwPJ6nKAAcUYNE4ASc1zAGwAukA).

TanStack Table needs a non-empty string id for every column. `js/data-frame/index.tsx` did not set an explicit `id`, so TanStack then derived the id from the string `header`, which was the column name. An empty name gives an empty id, and TanStack throws an error inside `createColumn()`. This error stops the whole React render.

The position of the empty column does not matter. The fault happens for an empty name in any position in the dataframe.

## The fix

Column ids are explicitly set in a positional manner. The id of a column is `col-<index>`. The code never derives an id from a column name.

A column name is not a safe id. pandas accepts column names that are not strings, and these names reach the client as raw JSON values. A column with the name `0` gives the id `0`. In JavaScript `!0` is true, so TanStack rejects that id and the grid stays blank for the same reason as an empty name. Two different names can also share one string form, for example the integer `0` and the string `"0"`.

Names that are numbers also broke the translation from id to index. TanStack converts every id to a string, but the array of column names held the raw values. `indexOf` then returned `-1`, and the client sent `col: -1` to the server. User code that reads `data.columns[-1]` gets the last column, and no error appears.

Positional ids remove all of these faults at the same time. The client no longer looks up a column index by name. An id that does not resolve to a column gives `null`, and the client skips it instead of sending a wrong index. The doc comments on `columnIndexToId` and `columnIdToIndex` give the reasons.

The id must also work for server updates. schloerke noted in #1948 that a new id is safe only when the id stays in the browser. This constraint is not new, and this PR does not change any message shape. On current `main`, the server already exchanged column indices in `input.<id>_sort`, `input.<id>_filter`, `.update_sort()`, and `.update_filter()` methods. The change is on the client, where a positional id makes the translation from id to index exact.

### Sort and filter state across `update_data()`

Sorting and filtering state is keyed by column id, so positional ids on their own would key that state by column position. A user sorts a column by its name, and on `main` the state followed the name, because ids were names. `updateData` therefore migrates the state by name: it resolves each existing id to a name in the outgoing columns, looks that name up in the incoming columns, and re-keys the state onto the new position. State whose name is not in the incoming columns is dropped, as before. Behavior is unchanged from `main`, so a sort stays with its column across an `update_data()` that moves that column to another position.

Names are compared as they arrive rather than stringified, so a column named `0` and one named `"0"` stay distinct. narwhals rejects duplicate names server-side, so an outgoing name matches at most one incoming column.

## Changes outside the original scope

Three changes were not strictly necessary for #1844, but they are all in code that this PR must change anyway and I think enhance overall correctness:

- Dead store in `update_sort()`, in `shiny/render/_data_frame.py`. For a bare column index, the code computed `desc` from the column dtype and then replaced it with `True`. Every bare index sorted in descending order, whatever the dtype. The docstring promises that `desc` follows the dtype.

  *Why it is in this PR:* the test app drives `.update_sort()` on the empty-named column, and a new test covers a bare column index. That test has to assert one behavior or the other. A test that asserts the current behavior writes the fault into the test suite and makes it harder to correct later, so this is a change to make the code consistent with the documentation.

- New `DataFrameColumnDef` type, in `js/data-frame/types.ts`. `ColumnDef.id` and `ColumnDef.meta` are optional in TanStack, so `sort.ts` and `cell.tsx` used non-null assertions. If a column definition does not set `id`, the guard for HTML columns fills with `undefined`. The guard then does nothing and gives no error.

  *Why it is in this PR*: the guard in `sort.ts` read `col.header!`, because the id was the column name. The id is now positional, so this PR has to change that line to `col.id`. `ColumnDef.id` is optional, so the choice was a new non-null assertion or a type that requires the field. I picked the type. `cell.tsx` then lost its assertion at no extra cost.

- `PandasData.columns` is now typed `ColumnNames` (`ReadonlyArray<string | number>`), in `js/data-frame/types.ts`. It was `readonly string[]`, which is wrong: a non-string column name reaches the client as its raw JSON value.

  *Why it is in this PR*: this PR adds `header: String(colname)` so that a column named `0` does not render a blank label, and against a `string` type that coercion reads as redundant. The widening reaches three annotations that only forward the same array (`cell.tsx`, `data-update.tsx`, and the `updateData` parameter), and nothing else treats a column name as a string.

## Cases covered

The tests cover these cases:

- A column with the name `""` in the first, middle, and last position.
- Column names that are numbers: integer names in `pd.DataFrame({0: ..., 1: ...})`, and a name that is not an integer in `pd.DataFrame({1.5: ..., 2.5: ...})`. The grid renders, the labels are correct, and the sort works. The name `0` is the case that a falsy check mistakes for "no name". The two frames stay separate on purpose, because pandas unifies the column index dtype: one frame with both names retypes `0` as `0.0` and stops covering integer names at all.
- Sort and filter on these columns in both directions. This includes `input.<id>_sort`, `input.<id>_filter`, `.update_sort()`, and `.update_filter()`.
- Both branches of the dtype default in `.update_sort(<int>)`. Number-like columns sort in descending order. Other columns sort in ascending order.
- Sort and filter state across `update_data()`. The same two columns in the opposite order keeps both on the named column and reports its new index to the server. Columns that share no name with the old ones drops both, instead of reapplying them to whatever now sits at that position.

## Cases not covered

- Duplicate column names, and two empty names. narwhals raises `DuplicateError`. This restriction applies to all duplicate names. It is not specific to empty names.
- Column names that are `None` or `NaN`. pandas turns a `None` key into `nan`, and `serialize_frame()` accepts it. The message to the client then fails in transport, because `json.dumps()` writes a bare `NaN` literal and `JSON.parse` rejects it. This fault is in the JSON layer, and the column id never takes part in it. A correction needs a change to serialization on the server, so I left this case out of scope.

A name of only whitespace (`" "`) worked before this PR. This is why the issue thread gave `" "` as a workaround. It still works.

No bundled Agent Skill in `shiny/.agents/skills/` documents restrictions on column names. No skill needed an update.

## Tests

- `tests/playwright/shiny/bugs/1844-datagrid-empty-column-name/` is the regression guard. It has four grids. The first grid has the empty name at index 0. The second grid has the empty name at index 1, so an implementation that handles only the first column fails. The third grid has the integer names `0` and `1`. The fourth grid has the non-integer names `1.5` and `2.5`.
- `tests/pytest/test_render_data_frame_tbl_data.py::test_serialize_frame_empty_column_name` uses three columns, and covers the first, middle, and last position.
- `tests/playwright/shiny/components/data_frame/update_data/test_update_data.py::test_update_data_keeps_sort_and_filter_on_named_column` pins the state migration. Its app gained a "Reorder columns" button and readback outputs for `.sort()` and `.filter()`.


CHANGELOG.md
```diff
@@ -37,6 +37,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Download renderers (`@render.download_button`, `@render.download_link`, and the deprecated `@render.download`) now honor `@output(id=)`. The download handler was registered under the decorated function's name, but the URL rendered by the control used the `@output(id=)` value, so clicking the control returned a 404. (#2415)
 
+* `@render.data_frame` is now able to render data frames whose column names are empty (`""`) or are not strings. Column ids are now positional and are never derived from the column name. A column named `0` also no longer renders a blank header. (#2421)
```
schloerke, line 46: The behavior change you flag in the PR description -- sort/filter state keyed by column **position** instead of column **name**, so a sort now stays on the same position across `update_data()` rather than following the name -- isn't mentioned in either CHANGELOG entry, and no test covers it.

I agree with the choice: every message to the server already carries indices, so the client was the inconsistent side. But it is user-visible for anyone whose `update_data()` reorders columns, so I'd like it recorded rather than left implicit. Could you add a line here, plus a case in `tests/playwright/shiny/components/data_frame/update_data/` pinning the new semantics? That's the only thing I'd ask for before merge.

js/data-frame/index.tsx
```diff
@@ -707,8 +752,12 @@ const ShinyDataGrid: FC<ShinyDataGridProps<unknown>> = ({
   useEffect(() => {
     if (!id) return;
     const shinySort: { col: number; desc: boolean }[] = [];
-    sorting.map((sortObj) => {
-      const columnNum = columns.indexOf(sortObj.id);
+    sorting.forEach((sortObj) => {
+      const columnNum = columnIdToIndex(sortObj.id, columns.length);
+      // Skip state left over from columns that no longer exist, rather than
+      // reporting a column index the server cannot resolve.
```
schloerke: I think this branch is unreachable in practice. `updateData` (above, ~L368) already filters `sorting` and `columnFilters` against the rebuilt `newTypeHintMap` and calls `setSorting`/`setColumnFilters` with the survivors, so by the time this effect runs a stale id should already be gone. The comment (and the doc comment on `columnIdToIndex`) states the scenario as live, which will send the next reader looking for a path that isn't there.

Not asking you to drop the guard -- returning `null` beats the old `indexOf` returning `-1` and shipping `col: -1` to the server, and it's cheap insurance. Just either soften the wording to "defensive: sorting state should already have been filtered by `updateData`, but never report an index the server cannot resolve", or add a case that actually exercises it (shrink the frame via `update_data()` while a sort sits on a trailing column). Same note applies to the matching branch in the filter effect below.

js/data-frame/index.tsx
```diff
@@ -199,13 +239,17 @@ const ShinyDataGrid: FC<ShinyDataGridProps<unknown>> = ({
...
             return row[colIndex];
           },
           // TODO: delegate this decision to something in filter.tsx
           filterFn:
             typeHint?.type === "numeric" ? "inNumberRange" : "includesString",
-          header: colname,
+          // Coerced to a string because TanStack's `flexRender()` renders
+          // nothing for a falsy header, which would blank out the label of a
+          // column whose (non-string) name is `0`.
+          header: String(colname),
```
schloerke, line 260: This coercion is necessary precisely because the server sends non-strings -- but `PandasData.columns` is still typed `readonly string[]` in `types.ts`, so a reader who trusts the type will conclude this `String()` is redundant and delete it.

Widening to `readonly (string | number)[]` would make the coercion visibly required and stop the next person from assuming `string` elsewhere. (It would surface anywhere else that treats a column name as a string, which is worth knowing about.) Optional, and fine as a follow-up if it ripples further than expected.

shiny/render/_data_frame.py
```diff
@@ -1138,12 +1138,19 @@ async def update_sort(
...
-                val_dict: ColumnSort = (
-                    val if isinstance(val, dict) else {"col": val, "desc": True}
-                )
+                # Each entry is either a full `ColumnSort` dict or a bare
+                # column index to be upgraded per the docstring default.
+                if isinstance(val, dict):
+                    val_dict = val
+                else:
+                    assert isinstance(val, int)
+                    assert 0 <= val < ncol
```
schloerke, line 1151: This duplicates the `assert 0 <= val_dict["col"] < ncol` a few lines below, but it is doing real work: without it a negative index would silently wrap in `nw_data[:, val]` and compute `desc` from the wrong column before the later assert fires. Worth a short comment saying so, otherwise it reads as redundant and invites deletion.

On the fix itself: agreed, and your justification holds -- a test had to assert one behavior or the other, and encoding the dead store into the suite would have been worse. Flagging for the record that it is user-visible beyond #1844: `update_sort(<int>)` on a string column now sorts ascending where it previously always sorted descending. The docstring already promised dtype-based `desc`, so this is a fix, and the CHANGELOG entry below covers it.

schloerke (commented):
Reviewed the full diff and verified it locally against the PR head:

| Check | Result |
| --- | --- |
| `tests/pytest` (full) | 1070 passed, 2 skipped |
| New `tests/playwright/shiny/bugs/1844-datagrid-empty-column-name/` | 5 passed |
| `tests/playwright/shiny/components/data_frame` (all 24 apps) | 42 passed, no regressions |
| `uv run make check-lint check-types` + `make check-pyright` | clean (0 errors) |
| Checked-in `shiny/www/py-shiny/data-frame/data-frame.js` | in sync with source |

On that last row: no CI job rebuilds the JS bundle, so I checked the minified asset by hand. It contains `var Ln="col-"`, `id:It(S)`, `header:String(v)`, and the `startsWith`/`slice` id parser, so the committed bundle matches the final TypeScript.

The approach is right, and it answers the open question from the review on #1948 directly: the TanStack id never leaves the browser, so no new server-side restriction on column names is needed. Nothing here is blocking. Four inline notes, only the first of which I'd want addressed before merge.

schloerke:
## How this compares to #1948

Since this supersedes #1948, here's a side-by-side, mostly so the record shows why the swap is worth it and that the review question left on #1948 is now answered.

| | #1948 (WIP) | this PR |
| --- | --- | --- |
| Approach | Keep name-derived ids; synthesize an id for the empty-named column by appending spaces (`findUnusedColumnName`, tibble-inspired) | Never derive ids from names; ids are positional `col-<index>` |
| id → server index | still `columns.indexOf(id)`; the synthetic id `" "` is not in `columns`, so `indexOf` returns `-1` and **`col: -1` goes to the server**. `data.columns[-1]` then quietly resolves to the last column | exact by construction; a stale or unparseable id is skipped rather than sent |
| More than one empty name | breaks (the uniquifier only handles one) | n/a — narwhals rejects duplicate names regardless |
| Non-string names (`0`, `1.5`) | still broken: `0` is falsy so TanStack rejects the derived id, and `indexOf` can't match a raw JSON number against a stringified id | fixed, including `header: String(colname)` so a column named `0` doesn't render a blank label |
| Tests | 1 app, 1 test (grid renders) | 4 grids, 5 tests: full client→server and server→client sort/filter round trips with the empty name at index 0 *and* index 1, plus integer and non-integer name frames, plus a `serialize_frame` unit test |

The review comment on #1948 was:

> If the tanstack `id` is only browser specific, then this is great. If the id is used for updates to the server (ex column sorting input values), then we should raise an error server-side before it is even sent to the browser as the column names will not match.
>
> Narwhals has a restriction of non-duplicated column names. It's fair if we add non-empty column names for tanstack.

This PR takes the first branch of that, and I'm satisfied it earns it. The id is strictly browser-local: no message shape changes, and the id → index translation is now exact by construction rather than a name lookup that can miss. So the server-side error isn't needed, and we don't have to impose a non-empty-name restriction — which is the better outcome for #1844, since empty column names are a real thing people have in their data (hence the `" "` workaround in that thread, which still works).

The one thing I'd flag as genuinely new scope versus #1948 is that keying sort/filter state by position rather than name changes behavior across `update_data()`. I've left an inline note asking for a CHANGELOG line and a test, not a revert.

Once this merges, #1948 should be closed pointing here, with credit to @ErdaradunGaztea — that PR did the diagnosis (TanStack rejecting an empty id) that this one builds on.

js/data-frame/index.tsx
```diff
@@ -718,15 +767,19 @@ const ShinyDataGrid: FC<ShinyDataGridProps<unknown>> = ({
 
     // Deprecated as of 2024-05-21
     window.Shiny.setInputValue!(`${id}_column_sort`, shinySort);
-  }, [columns, id, sorting]);
+  }, [columns.length, id, sorting]);
```
schloerke, line 810: This threw me off as "length is not a value for comparison".

... but it is. Ex: `4` and the next time it'll be `4`. So nothing new.

CHANGELOG.md
jat255: On second look, I didn't care for this change of behavior just to fix this bug, since a user's sorting is using the semantic column names, and it would be unexpected, I think, to have the sorting attached to column position. The latest batch of changes makes `updateData` now migrate sorting and filtering state by column *name*, so there is no behavior change to note in the changelog. It resolves each existing id to a name in the outgoing columns, looks that name up in the incoming columns, and re-keys the state onto the new position, dropping state whose name is gone. Ids stay positional, which fixes the original bug, but a sort keeps following its column across an `update_data()` that moves it.

Names are compared as they arrive rather than stringified, so a column named `0` and one named `"0"` stay distinct, and the narwhals prohibition on duplicates protects us here.

Added tests for this in `test_update_data_keeps_sort_and_filter_on_named_column` in `tests/playwright/shiny/components/data_frame/update_data/`.

One incidental difference: sorting state now survives on the presence of the column name, where before it survived on the presence of a type hint for that column id. `typeHints` is `Required` in `SerializedFrame` and always set by `serialize_frame`, so that path never differed in practice. Filters still require a type hint, since the decision to keep or drop depends on it.

js/data-frame/index.tsx
jat255: Good catch. I softened the wording softened to say so in both the sorting and the filter effect, and in the doc comment on `columnIdToIndex`. The guard itself stays as defensive code, for the reason you gave: `null` and skip beats a wrong-but-valid index reaching user code.

js/data-frame/index.tsx
jat255: Done, as a named `ColumnNames = ReadonlyArray<string | number>` in `types.ts` with the justification included in the comment on the type.

shiny/render/_data_frame.py
jat255: Added a comment saying that the index must be rejected before it reaches `nw_data[:, val]`, where a negative value would wrap and compute `desc` from the wrong column.

cpsievert:
Thanks for the review @schloerke. Consider this approved on my side once @schloerke approves.

PR #2426, opened 2026-08-07
fix(playwright): Close the selectize dropdown without clicking the app

karangattu:
## Summary

`InputSelectize.expect_choices()` opens the selectize dropdown, because selectize renders its choices into the DOM only after the first open. `expect_choice_labels()` and `expect_choice_groups()` do the same.

The controller then closed the dropdown with a click on the page body. That click lands on app content and fires the click handlers of the app under test. A test could therefore record an interaction that it never made.

The controller now calls `close()` on the selectize instance. This call touches no app content.

## How the dropdown closes

The three close sites share one private helper, `_close_dropdown()`. The helper calls `close()` and `blur()` on the selectize instance.

`_populate_dom()` waits 50 ms between the open and the close. Selectize answers the opening click across several tasks: it queues a focus, and that focus opens the dropdown again. The dropdown reports `display: block` before this work finishes, so a close that runs first gets undone. The wait lets the work settle.

## Rejected approaches

Two other close methods failed in CI. Both left the dropdown open, and the next assertion then waited 5 seconds for `display: none`.

| Approach | Result |
| --- | --- |
| Press `Escape` on the selectize input | Failed on the first open. Selectize binds its keydown handler to the inner `input` element, so the key only closes the dropdown when that element holds DOM focus. The CI browser does not give it focus. |
| Call `close()`, then close again after one tick | Failed on the second open of the same control. The pending work is not a fixed number of tasks, so one tick is not enough. |

Both approaches passed on a local macOS machine. That machine does not reproduce this fault with any close method, so a local pass is not evidence here.

## Changes

- The three close sites share one private helper, `_close_dropdown()`.
- The helper passes `timeout` to the call. The previous code dropped this value.
- The 50 ms wait now carries a comment that states what it waits for.
- A new test, `test_expect_choices_does_not_click_app_background`, counts app clicks during `expect_choices()`.
- Two `expect_choices()` assertions in the selectize kitchensink test are active again. They carried a `TODO-karan` comment that reported a hang.
- The docstrings of `set()` and `_populate_dom()` describe the current behavior.
- The comment on the item transition returns to its original text, which was correct.

## Tests

The close behavior is the same as in 91f9c46, which passed a full CI run. The difference from that commit is the helper method, the passthrough of `timeout`, the comments, and the docstrings.

`make format check-lint check-types` reports no new errors.


schloerke (approved):
Thank you!

PR #2446, opened 2026-08-13
chore: upgrade html deps (bslib 0.11.9000 → 0.12.0.9000)

gadenbuie:
## Summary
* Vendors bslib 0.12.0.9000 (from 0.11.0.9000) via `make upgrade-html-deps`, along with the latest shiny/htmltools HTML dependency assets.
* Includes a new `tooltip-popover.scss` rule (upstream bslib #1326 / shiny #4388 workaround): a dynamically-rendered output inside a title-less `popover()`/`tooltip()` no longer collapses to 0 width and gets stuck "recalculating".

## Test plan
- [x] `uv run make format check-lint check-types`
- [x] Added `tests/playwright/shiny/bugs/4388-popover-tooltip-output-collapse/` — verifies a `@render.ui` output inside a title-less popover/tooltip renders and that the `.shiny-html-output` min-width CSS fix is applied. Confirmed the test fails without the vendored CSS rule and passes with it.


schloerke (approved):
Thank you!

PR #2450, opened 2026-08-14
fix: show_offcanvas() accepts a string id and tag content

### LINKED ISSUE #2445: feat(ui): `show_offcanvas()` should accept an id, or fail with an actionable error
## Summary

`ui.show_offcanvas()` takes an `Offcanvas` object. Passing it the panel's `id` — the natural guess, given that its two siblings are id-keyed — fails at click time with an unhandled `AttributeError` from deep inside the implementation.

Found while writing the Offcanvas docs page for py-shiny-site (posit-dev/py-shiny-site#422, PR posit-dev/py-shiny-site#446); a reviewer's first instinct on reading the finished example was "why isn't this `show_offcanvas("panel")`?", which is what prompted this.

## Current API

| Function | Signature |
|---|---|
| `show_offcanvas` | `(offcanvas: Offcanvas, *, session=None)` |
| `hide_offcanvas` | `(id: Union[str, Offcanvas], *, session=None)` |
| `toggle_offcanvas` | `(id: str, show: Optional[bool] = None, *, session=None)` |

So within one trio: `hide` accepts `str | Offcanvas`, `toggle` accepts only `str`, and `show` accepts only `Offcanvas`.

## Reproduce

```python
from shiny import App, Inputs, Outputs, Session, reactive, ui

app_ui = ui.page_fluid(
    ui.offcanvas(ui.p("Body."), title="Server controlled", id="panel"),
    ui.input_action_button("show", "Show"),
)

    @reactive.effect
    @reactive.event(input.show)
    def _():
        ui.show_offcanvas("panel")

app = App(app_ui, server=server)
```

Click **Show**:

```
File ".../shiny/ui/_offcanvas.py", line 400, in show_offcanvas
    local_id = offcanvas.id if offcanvas.id is not None else rand_hex(8)
AttributeError: 'str' object has no attribute 'id'
```

The app does not crash, but the effect dies and nothing happens — silent from the user's point of view apart from the console. Tested on `main` @ f85a4446 (1.7.1.dev6).

The working form is `ui.toggle_offcanvas("panel", show=True)`.

## The complication (please weigh in)

The obvious fix — "accept `str` as the id" — collides with the precedent next door:

```python
show_toast(toast: str | Toast | Tag | TagList, ...)   # a str is the toast's CONTENT
hide_toast(id: str | Toast, ...)                      # a str is an ID
```

`show_toast("hello")` builds a toast whose body is "hello". If `show_offcanvas` accepted `str` as an *id*, the two `show_*` functions would read the same and mean opposite things. That is arguably worse than the current situation, so this may not be the right fix.

## Options

1. **`show_offcanvas(id: str | Offcanvas)`, where `str` is an id.** Most useful (it's what people try), consistent with `hide_offcanvas`, inconsistent with `show_toast`. An offcanvas has no natural content-only shorthand — a panel wants a `title` or `aria-label` regardless — so the `show_toast` reading is less tempting here.
2. **Keep the type, fix the error.** Raise `TypeError("show_offcanvas() takes an offcanvas() object; to open a panel already in your UI by id, use toggle_offcanvas(id, show=True)")`. Cheap, no API-design commitment, and turns a confusing `AttributeError` into a signpost. Worth doing even if option 1 lands.
3. Both: accept `str` as id *and* keep a clear error for other wrong types.

Independently and regardless of the above: `toggle_offcanvas(id: str)` could accept `str | Offcanvas` to match `hide_offcanvas`.

I lean toward **2 now, 1 if you're happy with the `show_toast` divergence** — but this is your call on API consistency, and I'm happy to send a PR for whichever you pick.

## Meanwhile

py-shiny-site documents the working form and the trap explicitly, so the docs are not blocked on this:

> …opening a panel by `id` is `ui.toggle_offcanvas(id, show=True)`, not `ui.show_offcanvas()` — that one takes the object returned by `ui.offcanvas()`, and passing it an `id` string raises `AttributeError`.

gadenbuie:
Fixes #2445

## Summary

`ui.show_offcanvas()` now dispatches on the type of its first argument, matching bslib's `show_offcanvas()` (rstudio/bslib#1346):

- a string `id` of an `offcanvas()` panel already in the UI → reveals it (same as `toggle_offcanvas(id, show=True)`)
- tag content (e.g. `ui.p(...)`, `htmltools.HTML(...)`, `ui.markdown(...)`) → wrapped into a new anonymous panel
- an `Offcanvas` object → rendered + shown (unchanged)

Previously, `show_offcanvas()` only accepted an `Offcanvas` object; passing a string id raised an unhandled `AttributeError` deep inside the implementation (#2445). A string that looks like body text instead of an id (empty, or containing whitespace) now raises an actionable `ValueError` instead of being silently misinterpreted.

## Examples

Revealing an existing panel by id:

```python
from shiny import App, Inputs, Outputs, Session, reactive, ui

app_ui = ui.page_fluid(
    ui.offcanvas(ui.p("Body."), title="Settings", id="settings"),
    ui.input_action_button("show", "Show settings"),
)

    @reactive.effect
    @reactive.event(input.show)
    def _():
        ui.show_offcanvas("settings")

app = App(app_ui, server=server)
```

Showing server-rendered Markdown content in a new anonymous panel:

```python
from shiny import App, Inputs, Outputs, Session, reactive, ui

app_ui = ui.page_fluid(
    ui.input_action_button("show_notes", "Show release notes"),
)

    @reactive.effect
    @reactive.event(input.show_notes)
    def _():
        ui.show_offcanvas(
            ui.markdown(
                """
                ## Release notes

                - Added **markdown** support in server-rendered panels.
                - See the [docs](https://shiny.posit.co) for details.
                """
            )
        )

app = App(app_ui, server=server)
```

## Implementation notes

- Related bslib PR: rstudio/bslib#1346. bslib assets are already up to date in py-shiny (#2446), so this is a pure Python port with no client-side changes.
- The string-id path delegates to the existing `toggle_offcanvas(id, show=True, session=session)` rather than duplicating the send logic.
- `hide_offcanvas()` / `toggle_offcanvas()` are unchanged; `toggle_offcanvas` accepting an `Offcanvas` object too (for full parity with `hide_offcanvas`) was raised in #2445 as a separate, independent question and is out of scope here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug fixes
 
+* `ui.show_offcanvas()` now accepts the `id` (a string) of an `ui.offcanvas()` panel already in the UI and reveals it, matching its sibling functions `ui.hide_offcanvas()` / `ui.toggle_offcanvas()`. Previously, passing a string raised an unhandled `AttributeError` from deep inside the implementation. `show_offcanvas()` also now accepts bare tag content (wrapping it into a new anonymous panel), in addition to an `ui.offcanvas()` object; a string that looks like body text instead of an id (empty, or containing whitespace) raises an actionable `ValueError`. (#2445)
+
 * A dynamically-rendered output (e.g. `@render.ui`) inside a `ui.popover()` or `ui.tooltip()` without a `title=` no longer gets stuck showing "recalculating". The container collapsed to 0 width, so the output's `ResizeObserver` never fired; vendored bslib CSS now gives it a non-zero minimum width. (#2446)
 
 * `playwright.controller.InputSelectize` no longer clicks the page body to close the selectize dropdown. `expect_choices()`, `expect_choice_labels()`, and `expect_choice_groups()` open the dropdown, because selectize renders its choices into the DOM only after the first open. The click that closed the dropdown again landed on app content and fired the app's own click handlers, so a test could record an interaction that it never made. The controller now calls `close()` on the selectize instance instead, which touches no app content. (#2426)
--- a/shiny/api-examples/show_offcanvas/app-core.py
+++ b/shiny/api-examples/show_offcanvas/app-core.py
@@ -1,14 +1,28 @@
 from shiny import App, Inputs, Outputs, Session, reactive, ui
 
 app_ui = ui.page_fluid(
-    ui.input_action_button("show_btn", "Show panel"),
+    ui.offcanvas(
+        ui.p("This panel is declared in the UI."),
+        title="Existing panel",
+        id="existing_panel",
+    ),
+    ui.input_action_button("show_existing_btn", "Show existing panel"),
+    ui.input_action_button("show_server_btn", "Show server panel"),
+    ui.input_action_button("show_markdown_btn", "Show markdown content"),
 )
 
 
 def server(input: Inputs, output: Outputs, session: Session):
     @reactive.effect
-    @reactive.event(input.show_btn)
+    @reactive.event(input.show_existing_btn)
     def _():
+        # Reveal a panel already declared in the UI, by its id.
+        ui.show_offcanvas("existing_panel")
+
+    @reactive.effect
+    @reactive.event(input.show_server_btn)
+    def _():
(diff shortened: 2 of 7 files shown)


schloerke (approved):
Thank you!

PR #2432, opened 2026-08-10
fix(playwright): stabilize flaky browser tests

karangattu:
## Summary

- Wait for the final tooltip state before the next action.
- Reacquire data-frame cells after a reactive update replaces them.
- Handle an overlay that closes before Playwright reads its state.
- Remove the extra `about:blank` navigation for each shared page.
- Retry one transient resource error and ignore failed Google Fonts requests.
- Use a TCP readiness probe that does not close a remote browser.
- Copy each external template into a separate temporary directory.
- Remove shinyapps.io deployment tests, credentials, and deployment records.
- Keep local and Posit Connect deployment tests.

## Root causes

The tests sometimes observed intermediate states during tooltip updates and data-frame renders. Reactive updates replaced elements that the tests still referenced.

An overlay can close while Playwright reads its state. The controller now treats the closed overlay as hidden.

The readiness probe opened and closed a remote browser. Concurrent template tests also shared one DuckDB file.

Static resources can briefly return HTTP 404 during startup. Google Fonts can also fail for reasons outside the application.

A retried shinyapps.io deployment can conflict with an earlier deployment task. The platform tests will not return, so this PR removes them.

## Impact

These changes remove browser-test flakes without increasing global timeouts.

Deployment tests now target local applications and Posit Connect only. CI no longer reads shinyapps.io credentials.

## Tests

- `uv run --extra dev --extra test pytest -c tests/playwright/playwright-pytest.ini tests/playwright/deploys --browser chromium -q` — 18 passed, 18 skipped.
- `uv run --extra test make check-tests` — 1,055 passed, 3 skipped.
- `uv run --extra dev --extra test make check-format check-lint`
- `uv run pyrefly check tests/playwright/utils/deploy_utils.py`

The Connect tests skipped because this machine does not have deployment credentials.

The full type check still reports three errors in files that this change does not modify.

Closes #2433
Closes #2431
Closes #2430
Closes #2427
Closes #2424
Closes #2414
Closes #2390

PR #2452, opened 2026-08-15
test(playwright): Make pytest-playwright's --tracing flag work

jat255:
## What was broken

`tests/playwright/conftest.py` overrides pytest-playwright's `page` fixture with a session-shared page built from `browser.new_page()`. `browser` is a `Browser`, not a `BrowserContext`, so `new_page()` creates an implicit context with default options. pytest-playwright records `--tracing`, `--video`, and `--screenshot` artifacts only in its own function-scoped `context` fixture, which this suite never instantiates. The flags parsed successfully and silently wrote nothing, which is easy to mistake for "the run produced no failures worth tracing".

## The change

A session-scoped `_session_context` fixture now creates the context via `browser.new_context(**browser_context_args)` and owns the shared page, and it calls `context.tracing.start(...)` when `--tracing` is on or retain-on-failure.

Tracing normally starts and stops per test, but the whole point of this suite's `page` fixture is that the page (and therefore its context) outlives every individual test. `tracing.start_chunk()` / `stop_chunk()` is the mechanism for that case: an autouse function-scoped `_trace_chunk` fixture opens a chunk per test and writes it out, slicing one long-lived trace into a file per test without giving up the shared page. A `pytest_runtest_makereport` hookwrapper records the outcome on the item so retain-on-failure can tell a pass from a failure.

Traces are written to `<output>/<slug>/trace.zip` rather than a flat `<slug>.trace.zip`. That matches pytest-playwright's own artifact layout and the `test-results/*/trace.zip` location `make playwright-show-trace` already looked in; a flat layout would not.

## Using it

```bash
# One trace per test, written to test-results/<test>/trace.zip
make playwright-shiny SUB_FILE="inputs/test_foo.py" PYTEST_EXTRA_ARGS="--tracing on"

# Open the newest trace, or a specific one with TRACE="path/to/trace.zip"
make playwright-show-trace
```

`PYTEST_EXTRA_ARGS` is new: the `playwright` targets had no way to pass a one-off pytest option. Use `--tracing retain-on-failure` to keep traces only for failures. Tracing does not need `--headed`, so the default headless runs record fine: a trace's screenshots and DOM snapshots come from the renderer, which headless still rasterizes.

`make playwright-show-trace` also changes. It passed the whole `test-results/*/trace.zip` glob to `playwright show-trace`, which accepts exactly one trace and exits with "too many arguments" given more. That never came up while the flag recorded nothing, but one trace per test makes it the normal case, so the target now opens the newest trace and takes `TRACE=` to pick another.

Both commands are documented in CLAUDE.md's testing shortcuts and in the testing section of `.claude/references/architecture.md`.

## Caveats

`--video` now records, but as one continuous video for the whole session rather than one per test. Video is a context-level option and there is no video equivalent of `start_chunk()`, so per-test video would mean giving up the shared page, which is what makes this suite fast. `--screenshot` remains a no-op: the plugin captures screenshots when tearing down its own `page` fixture, which this conftest replaces. Both are documented in the NOTE in `tests/playwright/playwright-pytest.ini`, replacing the now-stale claim that all three options do nothing.

No CHANGELOG entry: this is developer-facing test infrastructure, matching how recent commits in this area were handled (#2432, #2306, #2295).

## Verification

`uv run make format check-lint check-types` is clean, and so is `uv run make check-pyright`. Both are needed here: `check-types` runs only pyrefly, while CI runs pyright as its own job, and pyright's strict mode rejected the first version of these fixtures (`FixtureRequest.node` is unannotated in pytest, `getoption("--output")` is `Any | None`, and a `hookwrapper=True` generator is sent pluggy's `Result`, not `None`).

`--tracing on` over `tests/playwright/shiny/components/data_frame/` (42 tests) produced exactly 42 `trace.zip` files, one per test, each containing `trace.trace`, `trace.network`, and screenshot resources. Rerunning under `-n 4` produced the same 42 with no filename collisions, and a serial (`-n 0`) run behaves the same. `--tracing retain-on-failure` over a scratch pass/fail pair wrote a trace only for the failing test. With the default `--output`, the resulting paths match the `test-results/*/trace.zip` glob (checked directly rather than by launching the interactive viewer).

Through the make targets: `make playwright-shiny ... PYTEST_EXTRA_ARGS="--tracing on"` wrote one trace per test under `test-results/`, and `make playwright-show-trace` resolved to exactly one trace both by default (newest) and via `TRACE=`.

The default path is unregressed: `tests/playwright/shiny/components/data_frame/` plus `tests/playwright/shiny/bugs/` (62 tests) passed on chromium with no tracing flags and wrote no artifacts.


tests/playwright/conftest.py
```diff
@@ -91,6 +96,102 @@ def goto_and_verify_commit(url: str, **kwargs: typing.Any) -> Response | None:
...
+def pytest_runtest_makereport(
+    item: pytest.Item, call: pytest.CallInfo[None]
+) -> typing.Generator[None, typing.Any, None]:
+    """Record the outcome so `_trace_chunk` can honor `retain-on-failure`."""
+    # A `hookwrapper=True` generator is sent pluggy's `Result` object, which is
+    # why the generator's send type is `Any`: pluggy is not a direct dependency
+    # here, so `Result` cannot be named. The report it wraps is annotated below.
+    outcome = yield
+    report: pytest.TestReport = outcome.get_result()
+    if report.when == "call" and report.failed:
```
karangattu: Could we also keep the trace when setup fails? Right now we only mark failures that happen during the test itself, so if a fixture fails during setup, the trace gets thrown away.

jat255: Good idea. This is fixed in ec0ce774. 

`pytest_runtest_makereport` now marks the item on any failed setup or call phase. pytest reports the setup phase before it runs finalizers.

Teardown is deliberately left out rather than just checking `report.failed`: the teardown report is only produced once every finalizer has run, so by then the chunk is already saved or discarded and including it would do nothing.

One case this approach doesn't cover: a session-scoped fixture that fails before `_trace_chunk` is first instantiated leaves no chunk to save at all.

tests/playwright/conftest.py
```diff
@@ -91,6 +96,109 @@ def goto_and_verify_commit(url: str, **kwargs: typing.Any) -> Response | None:
...
+
+    The shared page must come from a context we control, not from
+    `browser.new_page()`. `new_page()` creates an implicit context with default
+    options, which is why pytest-playwright's `--tracing`, `--video`, and
+    `--screenshot` flags used to have no effect on this suite: those artifacts
+    are only recorded by the plugin's own (function-scoped) `context` fixture,
+    which this suite never uses.
+    """
+    context_args = dict(browser_context_args)
+    if pytestconfig.getoption("--video") in ("on", "retain-on-failure"):
```
karangattu: Would `--video retain-on-failure` actually behave differently from `--video` on here? Since the browser context is shared for the whole session, it looks like the video would still be kept even if all the tests pass.
That's the only thing that looks off

karangattu (approved):
LGTM

tests/playwright/conftest.py
jat255: Ah, good catch. 

`pytest-playwright` gets the "retain" half of the mode from its `ArtifactsRecorder`, which records into a temp dir and copies out only failed tests' videos. That runs off its function-scoped `context`/`page` fixtures, which this conftest replaces, so we were setting `record_video_dir` and never deleting anything, meaning `retain-on-failure` behaved exactly the same as `on`.

Added some cleanup code in 74f6e8ef to that flag so it will delete the videos if there were no failures. We now collect the page videos as they're created and delete them after the context closes if no test failed.

PR #2455, opened 2026-08-18
refactor(data frame): Drop the unused `columns` prop from the cell edit path

schloerke:
Follow-up from reviewing #2421. Stacked on it because it touches the same lines.

`TableBodyCell` took a `columns` prop and forwarded it to `updateCellsData()`, but neither read it: the cell edit path identifies its target entirely by `rowIndex`/`columnIndex`, and the patch payload sent to the server is `{row_index, column_index, value}`. The prop was also a dependency of the save `useCallback`, so the callback was recreated whenever the columns array identity changed.

Also drops dependencies the surrounding effects do not read: `columns` from the `addPatches` and `updateData` listener effects, `setSorting` from the `addPatches` effect, and `resetCellEditMap` and `setTableData` from the `updateData` effect, whose body only calls `updateData(evtData)` and so reaches them through that callback's own closure.

**This does not make anything re-run less often**, and an earlier version of this description claimed otherwise — thanks @jat255 for catching it. `useCellEditMap()` returns a fresh `setCellEditMapAtLoc` and `resetCellEditMap` on every render, so those deps already invalidate on every render, which is a superset of every columns change. Making the re-runs less frequent means stabilizing those identities inside `useCellEditMap()` — a separate change, happy to do it as a follow-up. This PR stands on the deps being dead, not on a re-render win.

## Why this is worth doing

#2421 establishes that column *names* are not usable identifiers — they can be empty, and pandas allows names that are not strings. Auditing the client for other sites that carry names turned up exactly two, and both were dead. With them gone, the client reads a column name in exactly two places, both deliberate: the header label (`header: String(columnName)`) and the sort/filter migration in `updateData()` that #2421 added, which has to match by name to follow a column across a reorder.

Dropping the prop makes the `ColumnNames` import unused in both `cell.tsx` and `data-update.tsx`, so it goes too.

No behavior change.

## Verification

Rebased onto the updated #2421 head (`011c63b2`) and re-verified there.

- `make js-build` (runs `tsc --noEmit` + eslint + esbuild) — clean, and the checked-in `shiny/www` bundle is rebuilt in this commit. Confirmed again after the rebase that a clean rebuild of the parent branch reproduces its committed bundle byte-for-byte, so the bundle diff here is only this change.
- `tests/playwright/shiny/components/data_frame` + `tests/playwright/shiny/bugs/1844-datagrid-empty-column-name` — 48 passed.
- `tests/pytest` — 1104 passed, 1 skipped.

---


jat255 (approved):
The deletions look good. Neither `TableBodyCell` nor `updateCellsData` reads the prop, and the cell edit path identifies its target by `rowIndex`/`columnIndex` all the way through to the `{row_index, column_index, value}` payload. Two small suggestions before this gets merged:

**The PR description overstates the benefit.** The prop is dead, so removing it is the technically correct thing to do. But the description says that the change removes three rebuilds, but it doesn't actually result in fewer rebuilds:

Opus traced the mechanism as follows: an effect re-runs when *any* dep changes identity. Removing `columns` only reduces re-runs if every remaining dep is stable. Here, two remaining deps (`setCellEditMapAtLoc`, `resetCellEditMap`) get a fresh identity on every render, and "every render" is a superset of "every columns change", so the re-run frequency is unchanged. Same for `attemptUpdate` in `cell.tsx`.

So the deletion is still right on dead-code grounds, and the only thing wrong is the description's rationale. The likely real correction is inside `useCellEditMap`, which must give both functions a stable identity.

**Potential other things to remove.** Claude found two more dead dependencies in the arrays you're already editing. The `addPatches` effect body calls only `cellPatchPyArrToCellPatchArr` and `addPatchToData`, and never `setSorting`, which is still listed at `js/data-frame/index.tsx:647`. The `updateData` effect body calls only `updateData(evtData)`, so `resetCellEditMap` and `setTableData` at `:675` add nothing either (they are reachable through `updateData`'s own closure). Same class of thing this PR is removing.

Looks good, and up to you on these changes. Also worth noting for the stack: this removes the two `columns` annotations that #2421 retyped to `ColumnNames`, which shrinks that type's surface to `PandasData.columns` and the `updateData` parameter, which is better overall.

schloerke:
Thanks @jat255 — both points addressed in a94bccf.

**The description was wrong, and I've fixed it.** You (and Opus) are right about the mechanism: I traced `useCellEditMap()` and it builds `setCellEditMapAtLoc` and `resetCellEditMap` fresh on every render, with no `useCallback` — so those deps invalidate on every render, which strictly contains every columns change. Removing `columns` changes the re-run frequency by exactly nothing. The PR body and the commit message now say so explicitly and state that the change stands on the deps being dead, not on a re-render win.

The real fix is stabilizing those two identities inside `useCellEditMap()` (`cell-edit-map.tsx:28-54`), which would also make `attemptUpdate`'s `useCallback` in `cell.tsx` mean something. Deliberately not bolting that onto this PR — it's a behavior-adjacent change to a shared hook and deserves its own diff. Happy to open it as a fourth layer on the stack if you want it.

**Removed the other dead deps too.** Verified both bodies before touching them:

- `addPatches` calls only `cellPatchPyArrToCellPatchArr` and `addPatchToData({setData: setTableData, newPatches, setCellEditMapAtLoc})` — no `setSorting`. Deps are now `[id, setCellEditMapAtLoc, setTableData]`.
- `updateData` calls only `updateData(evtData)`, so `resetCellEditMap` and `setTableData` are reachable through that callback's closure. Deps are now `[id, updateData]`.

`make js-build` (which runs eslint, including `react-hooks/exhaustive-deps`) is clean, and the data frame playwright suite is 48 passed on this branch.

And agreed on the stack note — with these two gone, `ColumnNames` is down to `PandasData.columns` and the `updateData` parameter, which is where the wire format actually lives.

PR #2456, opened 2026-08-18
refactor(data frame): Rename the client's `columns` state to `columnNames`

schloerke:
Follow-up from reviewing #2421. Cosmetic; stacked on #2455.

Inside the data frame client, `columns` held column *names* while `cols`, `colIndex`, and the sort/filter payloads all hold column *indices*, so the two kinds of column identifier were hard to tell apart at a glance — which is how #1844 was able to hide: a name was being used where an id was needed.

## Scope

This is a rename of local identifiers only:

- `columns` state → `columnNames` (and `setColumns` → `setColumnNames`, `columnsProp` → `columnNamesProp`)
- the incoming `newColumns` in `updateData()` → `newColumnNames`
- the `colname` loop variable → `columnName`

**The wire contract is unchanged.** The `columns` key of the payload and of the `updateData` message is `FrameJson["columns"]` on the Python side, so renaming it would be a coordinated protocol change rather than a cosmetic one. `updateData()` destructures it as `columns: newColumnNames` at the boundary, with a comment saying so. If you'd rather rename the wire field too (and `PandasData.columns` with it), say the word and I'll extend this PR.

The type widening I'd noted on #2421 has since landed there as `ColumnNames = ReadonlyArray<string | number>`, so this PR is purely the naming half.

No behavior change.

## Verification

Rebased onto the updated #2421 head (`011c63b2`) and re-verified there. The rename was re-applied by hand rather than patch-merged, since #2421 rewrote `updateData()` in the meantime.

- `make js-build` (`tsc --noEmit` + eslint + esbuild) — clean; `shiny/www` bundle rebuilt in this commit. Only the source map changes: the minified `data-frame.js` is byte-identical, since the rename touches local identifiers only.
- `tests/playwright/shiny/components/data_frame` + `tests/playwright/shiny/bugs/1844-datagrid-empty-column-name` — 48 passed.
- `tests/pytest` — 1104 passed, 1 skipped.

---


jat255 (approved):
LGTM

PR #2457, opened 2026-08-18
refactor(data frame): Address columns by position on the Python side

schloerke:
Follow-up from reviewing #2421. Stacked on #2456, but independent of it — this one is Python-only and could be rebased onto `main` if you'd rather land it separately.

Two internal paths translated a column index into a column *name* and then back again. Column names are not dependable identifiers — they may be empty, and pandas allows names that are not strings — so both now stay positional.

## 1. `apply_frame_patches()` — fewer hops

It grouped patches into a `dict` keyed by column name, looked each name up with `nw_data.columns[index]`, and read the column back with `get_column(name)`. Patches arrive from the browser keyed by column index, so the dict is now keyed by index and the column is read positionally with `nw_data[:, index]`.

That drops both name lookups, and with them the `dict[Any, ScatterValues]` key type (and its explanatory comment) that the non-string-name case had forced. One implicit name step remains and is unavoidable: `with_columns()` matches the scattered Series back onto the frame by its name. That's noted in a comment rather than hidden.

## 2. `subset_frame(cols=)` — narrower internals, same caller-facing input

Callers may still pass names or positions, but resolution now happens once, in a new `as_col_indexes()`, and everything downstream is typed `ColIndexes` (positions only). Things that were previously implicit are now documented, validated, and tested:

- An integer is **always** a position, never a pandas integer label. The two can't be told apart, and narwhals reads a list of ints in the column slot positionally, so a frame labelled `{5: ..., 3: ...}` has to be addressed by position.
- "Integer" means anything integer-like, not just `int`. `isinstance(np.int64(5), int)` is `False`, so a numpy integer used to take the *name* branch and resolve by label equality — inverting the rule for exactly the values `np.where()` and `Index.get_indexer()` produce. Detection is `__index__`-based, and `ColsList` is widened from `str | int` to `str | SupportsIndex` to match.
- `bool` is a `TypeError`. It is a subclass of `int`, so `[True, False]` would otherwise read as positions `[1, 0]`. `StyleInfo`'s `cols` accepts a bool mask; this deliberately does not, rather than silently accepting a third form.
- Positions are bounds-checked. A negative position used to select from the end of the frame, and an out-of-range one escaped as narwhals' `IndexError: positional indexers are out-of-bounds`; both now raise a `ValueError` naming the position, as `update_cell_value()` and `style_info_cols()` already do.
- An unknown name raises `ValueError: Column name 'nope' not found in the data frame. To select a column by position, pass an int.` instead of `ValueError: 'nope' is not in list`.

## Also recorded, not fixed

While testing this I found that on **polars**, any selection that names columns positionally renames a column named `""` to `"column_<position in the result>"` — for `["a", "", "c"]`, `data[:, [0, 1, 2]]` gives `["a", "column_1", "c"]` and `data[:, [2, 1]]` gives `["c", "column_1"]`. It applies to the rows-and-columns branch too; only the rows-only branch is unaffected. No caller in shiny selects columns by position — `data_view()` passes `rows=` only — so this is latent. It's documented in the `subset_frame` docstring rather than worked around, since working around it would mean selecting by name and reintroducing the dependence on names this PR removes. Happy to file it upstream if you think it's a narwhals bug.

No user-facing behavior change, so no CHANGELOG entry.

## Verification

Rebased onto the updated #2421 head (`011c63b2`) — this commit applied cleanly, since the new work on #2421 is all client-side.

- New tests: `apply_frame_patches` on an empty-named column (pandas **and** polars), and `as_col_indexes` covering name resolution, int pass-through, mixed input, the empty list, the unknown-name error, the int-is-a-position rule (pinned with an int that is *also* a label, the only input where the two readings differ), numpy integers, negative and out-of-range positions, and `bool`.
- `tests/pytest` — 1111 passed, 2 skipped.
- `tests/playwright/shiny/components/data_frame` + `.../bugs/1844-datagrid-empty-column-name` — 48 passed, 1 rerun (the known `test_single_selection` flake).
- `uv run make format check-lint check-types` and `make check-pyright` — clean, 0 errors.

---

### DIFF
--- a/shiny/render/_data_frame_utils/_tbl_data.py
+++ b/shiny/render/_data_frame_utils/_tbl_data.py
@@ -6,12 +6,13 @@ import narwhals.stable.v1 as nw
 import orjson
 
 from ...session import Session, require_active_session
-from ...types import Jsonifiable, JsonifiableDict
+from ...types import Jsonifiable, JsonifiableDict, ListOrTuple
 from ._html import as_cell_html, ui_must_be_processed
 from ._types import (
     CellHtml,
     CellPatch,
     CellValue,
+    ColIndexes,
     ColsList,
     DataFrame,
     DataFrameT,
@@ -25,6 +26,7 @@ from ._types import (
 )
 
 __all__ = (
+    "as_col_indexes",
     "as_data_frame",
     "data_frame_to_native",
     "apply_frame_patches",
@@ -150,34 +152,34 @@ def apply_frame_patches(
     # This allows for a single column to be updated in a single operation (rather than multiple updates to the same column)
     #
     # In; patches: List[Dict[row_index: int, column_index: int, value: Any]]
-    # Out; cell_patches_by_column: Dict[column_name: Any, List[Dict[row_index: int, value: Any]]]
+    # Out; cell_patches_by_column: Dict[column_index: int, List[Dict[row_index: int, value: Any]]]
     #
-    # Narwhals types `get_column(name)`'s `name` as `str`, but its docs note that pandas
-    # does allow non-string column names and that they work when passed to it, as there
-    # is no ambiguity:
-    # https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.get_column
-    # This dict is not exported, so its key type is relaxed to match.
-    cell_patches_by_column: dict[Any, ScatterValues] = {}
+    # Patches arrive from the browser keyed by column *index*, and the column is read
+    # back out positionally, so the column name never enters the lookup. Names are not
+    # dependable keys: they may be empty or, in pandas, not even strings.
+    cell_patches_by_column: dict[int, ScatterValues] = {}
     for cell_patch in patches:
-        column_name = nw_data.columns[cell_patch["column_index"]]
-        if column_name not in cell_patches_by_column:
-            cell_patches_by_column[column_name] = {
+        column_index = cell_patch["column_index"]
+        if column_index not in cell_patches_by_column:
+            cell_patches_by_column[column_index] = {
                 "row_indexes": [],
                 "values": [],
             }
 
         # Append the row index and value to the column information
-        cell_patches_by_column[column_name]["row_indexes"].append(
+        cell_patches_by_column[column_index]["row_indexes"].append(
             cell_patch["row_index"]
         )
-        cell_patches_by_column[column_name]["values"].append(cell_patch["value"])
+        cell_patches_by_column[column_index]["values"].append(cell_patch["value"])
 
-    # Upgrade the Scatter info to new column Series objects
+    # Upgrade the Scatter info to new column Series objects.
+    # `nw_data[:, i]` selects positionally; the resulting Series keeps the column's
+    # name, which is what `with_columns()` below matches on.
     scatter_columns = [
-        nw_data.get_column(column_name).scatter(
+        nw_data[:, column_index].scatter(
             scatter_values["row_indexes"], scatter_values["values"]
         )
-        for column_name, scatter_values in cell_patches_by_column.items()
+        for column_index, scatter_values in cell_patches_by_column.items()
     ]
     # Apply patches to the nw data
     return nw_data.with_columns(*scatter_columns)
@@ -304,6 +306,47 @@ def serialize_frame(into_data: IntoDataFrame) -> FrameJson:
     }
 
 
+# as_col_indexes -----------------------------------------------------------------------
+def as_col_indexes(data: DataFrame[Any], cols: ListOrTuple[str | int]) -> ColIndexes:
+    """Resolve caller-supplied columns to column positions.
+
+    This is the single place where a column name is turned into a column position.
+    Everything downstream addresses columns by position only (:data:`ColIndexes`),
+    because names are not dependable identifiers: they may be empty, and pandas allows
+    names that are not strings.
+
+    Parameters
+    ----------
+    data
+        The data frame the columns belong to.
+    cols
+        Column names (`str`) or column positions (`int`).
+
+        An `int` is **always** read as a position, never as a label. The two cannot be
+        told apart otherwise, and narwhals reads a list of ints in the column slot
+        positionally, so a pandas frame whose labels are ints (e.g. `{5: ..., 3: ...}`)
+        must be addressed by position.
+
+    Raises
+    ------
+    ValueError
+        If a name is not a column of `data`.
+    """
+    col_indexes: list[int] = []
+    for col in cols:
+        if isinstance(col, int):
+            col_indexes.append(col)
+            continue
+        try:
+            col_indexes.append(data.columns.index(col))
+        except ValueError:
+            raise ValueError(
+                f"Column name {col!r} not found in the data frame. "
+                "To select a column by position, pass an `int`."
+            ) from None
+    return col_indexes
+
+
 # subset_frame -------------------------------------------------------------------------
 def subset_frame(
     data: DataFrameT,
@@ -311,14 +354,13 @@ def subset_frame(
     rows: RowsList = None,
     cols: ColsList = None,
 ) -> DataFrameT:
-    """Return a subsetted DataFrame, based on row positions and column names.
+    """Return a subsetted DataFrame, based on row positions and column names or positions.
 
     Note that when None is passed, all rows or columns get included.
-    """
-    # Note that this type signature assumes column names are string objects.
-    # This is always true in Polars, but not in Pandas (e.g. a column name could be an
-    # int, or even a tuple of ints)
 
+    Known narwhals behavior: on polars, selecting columns positionally renames a column
+    whose name is `""` to `"column_0"`. Selecting rows only is unaffected.
+    """
     # The nested if-else structure is used to navigate around narwhals' typing system and lack of `:` operator outside of `[`, `]`.
 
     # Note: pyright resolves `DataFrame.__getitem__` to the overload returning `Self`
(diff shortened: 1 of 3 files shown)


shiny/render/_data_frame_utils/_tbl_data.py
```diff
@@ -304,21 +306,61 @@ def default_orjson_serializer(val: Any) -> Jsonifiable:
...
+        told apart otherwise, and narwhals reads a list of ints in the column slot
+        positionally, so a pandas frame whose labels are ints (e.g. `{5: ..., 3: ...}`)
+        must be addressed by position.
+
+    Raises
+    ------
+    ValueError
+        If a name is not a column of `data`.
+    """
+    col_indexes: list[int] = []
```
jat255, line 361: **No bounds or sign validation on `int` entries** 

`cols=[-1]` returns the last column instead of raising, and `cols=[99]` on a two-column frame escapes as narwhals' `IndexError: positional indexers are out-of-bounds`. This is the same hazard #2421 added an explicit guard for in `update_sort()`, and `style_info_cols` already raises actionable messages for both cases (`_styles.py:102-107`). A helper whose docstring calls itself "the single place where a column name is turned into a column position" (`:313`) is probably a good place to validate.

shiny/render/_data_frame_utils/_tbl_data.py
```diff
@@ -304,21 +306,61 @@ def default_orjson_serializer(val: Any) -> Jsonifiable:
...
+    for col in cols:
+        if isinstance(col, int):
```
jat255: **`isinstance(col, int)` is `True` for `bool`** 

`cols=[True, False, True]` on a three-column frame resolves to `[1, 0, 1]`, so column 1 twice and column 0 once, rather than a `TypeError` or a mask. The codebase already knows this trap in both directions: `update_cell_value` guards it explicitly (`_data_frame.py:874`) and `style_info_cols` treats a bool list as a mask (`_styles.py:76-81`). Worth picking one and following it.

shiny/render/_data_frame_utils/_tbl_data.py
```diff
@@ -304,21 +306,61 @@ def default_orjson_serializer(val: Any) -> Jsonifiable:
...
+    names that are not strings.
+
+    Parameters
+    ----------
+    data
+        The data frame the columns belong to.
+    cols
+        Column names (`str`) or column positions (`int`).
+
+        An `int` is **always** read as a position, never as a label. The two cannot be
```
jat255: **numpy integers invert the documented rule**

`isinstance(np.int64(5), int)` is `False`, so a numpy int takes the name branch, and on a pandas frame labelled `[5, 3]` the lookup `data.columns.index(np.int64(5))` succeeds by equality and returns position 0, the column *labelled* 5. A plain `5` would raise for being out of range. So "an `int` is **always** read as a position, never as a label" is backwards for values out of `np.arange`, `np.where`, or `Index.get_indexer`, which are the most likely sources of positions. Either accept `np.integer` alongside `int` or narrow the claim.

shiny/render/_data_frame_utils/_tbl_data.py
```diff
@@ -304,21 +306,61 @@ def default_orjson_serializer(val: Any) -> Jsonifiable:
...
+    """Return a subsetted DataFrame, based on row positions and column names or positions.
 
     Note that when None is passed, all rows or columns get included.
-    """
-    # Note that this type signature assumes column names are string objects.
-    # This is always true in Polars, but not in Pandas (e.g. a column name could be an
-    # int, or even a tuple of ints)
 
+    Known narwhals behavior: on polars, selecting columns positionally renames a column
+    whose name is `""` to `"column_0"`. Selecting rows only is unaffected.
```
jat255: **The polars note is understated**

(minor nit)

On polars 1.43.2 with narwhals 2.24.0, the replacement name is `column_<position in the result>`, not always `column_0`: e.g. for a frame `['a', '', 'c']`, `data[:, [0, 1, 2]].columns` is `['a', 'column_1', 'c']` and `data[:, [2, 1]].columns` is `['c', 'column_1']`. It also applies to the `data[rows, col_indexes]` branch, which the note doesn't mention, so only the rows-only path is genuinely unaffected. As written, someone grepping for `column_0` will miss the case where the empty-named column isn't first in the selection. Agreed on not working around it, and yes, worth filing upstream.

shiny/render/_data_frame_utils/_tbl_data.py
```diff
@@ -304,21 +306,61 @@ def default_orjson_serializer(val: Any) -> Jsonifiable:
     }
 
 
+# as_col_indexes -----------------------------------------------------------------------
+def as_col_indexes(data: DataFrame[Any], cols: ListOrTuple[str | int]) -> ColIndexes:
+    """Resolve caller-supplied columns to column positions.
+
+    This is the single place where a column name is turned into a column position.
```
jat255: **The "single place" claim in the new docstring isn't true**

Two other paths still resolve a name to a position: `update_cell_value` inline at `_data_frame.py:866-871` (`Column '{col}' not found in data frame.`), and `style_info_cols` at `_styles.py:83-95` (a third message, and unlike `as_col_indexes` it does validate int sign and range at `:97-107`). So the sentence holds for `subset_frame` but not for the whole codebase. I would just soften this docstring to reflect this.

tests/pytest/test_render_data_frame_tbl_data.py
```diff
@@ -425,6 +443,34 @@ def test_subset_frame_numeric_column_names():
...
+
+
+def test_as_col_indexes_unknown_name_raises():
+    df = as_data_frame(pd.DataFrame({"chr": ["a"], "num": [1]}))
+
+    with pytest.raises(ValueError, match="Column name 'nope' not found"):
+        as_col_indexes(df, ["nope"])
+
+
+def test_as_col_indexes_int_is_always_a_position():
```
jat255, line 466: Both assertions here are already covered elsewhere in this file. `as_col_indexes(df, [0]) == [0]` is the int-passthrough case that `test_as_col_indexes_resolves_names_and_positions` asserts with `as_col_indexes(df, [2, 0]) == [2, 0]`. `subset_frame(df, cols=[0]).columns == [5]` is selecting position 0 from a frame with integer labels, which `test_subset_frame_numeric_column_names` asserts with labels `[10, 20]` and `cols=[0]` giving `[10]`. Both pass on any implementation those two pass on.

The rule this test is named for is that an `int` is a position and never a label, and the only input that separates those two readings is an int that is also a label. On this frame that input is `5`: read as a position it gives `[5]`, which is out of range for two columns, and read as a label it gives `[0]`. The test passes `0`, where both readings agree. `as_col_indexes(df, [5])` is the assertion that pins the rule.

Two inputs have no test here, and both currently return a value rather than raising: `[-1]` returns `[-1]`, so `subset_frame(df, cols=[-1])` returns the last column, and `[99]` returns `[99]`, so narwhals raises `IndexError: positional indexers are out-of-bounds` from inside `subset_frame`. Whether those should raise is the subject of the other comments on `as_col_indexes` directly

jat255 (commented):
Looks mostly good. Opus review found a few minor issues probably worth resolving before merging this in.

shiny/render/_data_frame_utils/_tbl_data.py
schloerke: Fixed in 4adddbf. `as_col_indexes()` now bounds-checks every position:

- negative → `ValueError: Column position -1 must be greater than or equal to 0.`
- out of range → `ValueError: Column position 99 is out of range. The data frame has 2 columns.`

Followed `update_cell_value()`/`style_info_cols()` in raising `ValueError` and naming the offending position, so nothing reaches narwhals that it would answer with `IndexError: positional indexers are out-of-bounds` (or, for `-1`, answer wrongly but silently). Covered by `test_as_col_indexes_out_of_range_position_raises`.

shiny/render/_data_frame_utils/_tbl_data.py
schloerke: Fixed in 4adddbf — `bool` now raises `TypeError: Column True must be a column name (`str`) or a column position (`int`), not a `bool`.`

I followed `update_cell_value()` rather than `style_info_cols()`: the mask reading is a `StyleInfo`-specific affordance, and `ColsList` is documented as names-or-positions, so quietly accepting a third input shape seemed worse than rejecting it. If you'd rather it accept a mask for consistency with `cols=` on styles, that's a small change — say the word.

Covered by `test_as_col_indexes_bool_raises`.

shiny/render/_data_frame_utils/_tbl_data.py
schloerke: Good catch — accepted rather than narrowed the claim, in 4adddbf.

Position detection is now `__index__`-based (`_is_col_position()`), so any integer-like value is a position, `np.int64` included. `ColsList` is widened from `str | int` to `str | SupportsIndex` so the type says the same thing the docstring does.

`bool` is excluded first, since it also satisfies `__index__` — that's the thread above.

Pinned by `test_as_col_indexes_numpy_int_is_a_position`, which uses the case you describe: on a frame labelled `[5, 3]`, `np.int64(5)` now raises out-of-range instead of resolving to position 0 by label equality.

shiny/render/_data_frame_utils/_tbl_data.py
schloerke: Reproduced exactly as you describe (polars 1.43.2, narwhals 2.24.0) and corrected the note in 4adddbf. It now says the replacement is `column_<position in the result>`, gives both of your examples, and states that it applies to the rows-and-columns branch as well, with only the rows-only branch unaffected. Version numbers included so the note is falsifiable later.

Will file it upstream with narwhals.

shiny/render/_data_frame_utils/_tbl_data.py
schloerke: You're right — softened in 4adddbf. It now reads "the single place where `subset_frame()` turns a column name into a column position", with a parenthetical pointing at `update_cell_value()` and `style_info_cols()` as the other name lookups.

Worth noting for later: those three now differ in both message text and validation strictness (`as_col_indexes()` picked up sign/range checks in this round, so it matches `style_info_cols()`; `update_cell_value()` has a third message). Consolidating them is a bigger change than this PR, but it's the obvious next step if we want one story about column addressing.

tests/pytest/test_render_data_frame_tbl_data.py
schloerke: Agreed on all counts — the test asserted nothing the two tests above it didn't already cover. Rewritten in 4adddbf to use `5`, the input where the two readings actually diverge:

```python
df = as_data_frame(pd.DataFrame([["a", 1]], columns=[5, 3]))

with pytest.raises(ValueError, match="Column position 5 is out of range"):
    as_col_indexes(df, [5])

# Position 1 is the column labelled `3`, not the label `1` (which does not exist)
assert as_col_indexes(df, [1]) == [1]
assert subset_frame(df, cols=[1]).columns == [3]
```

Reading `5` as a label would give `[0]`; reading it as a position is out of range on a two-column frame, so the test now fails on a label-reading implementation.

`[-1]` and `[99]` both raise now (the sign/range thread), and each has its own case in `test_as_col_indexes_out_of_range_position_raises`.

PR #2460, opened 2026-08-18
ci: Upgrade astral-sh/setup-uv to v10.0.1 for manifest fetch retries

### LINKED ISSUE #2459: ci: setup-uv manifest fetch has no retry, flaking CI at setup
`astral-sh/setup-uv@v8.2.0` resolves the uv version by fetching a manifest over the network, and that fetch has no retry. A single transient network blip kills the job during setup, before anything is built or tested.

## Observed failure

[`dev-shiny-rsconnect-local-connect` in "Run deploy tests"](https://github.com/posit-dev/py-shiny/actions/runs/32180773599/job/95853003964) failed in the `Setup py-shiny` step:

```
Fetching manifest data from https://raw.githubusercontent.com/astral-sh/versions/main/v1/uv.ndjson ...
##[error]fetch failed
```

This is a flake, not a real failure. The sibling job `github-shiny-dev-rsconnect-local-connect` in the same run fetched the identical URL successfully five seconds earlier, and this was the only failed run in the repo's previous 60.

## Root cause

1. `.github/py-shiny/setup/action.yaml` does not pass a `version:` input, and the repo has no `uv.toml` or `[tool.uv] required-version`, so setup-uv resolves `latest`.
2. Resolving `latest` requires fetching the version manifest from `raw.githubusercontent.com`.
3. In v8.2.0 that request has a 5 second timeout and zero retries ([`src/utils/fetch.ts`](https://github.com/astral-sh/setup-uv/blob/v8.2.0/src/utils/fetch.ts)). Any connection error fails the step outright. Ours failed in 47ms, so it was an immediate connection error rather than a timeout.

Pinning an exact uv version would not avoid this. The download path calls `getArtifact()`, which also fetches the same manifest, and the runner tool cache misses on a fresh image (the passing sibling job logged `Downloading uv from ...`).

## Fix

Upstream added retries for exactly this failure in [astral-sh/setup-uv#1016](https://github.com/astral-sh/setup-uv/pull/1016), which retries the manifest fetch three times with backoff. It shipped in [v10.0.1](https://github.com/astral-sh/setup-uv/releases/tag/v10.0.1) and is in no earlier release, so fixing this means moving from v8.2.0 to v10.0.1.

Three call sites pin setup-uv, and none of them have the fix:

- `.github/py-shiny/setup/action.yaml` (v8.2.0), the one that flaked here
- `.github/workflows/pytest.yaml`, oldest-supported-deps job (v8.2.0)
- `.github/workflows/pytest.yaml`, narwhals integration job (v7)

The upgrade crosses two major versions. v9.0.0 flips the `prune-cache` default from `true` to `false` to reduce load on PyPI infrastructure, which grows our Actions caches. Given that we have a cache entry per Python version and #2303 was specifically about CI speed, the upgrade should set `prune-cache: true` explicitly so cache behavior does not drift. v10.0.0's breaking change only affects `enable-cache: auto`, and we set `enable-cache: true`, so it does not apply.

jat255:
Fixes #2459.

`astral-sh/setup-uv@v8.2.0` resolves the uv version by fetching a manifest from `raw.githubusercontent.com`, and in that release the request has a 5 second timeout and no retry. One transient connection error kills the job during setup, which is what happened in [this "Run deploy tests" job](https://github.com/posit-dev/py-shiny/actions/runs/32180773599/job/95853003964). Upstream added three retries with backoff around that fetch in [astral-sh/setup-uv#1016](https://github.com/astral-sh/setup-uv/pull/1016), first released in v10.0.1.

All three call sites move to v10.0.1, including the narwhals integration job that was still on the floating `v7` tag and has the same exposure. The two guard messages in the narwhals composite actions referenced `@v7`, so they are updated to match.

## Notes for review

Pinning an exact uv `version:` instead would not have avoided this. The download path calls `getArtifact()`, which fetches the same manifest, and the runner tool cache misses on a fresh image.

The upgrade crosses two major versions, and only one breaking change affects us. v9.0.0 flipped the `prune-cache` default from `true` to `false`, so this PR sets `prune-cache: true` explicitly to keep cache behavior identical to today rather than letting it drift as a side effect of a reliability fix. The reasoning is in a comment next to the input. v10.0.0's breaking change only applies to `enable-cache: auto`, and we set `enable-cache: true`.

Verified that the retry is present in the shipped `dist/setup/index.cjs` bundle at the v10.0.1 tag, not only in source, and that every input we pass is still declared in that release's `action.yml`.

This only takes effect for CI runs after merge, so the fix cannot be demonstrated in this PR's own checks.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -411,8 +411,12 @@ jobs:
           python-version: "3.12"
       # Pinned to an exact release; astral-sh/setup-uv has no floating `v10` tag
       - uses: astral-sh/setup-uv@v10.0.1
-              with:
-                prune-cache: true
+        with:
+          # setup-uv v9 flipped this default to false. We keep pruning
+          # because we store a separate cache entry per Python version, and
+          # unpruned entries risk hitting the repo cache limit and evicting
+          # each other.
+          prune-cache: true
 
       - name: Narwhals integration setup
         uses: ./.github/py-shiny/narwhals-setup
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -96,10 +96,15 @@ jobs:
           fetch-depth: 0
 
       - name: Install `uv`
-        # Pinned to an exact release; astral-sh/setup-uv has no floating `v8` tag
-        uses: astral-sh/setup-uv@v8.2.0
+        # Pinned to an exact release; astral-sh/setup-uv has no floating `v10` tag
+        uses: astral-sh/setup-uv@v10.0.1
         with:
           enable-cache: true
+          # setup-uv v9 flipped this default to false. We keep pruning
+          # because we store a separate cache entry per Python version, and
+          # unpruned entries risk hitting the repo cache limit and evicting
+          # each other.
+          prune-cache: true
           cache-dependency-glob: "pyproject.toml"
           cache-suffix: py3.10-oldest-deps
           activate-environment: true
@@ -404,7 +409,8 @@ jobs:
       - uses: actions/setup-python@v6
         with:
           python-version: "3.12"
-      - uses: astral-sh/setup-uv@v7
+      # Pinned to an exact release; astral-sh/setup-uv has no floating `v10` tag
+      - uses: astral-sh/setup-uv@v10.0.1
 
       - name: Narwhals integration setup
         uses: ./.github/py-shiny/narwhals-setup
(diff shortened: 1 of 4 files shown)


.github/workflows/pytest.yaml
```diff
@@ -404,7 +409,8 @@ jobs:
       - uses: actions/setup-python@v6
         with:
           python-version: "3.12"
-      - uses: astral-sh/setup-uv@v7
+      # Pinned to an exact release; astral-sh/setup-uv has no floating `v10` tag
+      - uses: astral-sh/setup-uv@v10.0.1
```
karangattu, line 413:
```suggestion
      - uses: astral-sh/setup-uv@v10.0.1
              with:
                prune-cache: true
```

This is because after `v9.0` the default value for prune-cache was changed to false.

karangattu (approved):
LGTM....thanks

PR #2464, opened 2026-08-20
fix(playwright): Allocate remote server ports dynamically

### LINKED ISSUE #2444: Flaky CI: playwright-examples shard fails when the remote Playwright server port is already in use (EADDRINUSE)
Every test in the `playwright-examples (3.11, firefox, 0)` shard failed because the remote Playwright server never came up: the container's `playwright run-server` could not bind its port, so the readiness check timed out.

- Failed job: https://github.com/posit-dev/py-shiny/actions/runs/31626891028/job/94215473182
- Commit: `ed140dd9` (PR #2441)
- Job: `playwright-examples (3.11, firefox, 0)` (run attempt 1)
- Re-run of the same job on the same commit: passed (run attempt 2)

## Error

```
Timed out waiting for Playwright firefox server on ws://127.0.0.1:43424/: BrowserType.connect: WebSocket error: socket hang up
Error: listen EADDRINUSE: address already in use 0.0.0.0:43424
```

## Notes

The `EADDRINUSE` comes from inside the server container, which is started by
`.github/py-shiny/setup-playwright-remote/action.yaml` as:

```
docker run ... -p "127.0.0.1:43424:43424" ... \
  /bin/sh -c "npx --no -- playwright run-server --port 43424 --host 0.0.0.0"
```

The port appears to be derived per-job rather than probed for availability, so
two shards scheduled onto the same runner (or a lingering container from an
earlier step) can collide on it. `run-server` then exits immediately, the
"Wait for Playwright server" step gets `socket hang up` for the full timeout,
and the whole shard fails before running a single test.

This is an infra race in the setup action, not a test-code bug — the same
commit passed on re-run with no changes.

Related but distinct: #2424 is the same setup action failing *after* the server
is up (`TargetClosedError` from the readiness check closing a shared browser).
This one is the server failing to start at all.

karangattu:
## Summary

- The action selects an available container port when `port` is empty.
- Docker selects an available host port and returns it to the action.
- The readiness probe and pytest use the resolved host port.
- A caller can still set `port` for a fixed mapping.

## Problem

The action used port `43424` for every remote Playwright server. A port collision stopped `run-server` before the tests started.

Closes #2444

### DIFF
--- a/.github/py-shiny/setup-playwright-remote/start-server.sh
+++ b/.github/py-shiny/setup-playwright-remote/start-server.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [ -n "${PLAYWRIGHT_PORT:-}" ]; then
+  container_port="$PLAYWRIGHT_PORT"
+  published_port="127.0.0.1:${container_port}:${container_port}"
+else
+  container_port=$(python3 - <<'PY'
+import socket
+
+with socket.socket() as sock:
+    sock.bind(("127.0.0.1", 0))
+    print(sock.getsockname()[1])
+PY
+  )
+  published_port="127.0.0.1::${container_port}"
+fi
+
+if ! [[ "$container_port" =~ ^[0-9]+$ ]] || [ "$container_port" -lt 1 ] || [ "$container_port" -gt 65535 ]; then
+  echo "Invalid Playwright server port: $container_port" >&2
+  exit 1
+fi
+
+docker rm -f "$CONTAINER_NAME" >/dev/null 2>&1 || true
+
+# No --rm: if the server crashes, the container must survive so the
+# logs-on-failure step can read it. The command above removes an old container.
+# Playwright exits run-server when stdin closes. Keep stdin open in detached
+# mode, as required by Playwright's documented Docker setup.
+docker run -d \
+  --interactive \
+  --name "$CONTAINER_NAME" \
+  --ipc=host \
+  -p "$published_port" \
+  --init \
+  --workdir /home/pwuser \
+  --user pwuser \
+  "$IMAGE_REF" \
+  /bin/sh -c "npx --no -- playwright run-server --port $container_port --host 0.0.0.0"
+
+host_binding=$(docker port "$CONTAINER_NAME" "$container_port/tcp")
+host_port="${host_binding##*:}"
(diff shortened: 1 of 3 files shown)


.github/py-shiny/setup-playwright-remote/start-server.sh
```diff
@@ -0,0 +1,49 @@
...
+else
+  container_port=$(python3 - <<'PY'
+import socket
+
+with socket.socket() as sock:
+    sock.bind(("127.0.0.1", 0))
+    print(sock.getsockname()[1])
+PY
+  )
+  published_port="127.0.0.1::${container_port}"
```
karangattu, line 18: Do we need a 2nd container port value?
Similar to Line no 7 -> published_port

.github/py-shiny/setup-playwright-remote/start-server.sh
```diff
@@ -0,0 +1,49 @@
...
+fi
+
+if ! [[ "$container_port" =~ ^[0-9]+$ ]] || [ "$container_port" -lt 1 ] || [ "$container_port" -gt 65535 ]; then
```
karangattu, line 22: confirm 65... is the limit that socket.socket() binds to

tests/pytest/test_playwright_reliability.py
```diff
@@ -43,6 +47,51 @@ def test_remote_playwright_readiness_uses_server_metadata() -> None:
     assert "socket.socket" not in action
 
 
+def test_remote_playwright_server_uses_dynamic_default_ports(
```
karangattu: Remove this test

.github/py-shiny/setup-playwright-remote/start-server.sh
karangattu: No, that is intentional. `127.0.0.1::<container>` leaves the host port empty so Docker picks a free one — only the container port is fixed. Line 7 binds host = container because `PLAYWRIGHT_PORT` requests that exact host port.

.github/py-shiny/setup-playwright-remote/start-server.sh
karangattu: Yes — TCP/UDP port numbers are unsigned 16-bit integers, so 65535 is the maximum valid port. The `socket.socket()` bind above requests an ephemeral port (0), which the OS assigns within that range.

PR #2468, opened 2026-08-24
fix: make checkbox selection assertions order-independent

### LINKED ISSUE #2123: [Bug]: out-of-order values cause expect_selected() to fail for shiny.playwright.controller.InputCheckboxGroup
### Component

Other

### Severity

### Shiny Version

1.5.0

### Python Version

3.13.7

```python
from shiny import App, ui
app_ui = ui.page_fluid(
    ui.input_checkbox_group(
        "parameters",
        "Parameters",
        choices={"A": "a", "B": "b", "C": "c", "D": "d"},
        selected=["A"],
        inline=True,
    ),
)
    pass
app = App(app_ui, server)

## test_checkboxgroup.py
from playwright.sync_api import Page
from shiny.playwright import controller
from shiny.run import ShinyAppProc

def test_cbg_works(page: Page, local_app: ShinyAppProc):
    page.goto(local_app.url)
    input_control = controller.InputCheckboxGroup(page, "parameters")
    input_control.expect_selected(["A"])

def test_cbg_works_with_values_in_same_order(page: Page, local_app: ShinyAppProc):
    page.goto(local_app.url)
    input_control = controller.InputCheckboxGroup(page, "parameters")
    input_control.set(["B", "C"])  # works
    input_control.expect_selected(["B", "C"])  # works

def test_cbg_works_with_values_in_different_order(page: Page, local_app: ShinyAppProc):
    page.goto(local_app.url)
    input_control = controller.InputCheckboxGroup(page, "parameters")
    input_control.set(["B", "A"])  # works
    input_control.expect_selected(["B", "A"])  # does not work
```

### Behavior

Current: Verifying selected checkboxes in a InputCheckboxGroup fails if the list of values is not in the same order as in the checkbox definition.
Expected: Regardless of ordering, the .expect_selected() method should verify and assert that the checkboxes are selected.

### Error Messages (if any)

```shell
=========================== short test summary info ===========================
FAILED tests/features/test_checkboxgroup.py::test_cbg_works_with_values_in_different_order[chromium] - AssertionError: Locator expected to have attribute 'B'
Actual value: A 
Call log:
  - Expect "to_have_attribute" with timeout 5000ms
  - waiting for locator("div#parameters.shiny-input-checkboxgroup.shiny-bound-input").locator(locator("input[type=checkbox]:checked")).first
    9 × locator resolved to <input value="A" type="checkbox" name="parameters" checked="checked"/>
      - unexpected value "A"
======================== 1 failed, 2 passed in 13.85s =========================
Finished running tests!
```

### Environment

```markdown
OS: win11
Browser: Chromium 140
uv: 0.9.10
pytest: 8.4.2
playwright: 1.55.0
```

EltonChang1:
## Summary

- verify selected checkbox-group values as an unordered set
- preserve exact-count and unique-expectation checks
- keep string and regular-expression expectations supported
- exercise the existing kitchen-sink case with expected values in reverse DOM order

Fixes #2123

## Testing

- `pytest -c tests/playwright/playwright-pytest.ini tests/playwright/shiny/inputs/input_kitchensink/input_checkbox_group/test_input_checkbox_group_kitchensink.py --browser chromium -q` — 2 passed
- `black --check`
- `isort --check-only`
- `flake8`

### DIFF
--- a/shiny/playwright/controller/_input_controls.py
+++ b/shiny/playwright/controller/_input_controls.py
@@ -8,7 +8,7 @@ from playwright.sync_api import expect as playwright_expect
 
 from ...types import MISSING, MISSING_TYPE, ListOrTuple
 from .._types import AttrValue, ListPatternOrStr, PatternOrStr, Timeout
-from ..expect._expect import _attr_match_str
+from ..expect._expect import _attr_match_str, _xpath_match_str
 from ..expect._internal import (
     expect_attribute_to_have_value as _expect_attribute_to_have_value,
 )
@@ -22,6 +22,7 @@ from ._base import (
     not_is_missing,
 )
 from ._expect import (
+    assert_arr_is_unique,
     expect_locator_contains_values_in_list,
     expect_locator_values_in_list,
 )
@@ -923,20 +924,19 @@ class InputCheckboxGroup(
         timeout
             The timeout for the expectation. Defaults to `None`.
         """
+        assert_arr_is_unique(value, "`selected` must be unique")
+
         # Playwright doesn't like lists of size 0
         if len(value) == 0:
             playwright_expect(self.loc_selected).to_have_count(0, timeout=timeout)
             return
 
-        expect_locator_values_in_list(
-            page=self.page,
-            loc_container=self.loc_container,
-            el_type="input[type=checkbox]",
-            arr_name="selected",
-            arr=value,
-            timeout=timeout,
-            is_checked=True,
-        )
+        playwright_expect(self.loc_selected).to_have_count(len(value), timeout=timeout)
+        for item in value:
+            selected_item = self.loc_selected.locator(
+                f'xpath=self::*[{_xpath_match_str("value", item)}]'
+            )
+            playwright_expect(selected_item).to_have_count(1, timeout=timeout)
 
 
 class InputCheckbox(_InputCheckboxBase):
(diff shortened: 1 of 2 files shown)


schloerke:
Thanks for digging into this, @EltonChang1, and sorry for the slow response.

We looked at this and I think we want to **keep `expect_selected()` order-sensitive**, so I'm inclined to close this PR — but #2123 is a real problem, just with a different fix. Reasoning below so it's on the record.

**The order is meaningful.** We checked empirically with a 3-choice group (`a`, `b`, `c`), checking `c` first and then `b`:

```python
@render.text
def out():
    return repr(input.cbg())
# -> ('b', 'c')
```

The input value is always in choices/DOM order, never click order. So the ordered list passed to `expect_selected()` *is* the value the server sees, and we'd like to keep that correspondence — a reader of the test can map the assertion directly onto `input.cbg()`.

**Two implementation concerns with the order-independent version:**

1. *Atomicity.* `expect_locator_values_in_list()` deliberately composes a single locator (the chained `loc_container.locator("xpath=.", has=...)`) and makes one `to_have_count()` assertion, so the whole check resolves against one DOM state. Replacing it with a count assertion followed by a loop of per-item assertions gives each check its own independent auto-wait, so a group whose choices update mid-assertion can satisfy every individual check without ever having been in the asserted state.

2. *Set equality is harder than it looks.* Dropping the positional constraint also drops injectivity — nothing forces two expected items onto two *distinct* elements. For exact strings that's fine (an element has one `value`, so distinct strings match disjoint elements, and `assert_arr_is_unique` + the count assertion make set equality provable). For `re.Pattern` it isn't: with `{Option A, Option D}` checked, `expect_selected(["Option A", re.compile("A$")])` has both items resolve to the same `Option A` element, total count 2, so it passes while `Option D` matches nothing. A true set comparison needs a perfect bipartite matching, which can't be expressed as a single Playwright locator.

**What we'd like instead for #2123.** The actual defect there is the error message — `Locator expected to have attribute 'B' / Actual value: A` gives no hint that order is the contract. So:

- when the ordered assertion fails, compare the expected and actual values as sets first; if they match, raise an error saying the values must be listed in choices order, showing expected vs. found;
- document the order contract in the `expect_selected()` docstring.

That keeps the single atomic assertion and turns the confusing failure into a self-explanatory one. If you'd like to take that on, we'd happily review it — otherwise no worries at all, and thanks for the careful writeup and testing notes here.

PR #2474, opened 2026-08-26
ci(tests): pin plotly to <7 to fix failing example tests

karangattu:
### Summary

Plotly 7.0.0 was recently released and removed `plotly.figure_factory.create_distplot`, which caused the `nba-dashboard` external template tests in `test_external_templates.py` to fail with:
```
AttributeError: module 'plotly.figure_factory' has no attribute 'create_distplot'
```

This pins `plotly<7` in the optional test dependencies in [pyproject.toml](file:///Users/karangathani/Documents/GitHub/py-shiny/pyproject.toml) so CI installs a compatible version of Plotly for example tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,7 @@ test = [
     "shinychat>=0.6.0",
     "seaborn",
     "plotnine",
-    "plotly",
+    "plotly<7",
     "anywidget",
     "duckdb",
     "holoviews",


cpsievert:
It'd be good practice for us to generally be testing against the latest plotly. 

Would you mind taking a stab at replacing `create_distplot()` with plotly's "official" figure API? I would think most models would be fairly capable of that.

karangattu:
> It'd be good practice for us to generally be testing against the latest plotly.
> 
> Would you mind taking a stab at replacing `create_distplot()` with plotly's "official" figure API? I would think most models would be fairly capable of that.

I wil make the changes in the templates repo, so it is fixed upstream

karangattu:
@cpsievert https://github.com/posit-dev/py-shiny-templates/pull/56

cpsievert:
Thanks! Does that mean this could be closed?

karangattu:
Resolved it by fixing it upstream

PR #2475, opened 2026-08-26
feat: Add `ui.page_html()` and `express.ui.page_opts(html=)` for a complete, author-owned HTML document as the UI

### LINKED ISSUE #2462: App() cannot accept a pre-rendered full HTML document (HTMLTextDocument) with extra dependencies
## Use case

[shinyreact](https://github.com/posit-dev/shinyreact) serves apps whose UI is a complete `index.html` owned by the app author (the kind a Vite build emits). The server needs to inject its own `HTMLDependency` objects (a JS bundle plus a JSON config script tag) into that document's `<head>` and have their routes registered — i.e., the Python equivalent of R's `shinyApp(ui = htmlTemplate("index.html", document_ = TRUE))` with `attachDependencies()`, which works today.

## Current behavior (shiny 1.7.0)

`htmltools.HTMLTextDocument` is exactly the right primitive — py-shiny itself uses it for `App(ui=Path)` in `_render_page_from_file()`, injecting `require_deps() + jquery_deps() + shiny_deps()` at the `<meta name="shiny-dependency-placeholder">` marker and registering routes via `_ensure_web_dependencies()`. But there is no public path for a package to do the same with *additional* dependencies:

- `App(ui=HTMLTextDocument(...))` fails: `_render_page()` calls `ui.get_dependencies()`, which `HTMLTextDocument` doesn't have — and even if it did, the document would be wrapped in a second `HTMLDocument`, nesting `<html>` inside `<html>`.
- `App(ui=Path)` injects only shiny's own deps; there is no hook to add more.
- A `Callable[[Request], Tag | TagList]` UI goes through `_render_page()`'s `HTMLDocument` wrapper, so a full document can't be returned from it either.

```python
from htmltools import HTMLTextDocument
from shiny import App
App(HTMLTextDocument("<html>...{{ marker }}...</html>", deps=[...], deps_replace_pattern="{{ marker }}"), server)
# AttributeError: 'HTMLTextDocument' object has no attribute 'get_dependencies'
```

## Proposal

Accept `HTMLTextDocument` (or `RenderedHTML`) as `ui`:

```python
elif isinstance(ui, HTMLTextDocument):
    rendered = ui.render(lib_prefix=self.lib_prefix)
    self._ensure_web_dependencies(rendered["dependencies"])
    self.ui = rendered
```

(Callers would be responsible for including `require_deps()/jquery_deps()/shiny_deps()` in the document's deps list, or `App` could prepend them as `_render_page_from_file()` does.)

## Workaround we're shipping meanwhile

A small `App` subclass in shinyreact that does exactly the three lines above via the private `_ensure_web_dependencies()` — happy to upstream this as a PR if the approach sounds right.

schloerke:
Closes #2462.

Adds `ui.page_html()`, for apps whose UI is a complete HTML document the author owns — the `index.html` a JS bundler emits — rather than one Shiny builds from `ui.page_*()` components. `App` accepts its result and serves it as-is: no `HTMLDocument` wrapper, which would nest `<html>` inside `<html>`.

```python
from shiny import App, ui

app = App(
    ui.page_html(
        Path(__file__).parent / "dist" / "index.html",
        extra_deps=[my_bundle_dep],
    ),
    server,
)
```

It takes the document as a string or a `Path` (read as UTF-8 each time the function is called), and **`extra_deps=`** instead of a fully specified `deps=`, prefixing Shiny's own dependencies onto them. (The `HTMLTextDocument` subclass doing the work, `PageHtmlDocument`, is a private implementation detail.) So app authors don't have to know or track Shiny's dependency list — which is not the three objects it looks like: `shiny_deps()` returns `shiny` (`shiny.js` + `shiny.min.css`) *and* `shiny-busy-indicators`, plus `shiny-devmode` when `SHINY_DEV_MODE=1`. Order matters too: RequireJS must load first (Shiny ships a patched copy that sets `window.define.amd = false`, so UMD-packaged libraries don't hit [mismatched anonymous define()](https://requirejs.org/docs/errors.html#mismatch)), then jQuery, which `shiny.js` is written against.

`deps_replace_pattern=` defaults to `<meta name="shiny-dependency-placeholder" content="">`, documented in the docstring and echoed by the error raised when it's missing.

**Prefixing at construction keeps every dependency in one `<script type="application/html-dependencies">` manifest.** Injecting Shiny's into an already-rendered document would emit a second manifest, and the client reads those tags with a jQuery selector plus `.text()` (`shiny/www/shared/shiny.js:7629`), which concatenates every match with no separator — two manifests glue into `shiny[1.7.0]my-dep[1.0.0]` at the seam, registering one bogus dependency and silently re-rendering the real ones. Doing it at construction also means no reaching into py-htmltools private state, which an earlier revision of this PR needed.

### One place renders a UI, one place defines the dependencies

`_render_page()` is now the only place a UI is rendered, branching on what it got, and `_render_page_from_file()` folds into it. **It's also the method the UI-function path renders through**, so handling a type there is what makes it work in *both* positions — passed to `App(ui=)` directly, and returned by a UI function:

```python
App(lambda request: ui.page_html(index_html, extra_deps=[...]), server, bookmark_store="url")
```

That matters because bookmarking requires `App(ui=)` to be a function — `_init_bookmarking()` rejects a static UI, since the UI has to be reconstructed from the bookmarked state. So both bookmark and non-bookmark apps are covered.

The page-level dependency set now has a single definition, `html_dependencies._page_deps(include_css=)`, splatted by both page forms so they can't drift:

```python
deps=[*_page_deps(include_css=True), *(extra_deps or [])]          # ui.page_html()
ui_res = TagList(*_page_deps(include_css=not has_bootstrap), ui)   # the tag-tree branch
```

`include_css` differs because a complete document has no tag tree to inspect for the Bootstrap dependency (whose CSS bundles Shiny's). A test renders both forms and asserts identical dependency names in identical order. `_page_deps` is deliberately not exported: its contents and ordering are exactly what we want to stay free to change, and `ui.page_html()` covers what an app author would want it for. The other functions in `html_dependencies.py` gained docstrings.

Three things are rejected in `_render_page()`, so the errors cover the UI-function return value too, not just the direct argument:

- a plain `htmltools.HTMLTextDocument` → points at `ui.page_html()`, instead of the `AttributeError: ... has no attribute 'get_dependencies'` from the issue;
- an `htmltools.HTMLDocument` → it hoists dependencies into `<head>` in tag-tree order and its only public mutator is `append()`, so Shiny's could only land *after* the app author's (jQuery after the scripts needing it), and appending would push it off the single-`<html>`-child path into a nested `<html>`. Pass the tags themselves instead;
- a `Path` → it names a file to read once at startup rather than a per-pageview UI value, so it stays handled in `App.__init__()`. A UI function that wants to serve a file returns `ui.page_html(path)`, which reads it on every call, making the per-request read explicit.

That invariant — every UI *value* type works in both positions, with `Path` as the deliberate exception — is written down in `.claude/references/architecture.md` under "HTML Generation and Dependencies", since it was previously unstated and easy to miss.

### Express support

`ui.page_auto()` gains an `html: str | Path` argument that routes the whole page through `ui.page_html()`, and Express's `ui.page_opts()` forwards it like every other page option:

```python

ui.page_opts(html=Path(__file__).parent / "index.html")

@render.text
def greeting():
    return "Hello from the server!"
```

The markup of collected top-level UI is dropped -- the document already is the page -- but its HTML dependencies (`ui.include_css()`, widget assets) are harvested into `extra_deps=`, so they still load. Combining `html=` with `page_fn=`, or with page options like `title=`/`theme=` that cannot apply to a document served as-is, raises a `ValueError`. `page_fn=` (on both `page_auto` and `page_opts`) widens to `Callable[..., Tag | PageHtmlDocument]`.

The two `.tagify()` sites in `express/_run.py` (static and bookmark-wrapper paths) pass a `PageHtmlDocument` through untouched; `App._render_page()` already handles it in both positions, so bookmark-enabled Express apps get this for free. `PageHtmlDocument` deliberately does **not** grow a `.tagify()` method: htmltools discovers tagifiability by duck-typing (`getattr(x, "tagify", None)`), and a complete document must keep failing fast when someone tries to embed it in a tag tree (e.g. displaying a `page_html()` result inside `with ui.card():`), rather than silently nesting `<html>` inside `<html>`.

### R precedent

`renderPage()` has exactly this branch ([`R/shinyui.R#L31-L89`](https://github.com/rstudio/shiny/blob/62d2a6c1c772415e34201c4c5a1ee85d5278bfab/R/shinyui.R#L31-L89)):

```r
  # If the ui is a NOT complete document (created by htmlTemplate()), then do some
  # preprocessing and make sure it's a complete document.
  if (!inherits(ui, "html_document")) {
    ...
  }
  shiny_deps <- c(list(jqueryDependency()), shinyDependencies())
  html <- renderDocument(ui, shiny_deps, processDep = createWebDependency)
```

When the UI already *is* a complete document, R skips the wrapping and renders it directly — and it passes its own `shiny_deps` into `renderDocument()`, which merges them with the ones the author attached via `attachDependencies()` and routes them all through `createWebDependency` (R's `_ensure_web_dependencies()`). So Shiny supplying its own dependencies, rather than the author, matches R. `shinyApp()` reaches this via `uiHttpHandler()` ([`R/shinyapp.R#L83`](https://github.com/rstudio/shiny/blob/62d2a6c1c772415e34201c4c5a1ee85d5278bfab/R/shinyapp.R#L83)).

### One thing to review

**`App(ui=Path)` now raises when the file has no dependency placeholder.** Previously the replacement silently did nothing and you got a page with no Shiny on it. Technically a behavior change for a file lacking `<meta name="shiny-dependency-placeholder" content="">`, but such an app was already broken. Easy to scope the error to the new path if you'd rather not touch `Path` behavior.

### Testing

`tests/pytest/test_app_ui_page_html.py`: Shiny's deps are added and routed; `extra_deps` share one manifest and load after `shiny.js`; a custom `deps_replace_pattern` works; a document with no placeholder raises; the document and tag-tree forms carry identical dependencies; and a plain `HTMLTextDocument`, an `HTMLDocument`, and a `Path` each raise in both positions. Express: an app using `ui.page_opts(html=)` serves the document with harvested dependencies and dropped top-level markup, and the `html=` + `page_fn=`/page-option conflicts raise. The `page_html` api-example has both `app-core.py` and `app-express.py` variants, exercised by the playwright-examples CI job.


.claude/references/architecture.md
```diff
@@ -82,6 +82,34 @@ HTML is generated using the `htmltools` package:
 Asset vendoring (bslib CSS/JS, theme presets, `make upgrade-html-deps`) is
 covered in `.claude/references/assets.md`.
 
+### App UI types
+
+`App(ui=)` accepts a `Tag`/`TagList`, a `Path` to a complete HTML file, a
+`ShinyHTMLTextDocument` (a complete HTML document with Shiny's dependencies
+prefixed onto the app author's), or a function taking a `Request` and returning
+any of those.
```
schloerke:
```suggestion
any of those (except for `Path`, see below).
```

.claude/references/architecture.md
```diff
@@ -82,6 +82,34 @@ HTML is generated using the `htmltools` package:
...
+
+**Every UI *value* type must work in both positions: passed directly, and
```
schloerke:
```suggestion
**Every UI *value* type (except for `Path`) must work in both positions: passed directly, and
```

.claude/references/architecture.md
```diff
@@ -82,6 +82,34 @@ HTML is generated using the `htmltools` package:
...
+returned by a UI function.** The function form is what bookmarking requires --
+`App._init_bookmarking()` rejects a static UI, since the UI has to be
+reconstructed from the bookmarked state -- so a type supported only when passed
+directly is silently unavailable to bookmark-enabled apps.
+
+`App._render_page()` is the single place both are rendered, so handle a new UI
+value type there rather than in `App.__init__()`, and extend the
+`Callable[[Request], ...]` return union alongside the direct one.
+
```
schloerke: Remove

.claude/references/architecture.md
```diff
@@ -82,6 +82,34 @@ HTML is generated using the `htmltools` package:
...
+The page-level dependency set (requirejs, jQuery, Shiny) has one definition,
+`html_dependencies._page_deps()`. Both page forms -- the tag tree and the
+complete document -- splat it, so they cannot drift. `HTMLDocument` is rejected
+as a UI: it hoists dependencies into `<head>` in tag-tree order, so Shiny's
+could only land after the app author's.
```
schloerke: Remove last sentence

.claude/references/architecture.md
```diff
@@ -82,6 +82,34 @@ HTML is generated using the `htmltools` package:
...
+
+`Path` is the deliberate exception: it names a file to read once at startup
```
schloerke, line 104:
```suggestion
### `Path` exception

`Path` is the deliberate exception: it names a file to read once at startup
```

shiny/__init__.py
```diff
@@ -50,6 +50,7 @@
     "otel",
     # _app.py
     "App",
+    "ShinyHTMLTextDocument",
```
schloerke: Having `ShinyHTMLTextDocument` at the top level makes it feel too important. 

Maybe within `ui`?

shiny/_app.py
```diff
@@ -499,38 +577,73 @@ def _register_web_dependency(self, dep: HTMLDependency) -> None:
...
-            require_deps(),
-            jquery_deps(),
-            *shiny_deps(include_css=not has_bootstrap),
-            ui,
-        )
-        rendered = HTMLDocument(ui_res).render(lib_prefix=lib_prefix)
-        self._ensure_web_dependencies(rendered["dependencies"])
-        return rendered
+    def _render_page(
+        self, ui: Tag | TagList | ShinyHTMLTextDocument, lib_prefix: str
```
schloerke:
```suggestion
        self, ui: Tag | TagList | ShinyHTMLTextDocument, lib_prefix: str,
```

This will cause a format change as well

.claude/references/architecture.md
schloerke: Applied in 430dc583.

.claude/references/architecture.md
schloerke: Applied in 430dc583.

.claude/references/architecture.md
schloerke: Removed in 430dc583.

The comment anchored to the blank line after the paragraph, so I read "Remove" as the whole `App._render_page()` paragraph rather than just the trailing `Callable[[Request], ...]` clause. Reopen if you meant the narrower cut.

.claude/references/architecture.md
schloerke: Removed the `HTMLDocument` sentence in 430dc583. The behavior it described is still enforced (and tested) in `_render_page()`; it is just no longer restated here.

.claude/references/architecture.md
schloerke: Applied in 430dc583.

shiny/_app.py
schloerke: Applied in bd8c43da — and yes, black then split it one parameter per line.

shiny/__init__.py
schloerke: Agreed — moved out of the top level in bd8c43da, then renamed in bd2e4d7e. It is now **`ui.PageDocument`**, in a new `shiny/ui/_page_document.py`, documented under **Page containers** next to `ui.page_output`:

```python
from shiny import App, ui

app = App(ui.PageDocument(index_html, extra_deps=[my_bundle_dep]), server)
```

I scanned the rest of the exported surface first, and `shiny.ui` really is the only fit — `shiny.__all__` is just `App`, `run_app`, `module`, `Session`, `Inputs`, `Outputs`, `req`, the legacy `render_*` aliases and `event`, and the other subpackages (`reactive`, `render`, `session`, `quarto`, `otel`) are the wrong domain. A class there is well-precedented: `ui.Theme`, `ui.Progress`, `ui.Sidebar`, `ui.Chat` and friends.

Two follow-on decisions worth a look:

- **`ShinyHTMLTextDocument` → `PageDocument`.** `Shiny` was redundant inside `shiny.ui`, and the name now matches the family it belongs to. I did *not* use `ui.HTMLTextDocument`: it would shadow htmltools' class of that name with different semantics (`extra_deps=` rather than `deps=`, dependencies added for you), unlike `ui.HTML` / `ui.Tag` / `ui.TagList`, which are identity re-exports.
- **It needed a `_known_missing_express_ui` entry.** Anything in `shiny.ui.__all__` must either exist in `shiny.express.ui` or be listed there, or `test_express_ui_is_complete` fails. Listed with the reason inline: Express builds the page itself, so an author-owned document has no Express counterpart. (This is a cost the top-level export did not have.)

schloerke:
CI note: the three `playwright-examples` failures on this PR are unrelated to it — `nba-dashboard` from py-shiny-templates hits `AttributeError: module 'plotly.figure_factory' has no attribute 'create_distplot'`, after plotly 7.0.0 (released yesterday) removed twelve deprecated Figure Factory functions. Filed as #2476 (CI) and posit-dev/py-shiny-templates#57 (the fix), to be handled in a separate PR. Every check owned by this PR passes.

docs/_quartodoc-core.yml
```diff
@@ -22,6 +22,7 @@ quartodoc:
         - ui.page_bootstrap
         - ui.page_auto
         - ui.page_output
+        - ui.PageDocument
```
cpsievert: I'm not in love with the combination of these two names (`Page` and `Document`). It doesn't give me much info about what it's about (`index.html`). That said, I can see how the `htmltools` `HTMLDocument`/`HTMLTextDocument` precedence makes this a hard naming problem.

There's a part of me saying that we should go with `HTMLDocument`, even if it's different from `htmltools.HTMLDocument`. In fact, I think there's an argument to be made that `htmltools` got it wrong: the current `HTMLTextDocument` should be `HTMLDocument` and the current `HTMLDocument` should be something like `TagDocument` instead.

shiny/ui/_page_document.py
```diff
@@ -0,0 +1,68 @@
...
+    ----------
+    html
+        A complete HTML document, including ``<html>``. It must contain
+        ``deps_replace_pattern`` to mark where the dependencies are inserted.
+    extra_deps
+        Additional HTML dependencies to include, alongside Shiny's own. These are
+        inserted after Shiny's, and their files are served by the app.
+    deps_replace_pattern
+        The string in ``html`` to replace with the dependencies. The first instance is
+        replaced. Defaults to :attr:`DEPS_PLACEHOLDER`.
```
cpsievert:
```suggestion
        The string in ``html`` to replace with Shiny's dependencies. The first instance is
        replaced. Defaults to :attr:`DEPS_PLACEHOLDER`.
```

docs/_quartodoc-core.yml
schloerke: `ShinyHTMLDocument` . `ShinyTagDocument` ?

Feels more like a shim, not a template

schloerke: Let's go with `page_html(html: str | Path, extra_deps) -> HTMLDocument`

A "page_ui()" fn isn't needed as we already support tag objects which can hold html dependencies. 

The benefit of `page_html` is that we get a file / text that also support html deps.

schloerke: Done in 666a957f: `ui.PageDocument` is now `ui.page_html(html: str | Path, *, extra_deps=, deps_replace_pattern=)`, joining the existing `page_*()` family.

- A `Path` is read (as UTF-8) each time the function is called, so a UI function returning `ui.page_html(path)` gives an explicit per-pageview read; `App(ui=Path)` also routes through it.
- The `HTMLTextDocument` subclass survives only as a private implementation detail (`PageHtmlDocument`, needed for the isinstance check in `_render_page()` and as the return type) — the public API is just the function.
- `extra_deps=` is now `ListOrTuple[HTMLDependency]`, and the docstring example moved to a tagged `@add_example()` app under `shiny/api-examples/page_html/`.

shiny/api-examples/page_html/index.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>A page Shiny did not build</title>
+    <meta name="shiny-dependency-placeholder" content="">
```
schloerke, line 6:
```suggestion
    <!-- A meta tag that is replaced with shiny's dependencies (and any provided `extra_deps=`) -->
    <meta name="shiny-dependency-placeholder" content="">
```

shiny/express/ui/__init__.py
```diff
@@ -359,6 +359,9 @@
         "page_fluid",
         "page_auto",
         "page_output",
+        # The whole page is an author-owned document; in Express, use
+        # `ui.page_opts(html=)` instead
```
schloerke: Remove

shiny/express/_run.py
```diff
@@ -131,7 +132,10 @@ def create_express_app(file: Path, package_name: str) -> App:
             # catch them here and convert them to a different type of error, because uvicorn
             # specifically catches AttributeErrors and prints an error message that is
             # misleading for Shiny Express. https://github.com/posit-dev/py-shiny/issues/937
-            app_ui = run_express(file, package_name).tagify()
+            ui_res = run_express(file, package_name)
+            # A `PageHtmlDocument` (from `ui.page_opts(html=)`) is a complete document
+            # that `App` renders as-is; it has no `.tagify()`.
```
schloerke:
```suggestion
```

shiny/api-examples/page_html/index.html
schloerke: Applied in 6e8f2cce.

shiny/express/ui/__init__.py
schloerke: Done in 6e8f2cce: removed from the known-missing list, which (per the parity test) means `page_html` is now re-exported from `shiny.express.ui` and registered in `docs/_quartodoc-express.yml`.

shiny/express/_run.py
schloerke: Done in 6e8f2cce — collapsed to a single line (`# A \`PageHtmlDocument\` (from \`ui.page_opts(html=)\`) has no \`.tagify()\`.`) so the first line isn't left dangling mid-sentence.

PR #2490, opened 2026-09-07
test: run session teardown in bookmark file-input restore test (#2482)

### LINKED ISSUE #2482: Bookmark file-input restore leaks a TemporaryDirectory when a session never ends
## Summary

A `tempfile.TemporaryDirectory` created while restoring a bookmarked file input is cleaned up only via `session.on_ended()`. If the session never ends, the directory survives until the object is garbage collected, at which point `tempfile` emits a `ResourceWarning: Implicitly cleaning up <TemporaryDirectory ...>`.

https://github.com/posit-dev/py-shiny/blob/main/shiny/input_handler.py#L252-L254

```python
if len(value_list) > 0:
    tempdir_root = tempfile.TemporaryDirectory()
    session.on_ended(lambda: tempdir_root.cleanup())
```

## Impact

Low. In a running app the session ends and the callback fires, so this is mostly a test-environment artifact — a unit test that builds a session and never closes it leaves the directory to the collector.

It became visible when the unit suite adopted `filterwarnings = error` in #2480: the `ResourceWarning` is raised inside a finalizer, becomes an unraisable exception, and pytest reports it as a `PytestUnraisableExceptionWarning` against whichever test happened to be running when the collector fired (`test_theme_css_compiles_and_is_cached` on the oldest-deps job). Because the attribution is random, `ResourceWarning` is now ignored in `pytest.ini`; that ignore hides this leak rather than fixing it.

Note that `shiny/ui/_theme.py` deliberately relies on GC-time cleanup for its own temp directories (see `theme_temporary_directories`), so it is a second, intentional source of the same warning class.

## Possible resolutions

1. Leave as-is and close — the on-ended cleanup is correct for real sessions, and the warning only appears in tests.
2. Make the test harness close its sessions, so the callback runs and the ignore in `pytest.ini` can be narrowed or dropped.
3. Suppress the warning at the source by cleaning up deterministically where the lifetime is known.

Filing this so the decision is recorded rather than lost behind the `pytest.ini` ignore.

drewe7192:
Closes #2482

### Description

In `tests/pytest/test_test_mode.py`, `test_file_restore_handler_registers_snapshot_preprocess` exercises the `shiny.file` bookmark restore handler. This handler instantiates a `tempfile.TemporaryDirectory` and registers cleanup via `session.on_ended(lambda: tempdir_root.cleanup())`.

Because the test previously exited without ending the session lifecycle, the cleanup callback never fired during the test execution. The directory was left to Python's cyclic garbage collector, intermittently emitting an unraisable `ResourceWarning: Implicitly cleaning up <TemporaryDirectory ...>` during test teardown.

### Solution

- Marked `test_file_restore_handler_registers_snapshot_preprocess` with `@pytest.mark.asyncio` and converted it to an `async def`.
- Wrapped the assertion block in a `try...finally` block that calls `await session._run_session_ended_tasks()`, ensuring the registered `on_ended` cleanup hook executes deterministically at test completion.

### Testing

- Verified that `pytest tests/pytest/test_test_mode.py -k test_file_restore_handler_registers_snapshot_preprocess` passes cleanly without unhandled warnings.

### DIFF
--- a/tests/pytest/test_test_mode.py
+++ b/tests/pytest/test_test_mode.py
@@ -906,7 +906,8 @@ async def test_snapshot_preprocess_output_namespaced(
     assert body["output"]["mod1-out1"] == "HELLO"
 
 
-def test_file_restore_handler_registers_snapshot_preprocess(
+@pytest.mark.asyncio
+async def test_file_restore_handler_registers_snapshot_preprocess(
     monkeypatch: pytest.MonkeyPatch,
     tmp_path: Path,
 ) -> None:
@@ -942,5 +943,8 @@ def test_file_restore_handler_registers_snapshot_preprocess(
     }
     restored = cast("list[Any]", handler(value, ResolvedId("file1"), session))
 
-    assert isinstance(restored, list) and len(restored) == 1
-    assert ResolvedId("file1") in session.input._snapshot_preprocessors
+    try:
+        assert isinstance(restored, list) and len(restored) == 1
+        assert ResolvedId("file1") in session.input._snapshot_preprocessors
+    finally:
+        await session._run_session_ended_tasks()


schloerke:
Thank you, @drewe7192 ! 

comedy fox GIF by New Girl

